PR #5, opened 2024-04-09
Add brief style guidelines.

pelesh:
This should probably be merged to `develop`, but for opportunistic convenience, I suggest we merge it here. Since #1 will likely prompt discussion on new style guidelines, let's direct this PR to `composer-dev` branch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,15 @@ If an output is needed (for example, a warning needs to be displayed), use `std:
 std::cout << "index out of bounds. Row " << i << " starts at: " << start << " and ends at " << end << std::endl;
 ```
 
+### Local variables naming
 
-### Member variable naming
+Local variable names should use C-style name format.
+```c++
+double local_variable; // Yes
+double localVariable;  // No, using lowercase camel instead of C-style name format
+```
+
+### Member variables naming
 
 Member variable names should use C-style name format and end with trailing underscore `_`.
 ```c++
@@ -50,12 +57,14 @@ Always define `enum`s inside `GridKit` namespace. Type names should be capitaliz
 
 ### Constants
 
-If a constant is used in more than one file, define it in `Common.h`. Constants names should be capitalized.
+If a constant is used in more than one file, define it in `Common.h`. For
+constants with long names, use underscores to separate words in the constant
+name. Use all caps (screaming snake case).
 
 ```c++
-   constexpr double Pi = 3.1415; // No, it should be all caps
-   constexpr double SQRT_TWO = 1.4142 // No, there is an underscore
-   constexpr double SQRTTWO_ = 1.4142 // No, there is an underscore
+   constexpr double Pi = 3.1415; // No, the constant name should be all caps
+   constexpr double SQRT_TWO = 1.4142 // Yes
+   constexpr double SQRTTWO_ = 1.4142 // No, there is a trailing underscore
    constexpr double EXP = 2.7183 // Yes   
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,194 @@
+# Developer Guidelines
+
+## Code Style
+
+### Existing non-compliant code
+The code that does not comply with these guidelines will be fixed in separate pull request(s). Any new contribution should follow closely these guidelines but should not change style in the existing GridKit code.
+
+### Error handling
+Return values of member functions should be of type `int` and used for error handling. Functions return 0 if no error is encounter, return positive value for warnings and recoverable error, and negative value for irrecoverable errors.
+
+### Output
+If an output is needed (for example, a warning needs to be displayed), use `std::cout` and not `printf` as shown below. There should be a space before and after each `<<`.
+
+```c++
+std::cout << "index out of bounds. Row " << i << " starts at: " << start << " and ends at " << end << std::endl;
+```
+
+
+### Member variable naming
+
+Member variable names should use C-style name format and end with trailing underscore `_`.
+```c++
+double member_variable_; // Yes
+double another_member;   // No, there is no trailing underscore to distinguish it from nonmember variables
+double memberVariable_;  // No, using lowercase camel instead of C-style name format
+```
+
+### Function names
+
+Use lowercase camel format for function names.
+```c++
+int myFunction(double x); // Yes
+int another_function();   // No, using C-style name format
+int YetAnotherFunction(); // No, using uppercase camel name format
+```
+
+### Class names
+
+Class names should start with a capital letter. For instance, `Vector` and `Matrix` are valid class names, while `point` is not.
+
+### Enums (enumerated types)
+
+Always define `enum`s inside `GridKit` namespace. Type names should be capitalized and the constant names should be uppercase with underscores (but there is no underscore at the end!).
+
+```c++
+  enum ExampleEnum { CONST_ONE = 0,
+                     CONST_TWO = 8, 
+                     YET_ANOTHER_CONST = 17 };
+``` 
+
+### Constants
+
+If a constant is used in more than one file, define it in `Common.h`. Constants names should be capitalized.
+
+```c++
+   constexpr double Pi = 3.1415; // No, it should be all caps
+   constexpr double SQRT_TWO = 1.4142 // No, there is an underscore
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
...
+## Code Style
+
+### Existing non-compliant code
+The code that does not comply with these guidelines will be fixed in separate pull request(s). Any new contribution should follow closely these guidelines but should not change style in the existing GridKit code.
+
+### Error handling
+Return values of member functions should be of type `int` and used for error handling. Functions return 0 if no error is encounter, return positive value for warnings and recoverable error, and negative value for irrecoverable errors.
+
+### Output
+If an output is needed (for example, a warning needs to be displayed), use `std::cout` and not `printf` as shown below. There should be a space before and after each `<<`.
```
reid-g, line 12: Would there be cases where `printf`'s pretty print would be preferred over `std::cout`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
...
+
+```c++
+std::cout << "index out of bounds. Row " << i << " starts at: " << start << " and ends at " << end << std::endl;
+```
+
+
+### Member variable naming
```
reid-g, line 19: Will need to add specification that local variables are snake style.

CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
...
+                     YET_ANOTHER_CONST = 17 };
+``` 
+
+### Constants
+
+If a constant is used in more than one file, define it in `Common.h`. Constants names should be capitalized.
+
+```c++
+   constexpr double Pi = 3.1415; // No, it should be all caps
+   constexpr double SQRT_TWO = 1.4142 // No, there is an underscore
```
reid-g, line 57: This should be set to valid. So snake-style with all caps. `SQRTTWO` is hard to read. This would also standard with C library macros

alexander-novo: fyi this is called *screaming* snake case

CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
+# Developer Guidelines
+
+## Code Style
+
+### Existing non-compliant code
+The code that does not comply with these guidelines will be fixed in separate pull request(s). Any new contribution should follow closely these guidelines but should not change style in the existing GridKit code.
+
+### Error handling
+Return values of member functions should be of type `int` and used for error handling. Functions return 0 if no error is encounter, return positive value for warnings and recoverable error, and negative value for irrecoverable errors.
```
alexander-novo, line 9: This seems a little unwieldy - does this mean any outputs must be passed by reference or pointer? Perhaps we could consider returning `std::optional` for warnings/recoverable errors and throwing for unrecoverable errors?

CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
...
+The pointer `*` or reference `&` belong to the type and there should be no space between them and the type name.
+```c++
+double* x;     // Yes
+int& n;        // Yes
+double *x, *y; // No, the pointer symbol is a part of `double*` type
+int & n;       // No, the reference symbol is a part of `int&` type
+```
+
+### Indentation
+Use only spaces for indentation, not tabs. Indent size is 4 spaces.
```
alexander-novo, line 73: I'm fine with this if this really is the preference of everyone else, however this could be considered an accessibility issue. With tabs for indentation, people who are hard of sight can increase the indentation size to their preference while others can keep it as low as they want.

What I typically suggest is a tabs for indentation, spaces for alignment policy for this reason. This should be easy to enact with a `.clang-format` file.

alexander-novo:
I think it would also be a good idea to set a (soft) column limit of 120 characters.

CONTRIBUTING.md
pelesh: I don't see immediate cases where `printf` would be preferred. If so, we could consider `std::format` per @alexander-novo suggestion. For now, I would say this is outside the scope for this PR. If `printf` becomes needed, I suggest we create separate issue.

CONTRIBUTING.md
pelesh: Outputs are indeed passed by reference or a pointer in the current code. Even if the output is the function return value it would still be a pointer to an object such as Jacobian matrix. I don't see a major disadvantage if we reserve return value for function execution status.

Having said that, what you suggest is a good idea, so please create a separate issue with more details. That would require some refactoring of the GridKit code, as well.

CONTRIBUTING.md
```diff
@@ -0,0 +1,194 @@
...
+                     CONST_TWO = 8, 
+                     YET_ANOTHER_CONST = 17 };
+``` 
+
+### Constants
+
+If a constant is used in more than one file, define it in `Common.h`. Constants names should be capitalized.
+
+```c++
+   constexpr double Pi = 3.1415; // No, it should be all caps
```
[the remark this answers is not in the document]
pelesh: Numeric constants in STL do not include all the constants we need to define. We still need to define our own constants as needed.

CONTRIBUTING.md
pelesh: Addressed in [c63fbd8](https://github.com/ORNL/GridKit/pull/5/commits/c63fbd882ecb78183dcd95c7fd77fbc875b53538).

CONTRIBUTING.md
pelesh: Addressed in [c63fbd8](https://github.com/ORNL/GridKit/pull/5/commits/c63fbd882ecb78183dcd95c7fd77fbc875b53538).

CONTRIBUTING.md
```diff
@@ -0,0 +1,203 @@
...
+int main()
+{
+  //some code
+  return 0;
+}
+```
+The `system` includes should always be listed first.
+
+### Using namespaces
+All classes should be in namespace `GridKit`. If needed, define additional namespaces inside `GridKit`.
```
alexander-novo, line 181: @pelesh I've noticed there are some namespaces, such as [ModelLib](https://github.com/ORNL/GridKit/blob/60647ed1053a2ed3dab0c45855801246e6d0ee01/ModelEvaluatorImpl.hpp#L66) that do not follow this trend. Is this meant to be an exception or should this be fixed?

CONTRIBUTING.md
```diff
@@ -0,0 +1,203 @@
...
+{
+  std::cout 
+}
+```
+
+Also, leave one empty line between `system` includes and `GridKit` includes, i.e.,
+```c++
+#include <cstring>
+
+#include <SparseMatrix/CooMatrix.hpp>
```
alexander-novo, line 170: I think it would be nice to have 3 separate blocks: std libraries, external libraries, and GridKit. To facilitate this, it would maybe be a good idea to change local includes to use `#include "..."` rather than `<...>`

CONTRIBUTING.md
pelesh: This should be fixed. However, I suggest we grandfather noncompliant code and merge developer guidelines first. Then we can document issues where we don't see compliance with the guidelines.

CONTRIBUTING.md
pelesh: Included in [eda2cc2](https://github.com/ORNL/GridKit/pull/5/commits/eda2cc23966865beb11e923048898dcd588fe8d4).

reid-g:
Some details that I think should be added for clarification.

Initializing classes with the `:`. The member initializer list can add some weird issues on readability when there are a large number of them. This document should address this.

Structures localized as parameter inputs. When there are a lot of parameters it becomes a lot more manageable to use structures to handle them. Perhaps add some details on this?

These points come from details in the `DistributedGenerator.hpp` file.

Also with regards to file names and classes. Perhaps state that files with only one class names must match the name of their file (or something similar).

Otherwise looks good to me.

CONTRIBUTING.md
reid-g: I am of preference to scientific notation when looking at floats (and displaying them). However, sticking with `std::cout` only is fine for now. If `std::format` is available that would work as well.

alexander-novo:
I concur with @reid-g's points

pelesh:
> Structures localized as parameter inputs. When there are a lot of parameters it becomes a lot more manageable to use structures to handle them. Perhaps add some details on this?

This is important issue but somewhat out of scope of this PR. @reid-g would you mind creating a separate GitHub issue with details and examples?

PR #7, opened 2024-04-09
Remove parts of generator model from BusPV

### LINKED ISSUE #2: Bus Parameters of Generator
The `BusPV` depends on the parameter `Pg` for creation. Currently under the `BusData` does not contain `Pg`. This will need to be added or setup as separate input. However, the parameter generally comes from a generator. Then the parse will need to be updated for this as well.

pelesh:
Fixes #2.


reid-g (approved):
Looks good now and is consistent.

PR #11, opened 2024-05-08
Added Scalable Microgrid Example

reid-g:
- Added example of scalable microgrid with arbitrary size


pelesh (approved):
All tests pass, clean example implementation. Minor corrections suggested before the merge.

CMakeLists.txt
```diff
@@ -88,6 +88,9 @@ set(CMAKE_MACOSX_RPATH 1)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/CMake)
 
+# Add source directory to macros for file access
+add_compile_definitions("SOURCE_ROOT=${CMAKE_SOURCE_DIR}")
+
```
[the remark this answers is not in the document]
pelesh: I think something like [this](https://github.com/ORNL/ReSolve/blob/develop/tests/functionality/testKLU.cpp#L35) is a better solution. The compiled binary would look for the data in the current directory unless user specifies a [different path](https://github.com/ORNL/ReSolve/blob/develop/tests/functionality/CMakeLists.txt#L148).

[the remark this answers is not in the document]
pelesh: Okay, I see, if I install `scalmicrogrid` and run it, the test will fail because it cannot find reference file.

Examples/Scale_Microgrid/Scale_Microgrid.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+    idas->runSimulation(t_final);
+
+    std::vector<double>& yfinial = sysmodel->y(); 
+
+    std::cout << "Final Vector y\n";
+    for (size_t i = 0; i < yfinial.size(); i++)
+    {
+        printf("%lu: % 2.16e\n", i, yfinial[i]);
+    }
```
pelesh, line 281: Use I/O streams because they are overloaded unlike C-style print format functions.

Examples/Scale_Microgrid/Scale_Microgrid.cpp
```diff
@@ -0,0 +1,317 @@
...
+    //All data generated with abstol=1e-12 and reltol=1e-12 with ode23tb
+    if (orgparams && compDataSizes.end() != std::find(compDataSizes.begin(), compDataSizes.end(), Nsize))
+    {   
+        //get the cmake source dir from the cmake configuration
+        std::string base_path = XSTRING(SOURCE_ROOT);
+
+        base_path += "/Examples/Scale_Microgrid/microgrid_N";
+        base_path += std::to_string(Nsize);
+        base_path += ".txt";
+        std::cout << base_path << std::endl;
```
pelesh, line 294: Instead of this, use current directory as a default and allow user to specify directory as an input parameter. Also, see `ParameterEstimation.cpp` example.

Examples/Scale_Microgrid/Scale_Microgrid.cpp
```diff
@@ -0,0 +1,317 @@
...
+            true_vec[i++] = val_cur;
+        }
+        data_file.close();
+
+        //check relative error
+        std::cout << "Test the Relative Error\n";
+        for (size_t i = 0; i < true_vec.size(); i++)
+        {
+            printf("%lu: % 2.16e\n", i, abs(true_vec[i] - yfinial[i]) / abs(true_vec[i]));
+        }
```
pelesh, line 312: Use I/O stream rather than print format function here.

pelesh (changes requested):
Minor corrections as suggested in this review would make this example cleaner.

pelesh (approved):

PR #12, opened 2024-06-07
Organize component models by model family

pelesh:
Create subdirectories in `ComponentLib` where each subdirectory contains components of specific model family. Currently existing families are Power Flow and Power Electronics models. Dynamic Phasor family of models will be added, right now there is only some model documentation there.

There are ~100 files changed but all of it is just moving existing files and updating `#include` directives.


Paulm991 (commented):
Compiled and ctested without error

reid-g (approved):
Looks good to me. Compiles and all test pass.

PR #13, opened 2024-06-25
Genrou test

Paulm991:
image
[genrououtputjac.txt](https://github.com/user-attachments/files/15974351/genrououtputjac.txt)
I believe there is an issue with how I am handling the imaginary numbers.  When I change f_[17] or f_[18] to -y_[17] + y_[16]/X_, it fails to calculate it at t=0.

The files to look for are GENROU.cpp, GENROU.hpp, and GenrouTest.cpp. I also made minor changes to Ida.cpp for printing and SystemModel.hpp so that it detects Jacobians.


pelesh:
It seems that `genrou-test` was branched off from a dated `develop`. I will close this for now. Please reopen when the branch is rebased and ready for review.

PR #15, opened 2024-07-09
Genrou test 2

Paulm991:
GenrouGridkitWithSat7-8-24

GENROU test produces results that are feasible. Some tests now fail:

9 - AdjointSens (Failed)
12 - ScaleMicrogrid (Failed)
13 - DynamicConOpt (Failed)
16 - ParameterEst (Failed)

I created the SlackBus files but haven't connected them to the GenrouTest yet. GenrouTest is still using PowerFlow buses.


pelesh (changes requested):
You need to fix all failing tests before this is ready to merge. :)

You should probably undo all the changes to `Ida` class since these are not necessary for GENROU implementation and are likely cause of test failures. `Ida` functionality should be controlled through its API, not by modifying its code.

Issue #16, opened 2024-07-10
Scale Microgrid Test Error


Paulm991:
Trying to run the ScaleMicrogrid test is throwing an error on the 'develop' branch. The error output is as follows:

"ScaleMicrogrid" start time: Jul 10 14:10 EDT
Output:
Test the Relative Error
Test with Nsize = 2 passes!
Test the Relative Error
Test with Nsize = 4 passes!
Test the Relative Error
Test with Nsize = 8 fails!
Some tests fail!!

Test time = 3.82 sec
Test Failed.
"ScaleMicrogrid" end time: Jul 10 14:10 EDT
"ScaleMicrogrid" time elapsed: 00:00:03
This was performed with the https://github.com/coin-or/Ipopt/issues/12 release of GridKit, Sundials 6.7.0, Ipopt 3.14.16, and SuiteSparse 5.10.1.

reid-g:
Some updates on this. I have verified that all model parameters are correct and rhs outputs are correct. So these are not it.

The "true" solution vectors I generate are from MATLAB based ODE form of the model. Since no analytical solution to the model available . No Jacobian is used. This is done with extreme tolerances.  Relative Tolerance `1e-14` and Absolute Tolerance `1e-14`. I have done both `ode15s` and `ode23tb` solutions. I have done also the new MATLAB integrated `IDA` way as well. They all match up correctly with only some numerical rounds difference between.

This same error appears in hardwired setups (no GridKit) as well. This is for both ODE and DAE forms hardwired.

Still actively looking to see what is the issue.

pelesh:
Unfortunately, #30 did not fix this issue. See this [log](https://github.com/ORNL/GridKit/actions/runs/12298699657/job/34322738496?pr=36).

nkoukpaizan:
The test failure seems to be non-deterministic. Different versions of SUNDIALS and/or different machines produce different error norms. 

For example, from my runs on Frontier:
```
Test the Relative Error for N = 2
2-Norm Relative Error: 1.85946e-06
Test with Nsize = 2 passes!
Test the Relative Error for N = 4
2-Norm Relative Error: 7.05324e-06
Test with Nsize = 4 passes!
Test the Relative Error for N = 8
2-Norm Relative Error: 6.77802e-06
Test with Nsize = 8 passes!
All tests pass!!
```

But from the Github actions:
```
12: Test the Relative Error for N = 2
12: 2-Norm Relative Error: 2.80821e-06
[...]

PR #19, opened 2024-08-06
Fix #17

### LINKED ISSUE #17: Ipopt 3.14.16 Error
There is an error with tests 13, 14, and 16 with the latest version of Ipopt (3.14.16). Changing the return value of eval_jac_g (line 212) in DynamicObjective.cpp to "true" fixes these errors.

pelesh:
Optimization model functions return `true` even when empty. Originally, these functions were set to return `false` to signify they don't do anything but it was a poor design choice. Ipopt may call these functions even if they are empty, there should not be a reason to report a failure in that case.


Paulm991:
Looks good to me. All tests pass as expected. Only ScaleMicrogrid fails now, as mentioned in issue #16.

Paulm991 (commented):
Looks good to me. All tests pass as expected. Only ScaleMicrogrid fails now, as mentioned in issue #16 .

PR #22, opened 2024-08-08
Added callback function to IDA

### LINKED ISSUE #21: GridKit interface with IDA can't return values
Tests that are constantly collecting data over time don't have an easy way to obtain this data from the Ida.cpp file. To measure responses from perturbations, it is nice to be able to get the data generated at each time step.

In my tests on the genrou-test-2 branch, I used a callback function that returned the timestep and y vector after every time step. This was used by the GenrouTest.cpp file to print out each y value to a file.

Paulm991:
Added callback functionality as mentioned in issue #21 . All tests appear to pass as normal. The addition is in the runSimulation function in IDA.

### DIFF
--- a/Solver/Dynamic/Ida.cpp
+++ b/Solver/Dynamic/Ida.cpp
@@ -284,6 +284,10 @@ namespace Sundials
             checkOutput(retval, "IDASolve");
             //printOutput(tout); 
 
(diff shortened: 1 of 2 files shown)


pelesh:
Closing in favor of #91.

PR #23, opened 2024-08-09
Add Enzyme AD support

AVanDeBrook:
This PR adds a script to pull down specific releases of LLVM and Enzyme and build local versions of both.

On my machines LLVM compiles without issue, however, at run time it depends on `gcc` and/or `g++` headers. I found that `gcc-12` and `g++-12` provide the most stability with `llvm-16`.

With the versions pinned in the shell script, Enzyme should compile without issue and the tests will output 2 errors; these errors do not affect the functionality of Enzyme.

Programs can be compiled like the following:
```
clang test.c -fplugin=.enzyme_build/Enzyme/enzyme/build/Enzyme/ClangEnzyme-16.so
```

*Note(s)*: 
* Enzyme has had a recent burst in development, so the most current version is ahead of the one I've tested and verified
* This uses LLVM version 16.0.0, Enzyme version 0.0.133, and depends on GCC/G++ version 12

### DIFF
--- a/enzyme_support.sh
+++ b/enzyme_support.sh
@@ -1,0 +1,118 @@
+# This file pulls down specific verisons of enzyme and LLVM and compiles them. Use the clang binary
+# that is compiled locally with the enzyme LLVM/Clang plugin, this will guarantee that the plugin
+# works.
+#
+# This version of the enzyme plugin has been tested with the clang-16 available through the apt
+# package manager (does not work correctly with the enzyme plugin) and a locally compiled version
+# (does work).
+#
+# Notes:
+# - For the cmake build specify the location of the local clang binaries like so:
+# `CC=.enzyme_build/llvm-project/build/bin/clang CXX=.enzyme_build/llvm-project/build/bin/clang++ cmake ...`
+#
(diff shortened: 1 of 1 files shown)


reid-g:
Running the [enzyme_support.sh](https://github.com/AVanDeBrook/GridKit/blob/enzyme-support/enzyme_support.sh) script starting from nothing only clones in the `Enzyme` repo. The `llvm` repo does not clone. Then crashes since there is no `llvm` repo.

stonecoldhughes:
The two Dockerfiles below and associated shell scripts ilustrate 2 builds of Enzyme that pass all or most of the Enzyme builtin tests: "check-enzyme-integration" and "check-enzyme". One dockerfile passes one set of tests, and the other passes the other set of tests.

[resolve_enzyme_1.txt](https://github.com/user-attachments/files/17174561/resolve_enzyme_1.txt)
[resolve_enzyme_0.txt](https://github.com/user-attachments/files/17174562/resolve_enzyme_0.txt)
[run_1.txt](https://github.com/user-attachments/files/17174563/run_1.txt)
[build_1.txt](https://github.com/user-attachments/files/17174564/build_1.txt)
[build_0.txt](https://github.com/user-attachments/files/17174565/build_0.txt)
[run_0.txt](https://github.com/user-attachments/files/17174566/run_0.txt)

alexander-novo:
Since we've moved forward with a different branch, I'm closing this.

PR #29, opened 2024-10-30
Enzyme support example

nkoukpaizan:
This builds on earlier work from Asher and Slaven to add examples using the Enzyme automatic differentiation library. 
There are two new tests compared to `develop` that should pass.

Building Enzyme from source is discussed in #23. I tested that approach as well as building LLVM and Enzyme through Spack, and both approaches work for this use-case.

### DIFF
--- a/CMake/FindEnzyme.cmake
+++ b/CMake/FindEnzyme.cmake
@@ -1,0 +1,186 @@
+# 
+# Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+# Produced at the Lawrence Livermore National Laboratory.
+# Written by Slaven Peles <peles2@llnl.gov>.
+# LLNL-CODE-718378.
+# All rights reserved.
+# 
+# This file is part of GridKit™. For details, see github.com/LLNL/GridKit 
+# Please also read the LICENSE file. 
+# 
+# Redistribution and use in source and binary forms, with or without 
+# modification, are permitted provided that the following conditions are met:
+# - Redistributions of source code must retain the above copyright notice, 
+#   this list of conditions and the disclaimer below.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the disclaimer (as noted below) in the 
+#   documentation and/or other materials provided with the distribution.
+# - Neither the name of the LLNS/LLNL nor the names of its contributors may 
+#   be used to endorse or promote products derived from this software without 
+#   specific prior written permission.
+# 
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+# CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+# INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL LAWRENCE LIVERMORE NATIONAL
+# SECURITY, LLC, THE U.S. DEPARTMENT OF ENERGY OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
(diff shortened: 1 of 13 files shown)


pelesh:
CC @ashermancinelli

ashermancinelli:
Very cool to see!

CMake/FindEnzyme.cmake
```diff
@@ -0,0 +1,194 @@
...
+    OUTPUT ${PHASE2}
+    COMMAND ${GRIDKIT_LLVM_LINK} ${OBJS} -o ${PHASE2}
+    COMMENT \"Linking object files to LLVM bytecode for target ${enzyme_add_executable_NAME}\"
+    )
+  ")
+
+  add_custom_command(
+    DEPENDS ${PHASE2}
+    OUTPUT ${PHASE3}
+    COMMAND ${GRIDKIT_OPT} ${PHASE2} -load-pass-plugin=${ENZYME_LLVM_PLUGIN_LIBRARY} -passes=enzyme -o ${PHASE3} -S
```
ashermancinelli, line 173: If you don't want such a manual pipeline, you can use the plugin registration hooks that the opt pass plugin api provides (like `registerOptimizerLastEPCallback`, for example). [See the docs here](https://llvm.org/doxygen/classllvm_1_1PassBuilder.html). Sorry I left you with such cursed cmake! 😅  You could load the plugin and use a command like this:
```
clang++ -fpass-plugin=LLVMEnzyme.so -mllvm '-passes=default<O1>,enzyme,default<O3>' -o ./a.out
```

pelesh: So, in CMake configuration you would set something like `-DCMAKE_CXX_FLAGS="-fpass-plugin=LLVMEnzyme.so" -mllvm '-passes=default<O1>,enzyme,default<O3>'`?

ashermancinelli: Exactly. Instead of dumping the IR between phases, you can first register the enzyme pass as a plugin and then use the `-mllvm -passes=` argument to build a custom phase ordering so enzyme is thrown in there.

CMake/FindEnzyme.cmake
```diff
@@ -0,0 +1,194 @@
...
+  set(options)
+  set(oneValueArgs NAME)
+  set(multiValueArgs SOURCES LINK_LIBRARIES)
+  cmake_parse_arguments(enzyme_add_executable "${options}" "${oneValueArgs}"
+    "${multiValueArgs}" ${ARGN})
+
+  set(PHASE2 "${CMAKE_CURRENT_BINARY_DIR}/${enzyme_add_executable_NAME}.bc")
+  set(PHASE3 "${CMAKE_CURRENT_BINARY_DIR}/${enzyme_add_executable_NAME}_enzyme.ll")
+  set(PHASE4 "${CMAKE_CURRENT_BINARY_DIR}/${enzyme_add_executable_NAME}_opt.ll")
+  set(PHASE5 "${CMAKE_CURRENT_BINARY_DIR}/${enzyme_add_executable_NAME}")
```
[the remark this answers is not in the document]
nkoukpaizan: I have to admit I am not familiar with `ninja`. All my test were with `make`. Could you expand on the way to fix it for `ninja`?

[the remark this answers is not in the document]
nkoukpaizan: Should be fixed by [20470b6](https://github.com/ORNL/GridKit/pull/29/commits/20470b6d9980cfa18ef54239b3b7431da53cafd4)

ashermancinelli (approved):
Looks good! Let me know if you run into issues with clang/phase ordering/etc. Good luck!

reid-g (approved):

Examples/Enzyme/Library/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+enzyme_add_executable(NAME EnzymeLibCheck SOURCES main.cpp library.cpp)
+#install(TARGETS ${CMAKE_CURRENT_BINARY_DIR}/EnzymeLibCheck DESTINATION bin)
+
```
pelesh, line 3: Do we want to install Enzyme targets or this is out of scope for this PR?

Examples/Enzyme/Standalone/main.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+#include <limits>
+
+double square(double x) {
+  return x * x;
+}
+
```
pelesh, line 7: Should this be hard-wired `double` or configurable `real_type`?

pelesh (approved):
I suggest we merge this as it has basic infrastructure and tests for Enzyme support. We can expand testing and documentation through subsequent PRs.

I left a couple of nitpicking comments.

PR #30, opened 2024-12-02
BugFix: Scale Microgrid Error

reid-g:
This pull request fixes issue #16 .

Changed the error estimation to norm based instead of element wise. All test cases pass now. Solution vectors from the same system solved as a DAE are provided (from MATLAB).

Numerical differences in elements of the vector seems to be bottoming out any improvements on error. Step size drops to be quite small initially as well (h_0 < reltol). However, as discussed this does not seem to be an issue.


nkoukpaizan (commented):
I am a little worried that moving to the two norm is not fixing the underlying issue and may give a false-negative. If we expect the vectors to be close element-wise, then that should be the right thing to check for the test.

Examples/ScaleMicrogrid/SolutionKeys.hpp
```diff
@@ -4,8 +4,11 @@
  * @author Slaven Peles (peless@ornl.gov)
  * @brief Answer keys for Scaled Microgrid test with Nsize = 2, 4, 8
  * 
- * All data generated with Matlab ode23tb solver with tolerances set to
- * abstol=1e-12 and reltol=1e-12.
+ * Data generated with Matlab ode23tb solver with tolerances set to
+ * abstol = 1e-12 and reltol = 1e-12 for the ODE derivation.
+ * 
+ * DAE keys are generated with Matlab IDA interface with tolerances set to
+ * abstol = 1e-10 and reltol = 1e-10 for the DAE derivation.
```
pelesh, line 11: Which one of the two is true?

Examples/ScaleMicrogrid/SolutionKeys.hpp
```diff
@@ -515,3 +518,512 @@ const std::vector<real_type> answer_key_N8 = {29266.6517718661,
...
+    -1.736356080586935e+02,
+     3.020632590356075e+02,
+    -1.715847733944154e+02,
+     3.158572249521520e+02,
+    -1.757198561909239e+02,
+     3.062813650746419e+02,
+    -1.707987479996116e+02,
+     3.220107136834399e+02,
+    -1.761933472176073e+02
+};
```
pelesh, line 1029: End file with newline.

pelesh (changes requested):
The tolerances in the example need to be better documented. Use Doxygen `@todo` keyword to document issues we have not clarified yet.

pelesh:
> I am a little worried that moving to the two norm is not fixing the underlying issue and may give a false-negative. If we expect the vectors to be close element-wise, then that should be the right thing to check for the test.

SUNDIALS is using [weighted 2-norm](https://sundials.readthedocs.io/en/latest/idas/Mathematics_link.html#equation-idas-bdf) to compare with relative tolerance. Ideally, we should use the same. For this PR, it should be at least documented where these tolerances come from, why particular numbers are used and why we believe this is a reasonable verification of results.

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
```diff
@@ -327,16 +334,27 @@ int test(index_type Nsize, bool debug_output)
...
     // check relative error
-    std::cout << "Test the Relative Error\n";
+    std::cout << "Test the Relative Error for N = " << Nsize << "\n";
     for (index_type i = 0; i < true_vec->size(); i++)
     {
-        test_pass *= GridKit::Testing::isEqual(yfinal[i], true_vec->at(i), tol);
+        // test_pass *= GridKit::Testing::isEqual(yfinal[i], true_vec->at(i), error_tol);
+        //Produces the Elementwise Relative Error
         if (debug_output)
             std::cout << i << " : " << abs(true_vec->at(i) - yfinal[i]) / abs(true_vec->at(i)) << "\n";
```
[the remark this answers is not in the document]
pelesh: I think the latter is safer measure of discrepancy between two floating point scalars. If safeguards for cases where $`a t_i \sim 0`$.

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
```diff
@@ -24,14 +24,19 @@ using real_type = double;
...
 {
     int retval = 0;
-    retval += test(2);
-    retval += test(4);
-    retval += test(8);
+    bool debug_out = false;
+    bool DAE_keys = false;
+
+    real_type error_tol = 8e-5;
+
```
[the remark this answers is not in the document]
pelesh: By the way, all tolerances should be defined in `SolutionKeys.hpp` file because they depend on how these values in the solution key are generated. I would implement it like:
 ```suggestion
    real_type error_tol = MICROGRID_TOL;
```
and define `MICROGRID_TOL` in the `SolutionKeys.hpp` file. There should be an explanation why you picked `8e-5` and not `7e-5`, for example. The explanation should be commented in the code using Doxygen format.

**Comment by reid-g** [comment]
> > I am a little worried that moving to the two norm is not fixing the underlying issue and may give a false-negative. If we expect the vectors to be close element-wise, then that should be the right thing to check for the test.
> 
> SUNDIALS is using [weighted 2-norm](https://sundials.readthedocs.io/en/latest/idas/Mathematics_link.html#equation-idas-bdf) to compare with relative tolerance. Ideally, we should use the same. For this PR, it should be at least documented where these tolerances come from, why particular numbers are used and why we believe this is a reasonable verification of results.

The weighted 2-norm used is an estimate of temporal error. Since we have a reference solution to compare then the raw temporal accuracy measure I think should be used (standard 2norm relative error).

It was my mistake to originally have the element-wise relative error setup. The methods accuracy is bounded from analysis by the 2norm. Not the element-wise.

File: Examples/ScaleMicrogrid/ScaleMicrogrid.cpp (hunk 4 above)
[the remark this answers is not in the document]
Comment (reid-g) [reply Examples/ScaleMicrogrid/ScaleMicrogrid.cpp:36]: This was roughly the best error obtained.
The `N=8` case hovers around 1e-5 to 3e-5 when `reltol=1e-7; abstol=1e-7`. This the point where the error on the `N=8` system "bottoms out". Higher and lower tolerances increase error. `error_tol` was chosen based of this.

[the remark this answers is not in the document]
Comment (pelesh) [reply Examples/ScaleMicrogrid/ScaleMicrogrid.cpp:37]: I think how you came up to the tolerance used for testing needs to be documented in the code. The argument for selecting that specific tolerance is unconvincing. Eventually we will want to improve this test and it would be good to have documented what has been done to this point.

Also, I would put numerical value for the tolerance and the comments in the `SolutionKeys.hpp` file.

File: Examples/ScaleMicrogrid/ScaleMicrogrid.cpp [hunk 5]
```diff
@@ -49,26 +55,31 @@ int main(int argc, char const *argv[])
...
  * @return int returns 0 if successful, >0 otherwise
  */
-int test(index_type Nsize, bool debug_output)
+int test(index_type Nsize, real_type error_tol, bool debug_output, bool use_DAE_keys)
 {
-    real_type abstol = 1.0e-8;
-    real_type reltol = 1.0e-8;
+    real_type abstol = 1.0e-7;
+    real_type reltol = 1.0e-7;
     index_type max_step_number = 3000;
```
pelesh, line 62: This also needs to be documented. Where these numbers come from?

Examples/ScaleMicrogrid/SolutionKeys.hpp
pelesh: There is only one set of data with reference results. Comments in the code should describe only the data that is included.

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
```diff
@@ -327,16 +335,26 @@ int test(index_type Nsize, bool debug_output)
...
+    real_type sumbottom = 0.0;
 
     // check relative error
-    std::cout << "Test the Relative Error\n";
+    std::cout << "Test the Relative Error for N = " << Nsize << "\n";
     for (index_type i = 0; i < true_vec->size(); i++)
     {
-        test_pass *= GridKit::Testing::isEqual(yfinal[i], true_vec->at(i), tol);
+        //Produces the Elementwise Relative Error
         if (debug_output)
```
pelesh, line 347: Is this comment still correct?

Examples/ScaleMicrogrid/SolutionKeys.hpp
reid-g: ??? There are two set of keys. ODE keys are denoted `answer_key_N*` and DAE keys are denoted `answer_key_N*_DAE`. Both reference data sets are in the solution keys file.

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
reid-g: Added documentation on numbers.

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
reid-g: fixed

Examples/ScaleMicrogrid/ScaleMicrogrid.cpp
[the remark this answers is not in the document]
reid-g: Added documentation on why it was chosen. Results are commented in showing why as well.

pelesh (approved):

PR #31, opened 2024-12-02
Add clang format file

### LINKED ISSUE #26: Create a `.clang-format` file
It would be good to have a centralized authoritative styling document for auto-formatting in an editor.

It would be worth looking into a styling that matches how the codebase is already written (if possible) to reduce the amount of autoformatting changes.

alexander-novo:
I added options in order of:

1) How they appear in the [style guide](https://github.com/ORNL/GridKit/pull/5)

2) How they appear in the current code

3) If there are conflicts in the current code, then I picked the option I prefer

There currently isn't a formatting option for forcing a newline between includes and the rest of the file, however there is [a pull request](https://github.com/llvm/llvm-project/pull/78957) to add this feature, so it may exist in a future version.

Please see #32 for the current up to date formatting results of the file.


alexander-novo:
I've tested that this works with the current version of `clang-format` shipped with Ubuntu Lunar - version `15.0.7`. I don't know what versions we'd want to support, maybe others can comment what version of `clang-format` they have, and whether the file works for them.

There is [a new option added](https://github.com/llvm/llvm-project/pull/108241) in (I believe) the latest version - version `20.0.0`, which prevents function declarations from being aligned. I don't know exactly how much this will impact the codebase?

alexander-novo:
@pelesh I looked into the issue of putting system header includes before local includes. It turns out that `clang-format` identifies the "main header" of a source file (e.g. `Capacitor.hpp` for `Capacitor.cpp`) and separates this include into its own group before applying the regular expressions and grouping to every other include. The main header is put in priority 0, which is why it was showing up before every other include. This seems like a useful idea to me, but let me know which way you would like it. I could:

1) Leave the "main header" at the top, followed by system headers, followed by libraries and local includes
2) Put the system headers at the top, but leave the "main header" in its own group
3) Put the system headers at the top, and put the "main header" in the same group as local includes. From my testing it looks like this seems to put the "main header" at the top of this group.

pelesh:
Closing in favor of #68.

PR #35, opened 2024-12-10
Forward compatibility with SUNDIALS

pelesh:
Ensure GridKit is forward compatible with SUNDIALS, meaning it will require SUNDIALS version >= 7.0. There is no need in supporting older SUNDIALS versions in GridKit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,7 @@ if(GRIDKIT_ENABLE_IPOPT)
   include(FindIpopt)
 endif()
 if(GRIDKIT_ENABLE_SUNDIALS)
-  find_package(SUNDIALS 6.0.0 REQUIRED CONFIG 
+  find_package(SUNDIALS 7.0.0 REQUIRED CONFIG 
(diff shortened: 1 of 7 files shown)


pelesh:
> Tested and seems to work.
> 
> I suggest adding the constraint for sundials>7.x to the CMakeLists.txt as done [here](https://github.com/ORNL/GridKit/blob/65523d2e45b065343df33c6fc39c98e276442e05/CMakeLists.txt#L131). The [current implementation](https://github.com/ORNL/GridKit/blob/32eaf04b5c2f6d61e120fa8854108823adbbca35/CMakeLists.txt#L127) is asking for 6.0.0, and that's not being enforced.

Good catch, thanks! Fixed in [7920475](https://github.com/ORNL/GridKit/pull/35/commits/7920475ec9f86fec8a20192f54abed29a5b6c8b3) and tested.

alexander-novo:
A bit late to the party, but I can also confirm that I can compile and run examples with Sundials >= 7.0.0

PR #41, opened 2025-01-13
Fixing for file IO Parameter Estimation Example

reid-g:
- Crashes for file IO when the file is not found
 - ParameterEstimation can be run outside of its own directory without error


pelesh:
Closing in favor of #44.

PR #44, opened 2025-01-18
Fixes input file issue with parameter estimation example

pelesh:
Parameter estimation example crashes when input file cannot be found. Use a string to pretend to be a file in this example, so it can be run as a standalone binary without special input files. User still can easily modify this file to create their own app.

In this PR:
- `setLookupTable` function takes `std::istream` and not input file name as the input. It is user's responsibility to create an `std::istream`.
- Some compiler warning fixes.


Utilities/FileIO.hpp
```diff
@@ -251,13 +252,24 @@ void readMatPowerBaseMVA(SystemModelData<RealT, IdxT>& mp, std::string& line)
 
 namespace GridKit
 {
+
+/**
+ * @brief Reads a file in for tabulated data
+ *
```
reid-g, line 258: It's not too clear what this functions expects as the files data format. Perhaps clarify here on this.

Utilities/FileIO.hpp
```diff
@@ -251,13 +252,24 @@ void readMatPowerBaseMVA(SystemModelData<RealT, IdxT>& mp, std::string& line)
...
+ * @param filename 
+ * @param ti 
+ * @param tf 
+ */
 template <typename ScalarT>
 void setLookupTable(std::vector<std::vector<ScalarT>>& table,
-                    std::string filename,
+                    std::istream& idata,
                     ScalarT& ti,
                     ScalarT& tf)
```
reid-g, line 271: Are these references necessary to ti and tf? The tables format provides them in the first and last locations.

Examples/ParameterEstimation/ParameterEstimation.cpp
```diff
@@ -60,21 +60,25 @@
...
-#include <Solver/Optimization/DynamicObjective.hpp>
-#include <Solver/Optimization/DynamicConstraint.hpp>
 #include <Utilities/FileIO.hpp>
 #include <Utilities/Testing.hpp>
 
+#include "lookup_table.hpp"
+
 
-int main(int argc, char** argv)
+int main(int /* argc */, char** /* argv */)
```
reid-g, line 81: Remove input args since no input is handled anymore.

reid-g (changes requested):
Overall, looks good. Only some minor changes then it is good to go.

reid-g (approved):

Utilities/FileIO.hpp
pelesh: Updated comments to clarify that.

Utilities/FileIO.hpp
pelesh: I agree it may look redundant but I think these times should be part of the function signature. That way user does not need to know about the structure of the input file and/or how initial/final times are evaluated (e.g., at some point we might want to interpolate time series in the middle of data range).

Examples/ParameterEstimation/ParameterEstimation.cpp
pelesh: Done!

PR #46, opened 2025-01-22
Fix lookup install and downgrade to LLVM 16

nkoukpaizan:
#44 moved `Examples/ParameterEstimation/lookup_table.dat` to `Examples/ParameterEstimation/lookup_table.hpp`, but was still trying to install `lookup_table.dat` in the `CMakeLists.txt.`. Surprised CI didn't catch it.

This also downgrades LLVM from 18 to 16 in CI. While Gridkit tests currently work with later versions of LLVM, many tests fail within Enzyme itself. (Enzyme CI currently tests LLVM 15 and 16.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ You should have all of the following installed before installing GridKit™
 	- [Suitesparse](https://github.com/DrTimothyAldenDavis/SuiteSparse) >= 5.x (optional)
 	- [Ipopt](https://github.com/coin-or/Ipopt) >= 3.x (optional)
     - [Enzyme](https://github.com/EnzymeAD/Enzyme) >=0.0.131 (optional)
-        - [LLVM](https://github.com/llvm/llvm-project) 16.x 
+        - [LLVM](https://github.com/llvm/llvm-project) >= 15.x 
 - [CMake](https://cmake.org/) >= 3.12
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ You should have all of the following installed before installing GridKit™
 	- [Suitesparse](https://github.com/DrTimothyAldenDavis/SuiteSparse) >= 5.x (optional)
 	- [Ipopt](https://github.com/coin-or/Ipopt) >= 3.x (optional)
     - [Enzyme](https://github.com/EnzymeAD/Enzyme) >=0.0.131 (optional)
-        - [LLVM](https://github.com/llvm/llvm-project) >= 18.x 
+        - [LLVM](https://github.com/llvm/llvm-project) 16.x 
 - [CMake](https://cmake.org/) >= 3.12
```
pelesh, line 22: I would mirror what Enzyme team specifies for the dependency here:
```suggestion
        - [LLVM](https://github.com/llvm/llvm-project) >= 15.x 
- [CMake](https://cmake.org/) >= 3.12
```
We could add a note we are testing with v16.

pelesh (approved):
Looks good. I got only one minor suggestion.

PR #51, opened 2025-01-28
Document requirement of KLU support in SUNDIALS.

ianmtaylor1:
See issue https://github.com/ORNL/GridKit/issues/49. Note the requirement for KLU support and link to instructions in SUNDIALS documentation.


pelesh:
@ianmtaylor1, GridKit can be built with SUNDIALS does not have KLU support enabled, but if you request SuiteSparse support, then SUNDIALS has to support it, as well (i.e. it has to have KLU module built).

ianmtaylor1:
Thanks for the clarification!

PR #56, opened 2025-02-11
Rename namespace ModelLib -> GridKit.

pelesh:
Use `GridKit` name space for all models rather than generic `ModelLib`. This is a wholesale name change of the existing namespaces.


nkoukpaizan (approved):
Looks good to me. Looked through ModelLib -> GridKit changes in the files and tested on Frontier.

PR #32, opened 2024-12-02
Initial formatting with clang format file

alexander-novo:
I only minimally edited some files manually, such as [`PowerSystemData.hpp`](https://github.com/ORNL/GridKit/pull/32/files#diff-f55f4baa33985c59559428b3a28e74528e276acf3cd0af1f92b340cd682c8921R41)

PR #58, opened 2025-02-17
Fix include issue with Branch.hpp (#57)

### LINKED ISSUE #57: `Branch.hpp` doesn't include `PowerSystemData.hpp` even though it uses a type from that header
Not currently causing a mis-compilation because the only things including that header happen to also include `PowerSystemData.hpp` *before* `Branch.hpp`. But after auto-formatting (#32) the order of these includes change and it becomes a mis-compilation.

alexander-novo:
Fixes #57 

Also brought some adjacent includes in line with style guidelines

### DIFF
--- a/src/Model/PowerFlow/Branch/Branch.hpp
+++ b/src/Model/PowerFlow/Branch/Branch.hpp
@@ -62,9 +62,15 @@
 
 #include <ModelEvaluatorImpl.hpp>
 
+// Forward declarations.
 namespace GridKit
 {
     template <class ScalarT, typename IdxT> class BaseBus;
+
+    namespace PowerSystemData {
+        template <class ScalarT, typename IdxT>
+        struct BranchData;
+    }
 }
 
 namespace GridKit


alexander-novo:
This should fix the mis-compilation being experienced in the CI of #32

pelesh (changes requested):
Power flow branch header `Branch.hpp` is missing forward declaration to `BranchData` class. Adding the forward declaration should fix this issue. 

The currently proposed solution introduces additional build dependencies that would be good to avoid.

alexander-novo:
> introduces additional build dependencies that would be good to avoid

@pelesh for future reference can you explain what you mean by this?

pelesh:
> > introduces additional build dependencies that would be good to avoid
> 
> @pelesh for future reference can you explain what you mean by this?

In header files use forward declarations instead of including other headers whenever possible.

alexander-novo:
Changed to adding the forward declaration instead of including the definition. In the process I realized the phasor dynamics implementation already had this forward declaration, so we should be all good there.

@pelesh this forward declaration philosophy might be a good thing to put in the developer guidelines, since it seems like an easy thing to catch people unaware.

pelesh (approved):
Good to merge, thanks!

PR #61, opened 2025-02-20
Reorganize model evaluator implementations

pelesh:
Reorganize model evaluator interface implementations for different families of models in a following way:
- Pure virtual class `Model::Evaluator`, which defines interfaces to simulation models is in the file `src/Model/Evaluator.hpp`.
- Specific implementations of `Model::Evaluator` class are separate for each model family:
    - `PhasorDynamics::BusBase` and `PhasorDynamics::Component` for phasor dynamics components,
    - `CircuitComponent` for power electronics components, and
    - `ModelEvaluatorImpl` for power flow components.


abirchfield (approved):
Looks good to me

PR #67, opened 2025-02-28
pre-commit Github action

nkoukpaizan:
Setup to apply pre-commit fixes in Github actions.

This is currently set to apply changes automatically when necessary. However, for security reason, this then disables further checks unless using tokens (see https://github.com/EndBug/add-and-commit?tab=readme-ov-file#about-actionscheckout). Adding another commit ourselves should then re-instate checks. There potentially are better ways to handle this.

Another option is to not apply the changes automatically. The pre-commit pipeline will then be reported as failed and it will be on the owner of the pull request to fix it.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+repos:
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v19.1.7
```
alexander-novo, line 3: We should pick a version that we expect people to install with Enzyme, like v15 or so.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+    hooks:
+    - id: clang-format
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+    - id: check-toml
+    - id: forbid-new-submodules
+    - id: end-of-file-fixer
+    - id: check-yaml
```
alexander-novo, line 12: I'm curious - will these `check-` hooks actually do anything? For instance, if I make a new submodule and then submit a PR will it fail CI or?

nkoukpaizan: Correct, the check will work as you describe. You can see what it's trying yo do in CI (e.g. https://github.com/ORNL/GridKit/actions/runs/13594004211/job/38006645346#step:4:154). These are mostly legacy checks. We can remove anything we don't need.

pelesh:
Closing in favor of #68.

PR #68, opened 2025-03-04
Set up pre-commit hook for clang-format

### LINKED ISSUE #26: Create a `.clang-format` file

### LINKED ISSUE #33: Add pre-commit actions
We should add `clang-format` and any linters to pre-commit actions. There are a variety of ways that we can implement these actions, but these are the three ways forward I can see:

1) We make our own `pre-commit` hooks in Git's hook system.
- Pros: No extra dev dependencies than the programs we actually want to use in the script
- Cons: Git hooks aren't pushed, so we must distribute them in the repository along with a script that installs them. As well, any scripts we wish to support would limit the developer environment to being able to run those scripts (i.e. we may force developers to use bash)
2) We use the [pre-commit](https://pre-commit.com/) python framework, along with pre-made hooks for whatever actions we want
- Pros: Very little development effort - it seems like there are pre-developed hooks for a lot of what we want (formatting, linting, warning on committing to a protected branch, etc.). All of the dependencies for hooks will be automatically installed. Very cross-platform.
- Cons: Potentially very many dependencies, since each hook will be an extra dependency, and each of them may have sub-dependencies, even if the hook is simple and the software needed for the hook is already pre-installed (if `clang-format`, for instance, is already installed, a `clang-format` hook would install a separate version of `clang-format` specifically for the hook). Python also is added as a dependency.
3) We use the pre-commit python framework, and make our own hooks.
- Pros: Few dependencies, ease of use
- Cons: Python dependency. The pre-commit framework doesn't seem to be designed around small inline hooks, and rather expects you to install hooks on the fly from a separate repository. If we want to make our own hooks, it may be necessary to write some amount in Python

It's also worth noting that all of the above options are entirely opt-in... since Git hooks don't get pushed to the repository, there must be some additional effort put in by the developer to enable these hooks, so we can't *depend* on developers enabling these hooks. These hooks should only be considered, then, to be warnings or handy conveniences. If we want actual enforcement, then we need to actually put them in CI (#25)

In my opinion, I'm leaning towards option 2 (or 3 if we don't want extra dependencies) since I think we're going to be adding Python as somewhat of a dev dependency for spack anyway (#24).

pelesh:
- Set up:
    - `.clang-format` configuration
    - `pre-commit` github actions
- Reformat code per guidelines in CONTRIBUTING.md
- Includes prior work of @alexander-novo (PRs #31 and #32) and @nkoukpaizan (PR #67).

Closes #26
Closes #33


examples/DistributedGeneratorTest/DGTest.cpp
```diff
@@ -1,85 +1,84 @@
```
nkoukpaizan, line 1: Since we are checking for lines at the end of files and extra lines between code statements, do we want to suppress extra lines at the beginning of files as well? Do we care?

src/Model/PhasorDynamics/Branch/Branch.cpp
```diff
@@ -2,48 +2,49 @@
...
  * @brief Definition of a phasor dynamics branch model.
- * 
+ *
  * The model uses Cartesian coordinates.
- * 
+ *
  */
 
-#include <iostream>
+#include "Branch.hpp"
```
nkoukpaizan, line 10: Is this the desired behavior? `#include "*hpp"` before system includes, e.g., (`#include <iostream>`). I would intuitively expect the local includes to be last.

nkoukpaizan (approved):
A couple of comments. Otherwise, looks good to me.

alexander-novo (approved):

examples/DistributedGeneratorTest/DGTest.cpp
pelesh: We don't want extra lines lines at the beginning of files. We should probably put doxygen file documentation at the top of source files.

Issue #65, opened 2025-02-27
Compiler warnings in PhasorDynamics tests


pelesh:
Compiler produces warnings when building PhasorDynamics tests, e.g.,
```
In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/tests/UnitTests/PhasorDynamics/runBranchTests.cpp:1:
In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/tests/UnitTests/PhasorDynamics/BranchTests.hpp:6:
In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/src/Model/PhasorDynamics/Branch/Branch.hpp:11:
/lustre/orion/scratch/nkouk/stf006/Codes/GridKit/src/Model/PhasorDynamics/Component.hpp:51:22: warning: 'nnz' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        virtual IdxT nnz()

In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/tests/UnitTests/PhasorDynamics/runBranchTests.cpp:1:
In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/tests/UnitTests/PhasorDynamics/BranchTests.hpp:6:
In file included from /lustre/orion/scratch/nkouk/stf006/Codes/GridKit/src/Model/PhasorDynamics/Branch/Branch.hpp:11:
/lustre/orion/scratch/nkouk/stf006/Codes/GridKit/src/Model/PhasorDynamics/Component.hpp:56:22: warning: 'hasJacobian' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        virtual bool hasJacobian()
```

_Originally posted by @nkoukpaizan in https://github.com/ORNL/GridKit/pull/63#pullrequestreview-2641651293_

nkoukpaizan:
Addressed in #71. Closing.

PR #71, opened 2025-03-12
Add GENROUwS model for a synchronous machine

pelesh:
Added GENROUwS model for the synchronous machine and a 2-bus example with the generator connected to an infinite bus by a transmission line. The example checks its results against Powerworld and runs on the CI pipeline. This provides some basic verification of the model. 

Cloases #65.

What is outside the scope of this PR:
- Development of detailed unit testing for `PhasorDynamics::Genrou` model. See #78 
- Updating `PhasorDynamics::Genrou` documentation. See #79


pelesh:
CC @reid-g @abdourahmanbarry @alexander-novo

abirchfield (approved):
Looks good. Some of the comments on GENROU haven't been updated from branch but those can be quickly fixed.

reid-g (approved):
Looks good to me.
Error in the example seems higher then expected since default `reltol =  1e-7`. However, this seems due to how discontinuities are handled though.

alexander-novo (approved):

PR #85, opened 2025-04-19
Unit test and example for `Genrou` class.

### LINKED ISSUE #78: Create initial unit test for `PhasorDynamics::Genrou` model
Implement unit tests for `PhasorDynamics::Genrou` model, using the existing unit test [placeholder](https://github.com/ORNL/GridKit/blob/genrou-adam-dev/tests/UnitTests/PhasorDynamics/GenrouTests.hpp). Implement tests for
- [x] constructor
- [ ] residual evaluation
- [ ] initialization

Unit tests do not need to run any simulation, they just need to verify for correctness methods in `PhasorDynamics::Genrou` class.

pelesh:
## Description
 
The 6th order electrical machine model implemented in `Genrou` class has been tested in an [example](https://github.com/ORNL/GridKit/blob/develop/examples/PhasorDynamics/Example1/example1.cpp) where results of a simulation for a 2-bus system are compared to Powerworld simulations of the same system, comparing the results to within 3-4 significant digits. That example does not test `Genrou` methods to the machine precision, which is needed for models of such complexity. This PR adds a unit test for the model residual evaluation function and another example comparing results to Powerworld reference.
 
 Closes #78
 
 ## Proposed changes
 
 Add a unit test that initializes the `Genrou` model at a consistent DAE solution and then verifies that all residual vector elements are zero to within machine precision. In addition, a 3-bus example is added which verifies simulation result to the numerical integration tolerance. 
 
 ## Checklist
 

- [x] All tests pass.
 
 ## Further comments
 
This requires a little bit more work before it is ready for review.


abirchfield (approved):
Everything looks good. Compiled and ran example 2 on my machine with no issues.

pelesh:
@abirchfield, the `Load` component model test for residual evaluation fails in this branch:

```
test 24
      Start 24: PhasorDynamicsLoadTest

24: Test command: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics/test_phasor_load
24: Working Directory: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics
24: Test timeout computed to be: 10000000
24: --- PASS: Test constructor
24: --- FAIL: Test residual
24: 
24: 
24: Test Summary
24: ----------------------------
24: 	Successful tests:     1
24: 	Failed test:          1
24: 	Skipped tests:        0
24: 	Expected failures:    0
24: 	Unexpected successes: 0
24: 
24/26 Test #24: PhasorDynamicsLoadTest ...........***Failed    0.22 sec
```

pelesh:
@abirchfield, I believe docs for the `PhasorDynamics::Load` model need to be updated, as well, after you fixed the bug in the model?

pelesh:
> Everything looks good. Compiled and ran example 2 on my machine with no issues.

Perhaps we should use more descriptive names instead of "Example 1" and "Example 2"?

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -45,7 +45,43 @@ namespace GridKit
...
+        gen.allocate();
+        gen.initialize();
+        gen.evaluateResidual();
+
+        const std::vector<ScalarT>& f = gen.getResidual();
+        for (IdxT i = 0; i < 21; ++i)
+        {
+          if (!isEqual(f[i], 0.0, 1e-14))
+            success = false;
+        }
```
alexander-novo:
```suggestion
        for (auto f_val : f)
        {
          if (!isEqual(f_val, 0.0, 1e-14))
            success = false;
        }
```

Best convert this to a range-based for in case the dimension of the residual increases later and we accidentally miss a value not matching because we forgot to increase the bounds.

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -0,0 +1,125 @@
...
+  real_type worst_error      = 0;
+  real_type worst_error_time = 0;
+
+  const index_type stride = 94;
+  const index_type nt     = 2401;
+  scalar_type      results[stride];
+  buffer.seekg(0, std::ios::beg);
+  FILE* f = fopen("example2_results.csv", "w");
+  fprintf(f, "Time,gen2speed,gen3speed,v2mag,v3mag\n");
+  fprintf(f, "0,1,1,1,1\n");
```
alexander-novo: I think this is not up to gridkit style guidelines? Use C++-style file IO

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -0,0 +1,125 @@
...
+  double stop = static_cast<double>(clock());
+
+  /* Check worst-case error */
+  real_type worst_error      = 0;
+  real_type worst_error_time = 0;
+
+  const index_type stride = 94;
+  const index_type nt     = 2401;
+  scalar_type      results[stride];
+  buffer.seekg(0, std::ios::beg);
```
alexander-novo, line 93: Is this necessary? Seems confusing and unneeded to me.

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -0,0 +1,125 @@
...
+  FILE* f = fopen("example2_results.csv", "w");
+  fprintf(f, "Time,gen2speed,gen3speed,v2mag,v3mag\n");
+  fprintf(f, "0,1,1,1,1\n");
+  for (index_type i = 0; i < nt - 1; ++i)
+  {
+    for (index_type j = 0; j < stride; ++j)
+      buffer >> results[j];
```
alexander-novo, line 110: The existence of this `buffer` variable confuses me - we only ever read from it but do not write to it? What is it doing here?

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -0,0 +1,125 @@
...
+    scalar_type gen3speed_ref = reference_solution[i + 1][2];
+    scalar_type v2mag         = sqrt(results[2] * results[2] + results[3] * results[3]);
+    scalar_type v2mag_ref     = reference_solution[i + 1][4];
+    scalar_type v3mag         = sqrt(results[4] * results[4] + results[5] * results[5]);
+    scalar_type v3mag_ref     = reference_solution[i + 1][5];
+    fprintf(f, "%g,%g,%g,%g,%g\n", t, gen2speed, gen3speed, v2mag, v3mag);
+    real_type err = std::max(std::max(std::abs(gen2speed - gen2speed_ref),
+                                      std::abs(gen3speed - gen3speed_ref)),
+                             std::max(std::abs(v2mag - v2mag_ref),
+                                      std::abs(v3mag - v3mag_ref)));
```
alexander-novo:
```suggestion
    real_type err = std::max({std::abs(gen2speed - gen2speed_ref),
                              std::abs(gen3speed - gen3speed_ref),
                              std::abs(v2mag - v2mag_ref),
                              std::abs(v3mag - v3mag_ref)});
```

examples/PhasorDynamics/Example1/example1.cpp
```diff
@@ -1,29 +1,18 @@
...
+#include <Model/PhasorDynamics/Branch/Branch.hpp>
+#include <Model/PhasorDynamics/Bus/Bus.hpp>
+#include <Model/PhasorDynamics/Bus/BusInfinite.hpp>
+#include <Model/PhasorDynamics/BusFault/BusFault.hpp>
+#include <Model/PhasorDynamics/SynchronousMachine/GENROUwS/Genrou.hpp>
+#include <Model/PhasorDynamics/SystemModel.hpp>
+#include <Solver/Dynamic/Ida.hpp>
 #include <Utilities/Testing.hpp>
 
+#define _USE_MATH_DEFINES
```
alexander-novo: This define is useless down here - it must be included before `#include <cmath>`. Also this file doesn't use any math constants, which is why this didn't end up mattering.

alexander-novo (changes requested):
@pelesh I really don't like the littering of `_USE_MATH_DEFINES` everywhere - it's almost certainly going to continue to be a problem like this going forward for the MSVC platform (which requires its use to access math constants like `M_PI`). We should look into defining this macro in cmake rather than in every individual file which needs it. Otherwise I can foresee countless pull requests later adding these defines everywhere whenever someone on Linux needs to use `M_PI` and forgets to define this. Thoughts?

examples/PhasorDynamics/Example2/example2.cpp
alexander-novo: Wait I see where I missed the `Buffer` from earlier.

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  /* Set up simulation */
+  Ida<scalar_type, index_type> ida(&sys);
+  ida.configureSimulation();
+
+  /* Run simulation */
+  scalar_type start = static_cast<scalar_type>(clock());
+  // ida.printOutputF(0, 0, buffer);
+  ida.initializeSimulation(0.0, false);
+  ida.runSimulationFixed(0.0, dt, 1.0, buffer);
```
alexander-novo: I really don't like this `buffer` solution. It would be better to put these values in a data structure. See #91

pelesh:
> @pelesh I really don't like the littering of `_USE_MATH_DEFINES` everywhere - it's almost certainly going to continue to be a problem like this going forward for the MSVC platform (which requires its use to access math constants like `M_PI`). We should look into defining this macro in cmake rather than in every individual file which needs it. Otherwise I can foresee countless pull requests later adding these defines everywhere whenever someone on Linux needs to use `M_PI` and forgets to define this. Thoughts?

My suggestion would be to have a file with constants protected with a namespace and something like `constexpr real_type PI = M_PI`. We can def and undef `_USE_MATH_DEFINES` around it. In any case, we need one place where we define $`\pi`$ in the entire code. I would use a separate PR for that, though.

abirchfield (approved):
Fixed load bug, including impacts to tests and docs. Still runs great on my machine.

examples/PhasorDynamics/Example2/example2.cpp
pelesh: Let's include this in #91 scope.

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -85,17 +81,25 @@ int main()
...
   scalar_type      results[stride];
   buffer.seekg(0, std::ios::beg);
-  FILE* f = fopen("example2_results.csv", "w");
-  fprintf(f, "Time,gen2speed,gen3speed,v2mag,v3mag\n");
-  fprintf(f, "0,1,1,1,1\n");
+
+  std::ostream nullout(nullptr);
+  std::ostream& out = nullout;
+
+  // // Uncomment code below to print output to a file:
```
alexander-novo, line 98: Rather than uncommenting code this should probably be surrounded with `#IFDEF` so that you can switch with cmake.

pelesh: We will eventually add command line input for these things. This is temporary solution for now.

examples/PhasorDynamics/Example1/example1.cpp
pelesh: Removed both, the include statement and the define as they are not needed.

examples/PhasorDynamics/Example2/example2.cpp
pelesh: Fixed in [828e780](https://github.com/ORNL/GridKit/pull/85/commits/828e7803d2e348b94cdcd2c1ac8f050e64697775).

examples/PhasorDynamics/Example2/example2.cpp
pelesh: Fixed in [828e780](https://github.com/ORNL/GridKit/pull/85/commits/828e7803d2e348b94cdcd2c1ac8f050e64697775).

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
pelesh: Fixed in [828e780](https://github.com/ORNL/GridKit/pull/85/commits/828e7803d2e348b94cdcd2c1ac8f050e64697775).

alexander-novo (approved):

PR #86, opened 2025-04-22
Add dev container

alexander-novo:
## Description
 
Adds a dev container to simplify onboarding of new developers who aren't familiar with installing C++ dependencies, or who work in an environment like Windows with complex build processes. Developers using VS Code can simply build the included container, and re-open their workspace inside the container, which will mount the project directory into an environment with all of GridKit's prerequisites pre-installed.
 
 ## Further comments
 
In the future, this container could be modified to be used as a deployment environment for GridKit. This container is currently sitting at ~677 MB (which is mostly software like cmake, g++, etc.) but builds in < 1 minute for me.


superwhiskers (approved):
i've successfully built the container image and built gridkit in a running container with every test passing using podman (not in vscode, without docker).

my only comment would be that you could make the image smaller if you used an alpine linux base image instead of ubuntu. doing this is entirely up to you, but it could be worth it.

superwhiskers:
here's a dockerfile which builds one with alpine. it's around 417mb. the unfamiliarity of the environment may be a reason to not use it but i thought i'd provide it anyway as a suggestion. it's up to you
```dockerfile
FROM alpine:latest AS builder

RUN apk update && apk add --no-cache gcc g++ cmake samurai git openblas-dev
RUN git clone --single-branch --depth 1 --branch v7.10.2 https://github.com/DrTimothyAldenDavis/SuiteSparse.git /tmp/suitesparse && \
    cd /tmp/suitesparse && \
    mkdir -p build && cd build && \
    cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSUITESPARSE_ENABLE_PROJECTS="klu" .. && \
    ninja install && \
    rm -rf /tmp/suitesparse
# temporary switch to develop to handle LLNL/sundials#704
RUN git clone --single-branch --depth 1 --branch develop https://github.com/LLNL/sundials.git /tmp/sundials && \
    cd /tmp/sundials && \
    mkdir -p build && cd build && \
    cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_KLU=on .. && \
    ninja install && \
    rm -rf /tmp/sundials

FROM alpine:latest

RUN apk update && apk add --no-cache gcc g++ cmake samurai git openblas
COPY --from=builder /usr/local /usr/local
COPY --from=builder /usr/include /usr/include
```

alexander-novo:
I like the idea, but since the point was to make it as easy as possible to setup an environment for power systems people (several of whom I've talked to and aren't familiar with Linux), I think Ubuntu is probably the best option. My image is 677 MB - I don't know if the space savings is worth it?

I'd also prefer if we opened an issue to update the dockerfile with a new tagged version of Sundials once they release a new version with the fix, rather than downloading `develop`?

superwhiskers:
`develop` was necessary for the tests to pass. see the issue linked in the comment as well as [@pelesh's comment here](https://github.com/ORNL/GridKit/pull/115#issuecomment-2913885248)

alexander-novo:
Well we're officially advertising the prerequisite version of [SUNDIALS](https://github.com/LLNL/sundials) >= 7.0.0 in the README, so that either needs to be adjusted or that test needs to be disabled.

Regardless `develop` isn't an appropriate dependency version requirement - it should be pinned to the specific commit that fixes the issue. I have no clue how clean the `develop` branch in SUNDIALS is meant to be, and any random commit could just start breaking docker images.

pelesh:
> Well we're officially advertising the prerequisite version of [SUNDIALS](https://github.com/LLNL/sundials) >= 7.0.0 in the README, so that either needs to be adjusted or that test needs to be disabled.

We need to change this and set new depencency to SUNDIALS `develop` branch > 07d21c2c5ae33211a9a2fafd8eac56de5582dce0

This will likely be included in SUNDIALS release 7.3.1.

I suggest you just update the main README file and we can merge this.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# Install sundials using klu from previous step
+RUN git clone --single-branch --branch develop https://github.com/LLNL/sundials.git /tmp/sundials && \
+	cd /tmp/sundials && \
+	git checkout 07d21c2c5ae33211a9a2fafd8eac56de5582dce0 && \
+	mkdir -p build && cd build && \
+	cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DENABLE_KLU=ON \
+		-DKLU_ROOT=/usr/local/lib/cmake/KLU .. && \
+	ninja && \
+	ninja install && \
```
superwhiskers: you can combine these lines into just `ninja install` if you wanted to

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+	&& apt clean && rm -rf /var/lib/apt/lists/*
+
+# Install SuiteSparse for klu
+RUN git clone --single-branch --depth 1 --branch v7.10.2 https://github.com/DrTimothyAldenDavis/SuiteSparse.git /tmp/suitesparse && \
+	cd /tmp/suitesparse && \
+	mkdir -p build && cd build && \
+	cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DSUITESPARSE_ENABLE_PROJECTS="klu" .. && \
+	ninja && \
+	ninja install && \
```
superwhiskers: same here

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+	rm -rf /tmp/suitesparse
+
+# Install sundials using klu from previous step
+RUN git clone --single-branch --branch develop https://github.com/LLNL/sundials.git /tmp/sundials && \
+	cd /tmp/sundials && \
+	git checkout 07d21c2c5ae33211a9a2fafd8eac56de5582dce0 && \
+	mkdir -p build && cd build && \
+	cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DENABLE_KLU=ON \
+		-DKLU_ROOT=/usr/local/lib/cmake/KLU .. && \
```
superwhiskers: `KLU_ROOT` isn't necessary here. cmake is capable of detecting it.

superwhiskers (approved):
some minor suggestions i found while writing the alpine dockerfile

.devcontainer/Dockerfile
alexander-novo: Interesting. Did you test this? It wasn’t the case when I’ve installed on my machine and others’.

superwhiskers: it worked fine in the container. i can test it again but it worked with the ubuntu base image as well as the alpine base image

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:24.04
+
+RUN apt update -y && apt install -y \
+	gcc \
```
superwhiskers: if you want to reduce the size of the image down to about 656mb, you can swap out `apt install -y` with `apt install -y --no-install-recommends` and add `ca-certificates` to the list of packages to install (as this is needed to clone the git repositories later).

superwhiskers (approved):
i don't have anything else beyond this. i think this is good to merge as-is, but this one suggestion would reduce the size of the image a little bit more.

.devcontainer/Dockerfile
alexander-novo: That sounds good to me!

pelesh (approved):

pelesh:
We should probably use Spack to build GridKit within the container, but let's merge this first.

CC @nkoukpaizan

superwhiskers:
@nkoukpaizan, the `make` package was not strictly necessary to install, we have `ninja` already. probably the only instance in which it is necessary is if an end user using the container to develop something wanted to generate makefiles instead of a build configuration for `ninja`.

alexander-novo:
Yes that was precisely why I included it. This is a dev container so I wanted it to be a bit “batteries included”. In the future when we design a production containers, moving to alpine and stripping build tools like ninja and make will be good.

PR #89, opened 2025-04-22
Verify Branch and Load Jacobians computed with Sparse::Variable

nkoukpaizan:
## Description
 
This adds tests for Branch and Load Jacobians.
 
 ## Proposed changes

Added a Jacobian test for BranchTest and LoadTest. Using Gridkit::Sparse::Variable, we track model residual dependencies and compute the values of the Jacobian. The result is compared to the analytical Jacobian in std::map format. The Jacobians are constant matrices in these cases. 
 
 ## Checklist
 
- [x] All tests pass.

### DIFF
--- a/src/Utilities/Testing.hpp
+++ b/src/Utilities/Testing.hpp
@@ -232,7 +232,6 @@ namespace GridKit
       for (const auto& pair_a : a)
       {
         auto it = b.find(pair_a.first);
-        std::cout << pair_a.second << ", " << it->second << std::endl;
         fail += !isEqual(pair_a.second, it->second);
       }
 
--- a/src/Utilities/Testing.hpp
+++ b/src/Utilities/Testing.hpp
@@ -243,7 +243,7 @@ namespace GridKit
       for (const auto& pair_a : a)
       {
         auto it  = b.find(pair_a.first);
-        fail    += !isEqual(pair_a.second, it->second);
+        fail    += !isEqual(pair_a.second, it->second, tol);
       }
 
       return fail == 0;
--- a/src/Utilities/Testing.hpp
+++ b/src/Utilities/Testing.hpp
@@ -228,7 +228,7 @@ namespace GridKit
      *
      * @param[in] std::map<IdxT, RealT> - first map
      * @param[in] std::map<IdxT, RealT> - second map
-     * @param[in] RealT - tolerance 
+     * @param[in] RealT - tolerance
      * @return bool - true if the maps are equal; false otherwise
      */
     template <typename IdxT = size_t, typename RealT = double>
--- a/src/Utilities/Testing.hpp
+++ b/src/Utilities/Testing.hpp
@@ -223,7 +223,7 @@ namespace GridKit
     template <typename IdxT = size_t, typename RealT = double>
     inline bool isEqual(std::map<IdxT, RealT> a,
                         std::map<IdxT, RealT> b,
-                        const RealT tol = std::numeric_limits<RealT>::epsilon())
+                        const RealT           tol = std::numeric_limits<RealT>::epsilon())
     {
       int fail = 0;
 
@@ -231,8 +231,8 @@ namespace GridKit
 
       for (const auto& pair_a : a)
       {
-        auto it = b.find(pair_a.first);
-        fail += !isEqual(pair_a.second, it->second);
+        auto it  = b.find(pair_a.first);
+        fail    += !isEqual(pair_a.second, it->second);
       }
 
       return fail == 0;
--- a/src/Utilities/Testing.hpp
+++ b/src/Utilities/Testing.hpp
@@ -249,13 +249,22 @@ namespace GridKit
       for (const auto& pair_a : a)
       {
         auto it = b.find(pair_a.first);
-        if (!isEqual(pair_a.second, it->second, tol))
+        if (it != b.end())
+        {
+          if (!isEqual(pair_a.second, it->second, tol))
+          {
+            fail++;
+            errs() << "Mismatching map values! "
+                   << "a.first = " << pair_a.first << ", "
+                   << "a.second = " << pair_a.second << ", and "
+                   << "b.second = " << it->second << "\n";
+          }
+        }
+        else
         {
           fail++;
-          errs() << "Mismatching map values! "
-                 << "a.first = " << pair_a.first << ", "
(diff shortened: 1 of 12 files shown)


src/Utilities/Testing.hpp
```diff
@@ -220,6 +220,35 @@ namespace GridKit
       return fail == 0;
     }
 
+    /**
+     * @brief Equatlity comparison between maps with a tolerance for the scalar value
+     *
+     * @tparam IdxT
+     * @tparam RealT
+     *
+     * @param[in] std::map<IdxT, RealT> - first map
```
alexander-novo: Is this doxygen formatting correct? I think this should be the name of the parameter - not its type, right?

src/Utilities/Testing.hpp
```diff
@@ -220,6 +220,35 @@ namespace GridKit
...
+                        const RealT           tol = std::numeric_limits<RealT>::epsilon())
+    {
+      int fail = 0;
+
+      fail += a.size() != b.size();
+
+      for (const auto& pair_a : a)
+      {
+        auto it  = b.find(pair_a.first);
+        fail    += !isEqual(pair_a.second, it->second);
```
alexander-novo:
```suggestion
        fail    += !isEqual(pair_a.second, it->second, tol);
```

Should use tolerance here

alexander-novo: Also I think there should be some better error output here, like in the `std::vector` implementation - it would be good to output what pair we expected to find in `b` when we find something that is missing.

alexander-novo (changes requested):
Maybe I'm misunderstanding, but wasn't the point of `Sparse::Variable` that we wouldn't have to write these `analyticalJacobian` functions manually, and would be able to call e.g. `LoadTests<Sparse::Variable>::residual()` to compute the same thing?

src/Utilities/Testing.hpp
alexander-novo: I also think it's not necessary to annotate `param[in]` for non-pointer arguments but also I guess it doesn't hurt.

nkoukpaizan:
> Maybe I'm misunderstanding, but wasn't the point of `Sparse::Variable` that we wouldn't have to write these `analyticalJacobian` functions manually, and would be able to call e.g. `LoadTests<Sparse::Variable>::residual()` to compute the same thing?

This is exactly what is being done here. See for example, [load.evaluateResidual()](https://github.com/ORNL/GridKit/blob/16cad5979fbc9eb0a270cee6a1b2f80fbe6c06ee/tests/UnitTests/PhasorDynamics/LoadTests.hpp#L86). The point of the analytical Jacobian is to verify the answer we are getting. For a unit test, we need a reference.

src/Utilities/Testing.hpp
nkoukpaizan: type --> variable name fixed in [3ea65fb](https://github.com/ORNL/GridKit/pull/89/commits/3ea65fb2a569789533cf3a667568af94f6b69563).

alexander-novo (approved):

src/Utilities/Testing.hpp
```diff
@@ -238,12 +238,25 @@ namespace GridKit
...
+        errs() << "Containers do not have the same size! "
+               << "a.size() = " << a.size() << ", and "
+               << "b.size() = " << b.size() << "\n";
+      }
 
       for (const auto& pair_a : a)
       {
-        auto it  = b.find(pair_a.first);
-        fail    += !isEqual(pair_a.second, it->second, tol);
+        auto it = b.find(pair_a.first);
```
alexander-novo, line 253: Make sure to verify that `it != b.end()`, otherwise you'll be de-referencing an out-of-bounds pointer.

nkoukpaizan: Good catch. Addressed in [4acc608](https://github.com/ORNL/GridKit/pull/89/commits/4acc608ef8870d0f6bf91ca92ae0f6e24c64b29b).

shakedregev:
```
48%] Building CXX object examples/MatPowerTesting/CMakeFiles/test_parse_gen_row.dir/test_parse_gen_row.cpp.o
In file included from /Users/szb/GridKit_dir/GridKit/examples/MatPowerTesting/test_parse_gen_row.cpp:5:
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:235:30: error: no template named 'map' in namespace 'std'; did you mean 'max'?
  235 |     inline bool isEqual(std::map<IdxT, RealT> a,
      |                         ~~~~~^~~
      |                              max
/opt/homebrew/Cellar/llvm/19.1.7/bin/../include/c++/v1/__algorithm/max.h:29:1: note: 'max' declared here
   29 | max(_LIBCPP_LIFETIMEBOUND const _Tp& __a, _LIBCPP_LIFETIMEBOUND const _Tp& __b, _Compare __comp) {
      | ^
In file included from /Users/szb/GridKit_dir/GridKit/examples/MatPowerTesting/test_parse_gen_row.cpp:5:
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:235:17: error: redefinition of 'isEqual' as different kind of symbol
  235 |     inline bool isEqual(std::map<IdxT, RealT> a,
      |                 ^
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:207:17: note: previous definition is here
  207 |     inline bool isEqual(PowerSystemData::SystemModelData<RealT, IdxT> a,
      |                 ^
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:235:47: error: expected ')'
  235 |     inline bool isEqual(std::map<IdxT, RealT> a,
      |                                               ^
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:235:24: note: to match this '('
  235 |     inline bool isEqual(std::map<IdxT, RealT> a,
      |                        ^
/Users/szb/GridKit_dir/GridKit/src/Utilities/Testing.hpp:237:91: error: expected ';' after top level declarator
  237 |                         const RealT           tol = std::numeric_limits<RealT>::epsilon())
      |                                                                                           ^
      |                                                                                           ;
4 errors generated.
```

shakedregev (changes requested):
I cannot build this code. Details above.

shakedregev (approved):

pelesh (approved):
All tests pass with Clang and GCC builds. Code clean and well commented.

PR #91, opened 2025-04-28
Add IDA interface callbacks

### LINKED ISSUE #21: GridKit interface with IDA can't return values

alexander-novo:
## Description
 
A separate function is created for each example which needs output across different time values from IDA (until now only 1 example). This functionality should be consolidated into a single function which is general enough for all use cases/examples.
 
Closes #21
 
 ## Proposed changes
 
A new `callback` parameter has been added to `Ida::runSimulation` which is called every time IDA produces a new solution.
 
 ## Checklist
 
- [x] All tests pass.


rothpc:
The general idea of supporting a callback, callable by anything that understands operator() seems like a good one.  I have a few comments/suggestions about the implementation.

Driven by the example of using the interface shown in commit 2f7c33d, I suggest encapsulating into a functor the declaration of the data to be stored each time step, the function that does the storing, and in this specific situation also the data structure holding the data for all time steps.  In this specific example, this functor would look something like this:
```
struct Trace
{
    struct OutputData
    {
        double t;
        ...
    };
    vector<OutputData> output;

    operator() { output.push_back(t, yval[0], ...) }
};
```
To use, you'd declare a Trace object (assume it is named `trace`) and pass it to runSimulation(), and then access `trace.output` after the simulation is done to do the error check.  Encapsulating these together into a functor class should make it easier for someone else to use a different functor (e.g., one that writes to a file) or just to add a new value to the data that gets saved each time step.

My second comment/suggestion is just to be aware of the scalability aspect of storing data for each time step for the entire run to a vector.  For the example, I suspect it's fine because of the small amount of data saved per time step and/or a small enough number of iterations.  In production, if there's more data saved or many more time steps, it might need a more complex approach that preallocates a buffer and then flushes it to disk when the buffer gets full.  In the case of the example, if the number of iterations is known a priori I think I'd still use std::vector::reserve() to avoid having to reallocate and then copy its contents whenever the functor does a push_back() but the vector's current memory allocation is full.

In performance tools that save event traces (not dissimilar to what the example is doing), people deal with large data volume by using buffers that they flush to files when they get full.  It's not the point of the example, so I'm not suggesting that you should implement a tracer functor that does buffer+flush here, just that there might eventually be a need for an implementation that does something more complex

alexander-novo:
@rothpc Maybe it's worth doing what you describe for a more general class of "output capturing classes" that provide callbacks that are commonly used - such as ones that you describe here. Then rather than examples defining their own callbacks every time, they can just declare e.g. which components they care about. That sounds like it makes sense to me?

rothpc:
I was going to suggest combining the two ideas, namely using a functor for the callback, and defining base classes for the functors that provide basic functionality.  But as I thought about it - for the simpler use cases like the one shown in the example, it feels pretty artificial to declare a base class.  Even the approach of providing a templatized base class that takes the struct type saved for each iteration as a template parameter - all that the base class could really do is declare the vector and a stub operator() method that would have to be overwritten.  It's easier to foresee a base class for the more complicated storage situation I was talking about in my earlier comment, but an alternative approach that provides a generic (or templatized) buffering/flushing support outside of a functor, and then justing using it in functors as needed.  Note also that I think the project should hold off on implementing such a thing until there's an actual need - but just be thinking about the potential for such a need in the future when designing the callback support now to hopefully going to far down a path that precludes the more complicated/scalable callback output options later.

alexander-novo:
Yeah I think I prefer this approach in the API because it allows for your suggestion in use while also allowing for simpler lambdas in simple examples, and I'm not sure I see why I would want to force people to jump through the hoop of having to define their own class for all of this when the callback could be as simple as a single line.

examples/PhasorDynamics/Example1/example1.cpp
```diff
@@ -71,86 +71,96 @@ int main()
...
+  //   Genrou<double, size_t>&  gen;
+
+  //   void operator()(double t)
+  //   {
+  //     output.push_back({.ti = t,
+  //                       .Vr = bus1.Vr(),
+  //                       .Vi = bus1.Vi(),
+  //                       .dw = gen.y()[1]});
+  //   }
+  // } output_cb = {output, bus1, gen};
```
pelesh: I would think the lambda is equivalent to
```c++
struct OutData
{
  real_type ti;
  real_type Vr;
  real_type Vi;
  real_type dw;
};

std::vector<OutData> output;
OutData tmp;

tmp.ti = t;
tmp.Vr = bus.Vr;
tmp.Vi = bus.Vi;
tmp.dw = gen.y()[1];
output.push_back(tmp);
```
I think this description is a little easier to follow. In both cases, it is assumed `t`, `bus1`, `bus2`, and `gen` are instantiated and their values set.

src/Model/PhasorDynamics/SystemModel.hpp
```diff
@@ -648,10 +510,89 @@ namespace GridKit
...
+      /// @brief Whenever the internal state of a system is updated, the system
+      ///        needs to reflect those changes in its "children" aka buses and
+      ///        components. `updateChildren` will only update `y`, `yp`, and
+      ///        `param` - call when you know only the internal state of those
+      ///        variables has been updated.
+      void updateChildren()
+      {
+        // Update variables
+        IdxT varOffset = 0;
+        IdxT optOffset = 0;
```
pelesh: This is very helpful! I suggest we put this in a separate PR as it is out of scope here.

src/Solver/Dynamic/Ida.hpp
```diff
@@ -33,11 +35,21 @@ namespace AnalysisManager
...
+      ///                      initial conditions if needed) before calling `runSimulation`.
+      /// @param tf            The final time, used to calculate the size of each step IDA should take. The actual time
+      ///                      of the final IDA solution should be somewhat close to `tf`, however due to rounding error
+      ///                      the precise final time may be before or after `tf`.
+      ///
+      /// @param nout          The number of times the IDA integrator should be invoked to calculate the solution at `tf`.
+      ///
+      /// @param step_callback An optional callback which, if provided, will be called after each time the IDA solver has
+      ///                      been invoked with the value of `t` that IDA has calculated the last step at. The provided
+      ///                      model will be updated with the latest values of `y` and `yp` before the callback is invoked.
```
pelesh, line 50: Please have this in source file rather than here.

src/Solver/Dynamic/Ida.cpp
```diff
@@ -248,7 +223,18 @@ namespace AnalysisManager
...
         retval = IDASolve(solver_, tout, &tret, yy_, yp_, IDA_NORMAL);
         checkOutput(retval, "IDASolve");
-        // printOutput(tout);
+
+        if (step_callback.has_value())
+        {
+          // The callback may try to observe upated values in the model, so we should update them here
+          // (At this point, the model's values are one internal integrator step out of date)
+          model_->updateTime(tret, 0.0);
+          copyVec(yy_, model_->y());
```
pelesh, line 232: I believe the model time here should be set to `tout`
```suggestion
          model_->updateTime(tout, 0.0);
          copyVec(yy_, model_->y());
```

pelesh (changes requested):
I suggest adding `updateChildern()` function to aseparate branch/PR as it is out of scope here.

I also made a few other comments to discuss further.

examples/PhasorDynamics/Example1/example1.cpp
alexander-novo: No, the struct version is literally what the compiler constructs when it encounters the lambda. This makes it clear that the versions of `bus`, `gen`, and `output` that are used in the lambda are actually *references* to the external variables. It also makes it clear why you must put `auto` in front of the type for `output_cb` since the struct the compiler constructs is unnamed.

I have written it in a way such that if you were to uncomment the "equivalent" code and comment out the lambda, the code would compile and run identically to the lambda version. Your version would only run the lambda once - before the simulation is run.

src/Solver/Dynamic/Ida.hpp
alexander-novo: If you prefer this, then we should put this in the style guidelines, however I would push back a little against this since, in general, editors do not know where the implementation for a function is but do know where the declaration is, so documentation put on the declaration will be available to editors while documentation put on the implementation may or may not be available for editors. This also goes for developers - in general the first place I look for information about a function is where I know it will be, which is in the header file. And if anyone ever installs GridKit, they probably won't have the `.cpp` files around anymore - only the header files and compiled objects.

See e.g. [[1]](https://stackoverflow.com/questions/691954/should-i-comment-the-declaration-or-the-definition-in-c) [[2]](https://softwareengineering.stackexchange.com/questions/84071/is-it-better-to-document-functions-in-the-header-file-or-the-source-file)

src/Solver/Dynamic/Ida.cpp
alexander-novo: If everything goes well with IDA then `tout == tret` however `tret` is the time that IDA is asserting the current state is at - if anything happens to go wrong I'd rather trust what time IDA says it is than what time I'm assuming it should be. According to [the IDA documentation](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDASolve), there are 3 different reasons IDA can return successfully:
- `IDA_SUCCESS`: we reached `tout` and returned - this is what normally happens,
- `IDA_TSTOP_RETURN`: a stop time earlier than `tout` was configured and we stopped early - then if I use `tout` instead of `tret` this is a bug.
- `IDA_ROOT_RETURN`: IDA was configured to search for roots and found one - then if I use `tout` instead of `tret` this is a bug.

I don't think we'll ever use those configuration options but the way I see it using `tret` in this situation is just more future-proof than using `tout`

src/Model/PhasorDynamics/SystemModel.hpp
alexander-novo: Just the `updateChildren()` function or the entire 37ea7213041afef6ac1e9ce85855db1c83bef53f commit?

alexander-novo: Actually I think I'm just going to do the whole commit and expand on it in a different PR

pelesh: Yes, just `updateChildren()`. It has little to do with callbacks.

alexander-novo:
Should I add unit tests for this? I think I've been considering the examples as integration tests.

pelesh:
> Should I add unit tests for this? I think I've been considering the examples as integration tests.

Yes, a unit test would be helpful.

alexander-novo:
Can anyone help me with this test failure? I'm just trying to create a minimal problem for IDA to solve because I'm not trying to test the solving capabilities of IDA, so I tried a 0-dim problem (which Sundials' linear solver did not like) and now I'm trying to solve the scalar problem 0=0 (I also tried the problem y' = 0 and set `f_ = yp_` in `evaluateResidual`) but IDA is failing with a convergence error.

alexander-novo:
With Steven's advice, I updated the test problem to y = 0 instead of 0 = 0, but I'm still getting a failure.

pelesh:
> With Steven's advice, I updated the test problem to y = 0 instead of 0 = 0, but I'm still getting a failure.

I suggest we go without unit tests in this PR. The issue is tracked in #100.

pelesh (approved):

PR #93, opened 2025-04-30
Add input format specifications for phasor dynamics models

pelesh:
## Description
 
Add markdown file with specs for the input file format for phasor domain dynamics and (hopefully) electromagnetic transient models.

@superwhiskers 
 
 ## Proposed changes
 
This is a specification document for the parser yet to be developed. Here we propose to use JSON to specify inputs rather than using (and eventually extending) other plain text file formats, such as Matpower.

 ## Checklist
 
 N/A
 
- [ ] All tests pass.


abirchfield (approved):
Fixed a few quick things on the markdown version. Otherwise it looks great.

PR #102, opened 2025-05-07
Exciter Documentation

lukelowry:
## Description
 
Documentation for the phase dynamic exciter (IEEET1) that will be implemented following this PR.
 
 
 ## Proposed changes
 
Restructured folder to allow for many exciter types and added documentation for IEEET1
 
 
 ## Further comments
 
There are piece-wise differential equations in this system. I wrote a short proposal on how to implement this to cooperate with the general solver. Please let me know if a bump function is sufficient or if the general solver can handle piece-wise definition (not to my knowledge).


abirchfield (changes requested):
Documentation needs to be updated to reflect structure in  #127 and #128 .

I suggest documenting both piecewise and an initial smoothed approximation for the limits.

abirchfield:
Going to revisit this with another branch soon.

PR #103, opened 2025-05-07
Governor Documentation and Model

lukelowry:
## Description
 
Documentation for the phasor dynamics governor (TGOV1) that will be implemented following this PR.
 
 
 ## Proposed changes
 
Restructured folder to allow for many gov. types and added documentation for TGOV1
 
 
 ## Further comments
 
There are piece-wise differential equations in this system. I wrote a short proposal on how to implement this to cooperate with the general solver. Please let me know if a bump function is sufficient or if the general solver can handle piece-wise definition (not to my knowledge).

Also there was a small latex issue in GENROU that I included in this PR.

 ## Checklist
 
- [x] Governor Implementation
- [ ] Proper CMake build configuration.
- [x] Documentation in a form of README file with model equations.
- [ ] Example of usage.
- [ ] All tests pass.


src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
+# **TGOV1**
+
```
pelesh: Probably a good idea to spell out this acronym.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
...
+## Control Diagram
+
+Standard model of the stream turbine
+
+<div align="center">
+   <img align="center" src="../../../../../docs/Figures/TGOV1.JPG">
+
```
pelesh: Nor all variables from the equations show up in this graphics. For example, where is $P_{tx}$ input in this schematics?

src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
...
+   
+  Figure 1: Governor TGOV1 model. Figure courtesy of [PowerWorld](https://www.powerworld.com/WebHelp/)
+</div>
+
+## Nomenclature
+
+### Algebraic Variables
+- $\omega$ - Per-unit Speed Deviation
+- $P_{ref}$ - Reference Power
+- $P_{mech}$ - Mechnical Power to Generator
```
pelesh: Please specify units explicitly. Isn't $\omega$ expressed in [1/s] even in power units?

src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
...
+- $P_{v}$ - Valve Position
+
+### Parameters
+- $R$ - Droop Constant
+- $T_1$ - Valve Time Delay
+- $T_2$ - Turbine Numerator Time Constat
+- $T_3$ - Turbine Delay
+- $V_{max}$ - Max Valve Position
+- $V_{min}$ -  Min Valve Position
+- $D_t$ - Turbine Damping Coefficient
```
pelesh: Units need to be specified. If units are dimensionless, that should be stated explicitly or expressed as [1].

pelesh: Also $V_{max}$ and $V_{min}$ could be confused with voltages. Is this common notation? It would make more sense to use $P_{vmax}$ and $P_{vmin}$, respectively or even $\overline P_{v}$ and $\underline{P_{v}}$.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
...
+### Differential Equations
+The TGOV1 differential equations, as derived from the model diagram.
+```math
+\begin{aligned}
+   \dot{P}_{tx}   &= P_v - \dfrac{1}{T_3}(P_{tx}+T_2P_v) \\
+   \dot{P}_{v}    &= 
+      \dfrac{1}{T_1}
+   \begin{cases}
+      -P_{v} + \frac{1}{R}(P_{ref}-\omega)
+         &  \text{if } P_v \in (V_{min}, V_{max})\\
```
pelesh: Use `\dfrac` so that 1/R is more prominently displayed:
```suggestion
      -P_{v} + \dfrac{1}{R}(P_{ref}-\omega)
         &  \text{if } P_v \in (V_{min}, V_{max})\\
```

src/Model/PhasorDynamics/Governor/TGOV1/README.md
```diff
@@ -0,0 +1,80 @@
...
+            \left(\dfrac{x-x_0}{\epsilon}\right) \\
+   \phi(x)  &= I \left(x;V_{min}, \epsilon \right) -
+            I \left(x;V_{max}, \epsilon \right)
+            \qquad \epsilon <<1 \\
+   \dot{P}_{v}   
+            &= 
+            \dfrac{1}{T_1}\cdot \phi(P_v)
+            \left[
+               -P_{v} + \frac{1}{R}(P_{ref}-\omega)
+            \right] \\
```
pelesh: I am not sure this is correct approximation.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
pelesh: Also in the figure $\Delta \omega$ is used instead of $\omega$.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
pelesh: If I am not mistaken, label (2) in the figure denotes $P_v$ and label (1) denotes $P_{tx}$. If so, this has to be specified in the text.

Another question -- what is the input to the negative signal input to the summation on the left-hand-side of the diagram? It looks as if it is $\Delta \omega$, but if this is the case, units need to be explained.

pelesh (changes requested):
Great work! It would be good to make some additional clarification and make sure equations match the signal flow diagram.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: It is now **_Steam Turbine-Governor Model (TGOV1)_**

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: The differential variables $P_{tx}$ and $P_{v}$ are referenced to (1) and (2), respectively in the flow diagram.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: Corrected

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: No, p.u. frequency/speed is unitless, such that $\omega_{p.u.}=1$ represents nominal frequency. 

I will change $\omega \mapsto \Delta \omega$ since other models in the PhasorDynamics module use $\omega$ for speed, rather than speed deviation.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: 1.  Complete

2.   I agree here, I propose $P_{vmin}$ and $P_{vmax}$ as they are the clearest in my estimation.

3. Correct, this has now been specified under Nomenclature.  Also, the input you referenced is indeed $\Delta \omega$. Both  $\Delta \omega$ and $P_{ref}$ are in per-unit. The unit specification should clarify this now.

src/Model/PhasorDynamics/Governor/TGOV1/README.md
lukelowry: I will generalize this for now until we converge on a solution for smooth indicator.

lukelowry:
Comments addressed. Awaiting @abirchfield comments, especially in regards to clarifying units.

lukelowry:
Untested implementation in recent commit, aligned with documentation.

abirchfield (changes requested):
Documentation needs to be updated to reflect structure in #127 and #128 .

I suggest documenting both piecewise and an initial smoothed approximation for the limits.

abirchfield:
Also, name of model should be TGOV1 or Governor_TGOV1. TurbineGov could apply to other future governors and we want each to be unique.

pelesh:
> Also, name of model should be TGOV1 or Governor_TGOV1. TurbineGov could apply to other future governors and we want each to be unique.

Per developer guidelines class names should be upper camel case format, something like this: `GovernorTgov1`. Would that be acceptable to domain specialists?

We could also use just `Tgov1` and put it in a `Governor` namespace. In that case one would reference the class as `Governor::Tgov1`.

CC @superwhiskers @nkoukpaizan @alexander-novo

abirchfield:
GovernorTgov1 is ok.

pelesh:
Please make sure you rename the example. We already have `Example3`, it's the ten-gen model.

PR #104, opened 2025-05-07
Reorganize examples by model family

pelesh:
## Description
 
As there are more examples being added to GridKit, it would be desirable to have them organized in a way that one can navigate them more easily. For instance, experimental examples should be in a directory separate from more commonly used examples. 

 ## Proposed changes
 
 In this PR we reorganize examples in following directories:
- `Enzyme`: Enzyme usage examples.
- `Experimental`: Examples of dynamic-constrained optimization.
- `LinearAlgebra`: Examples and tests for dense and sparse matrices in GridKit.
- `PhasorDynamics`: all examples describing use of GridKit for phasor domain dynamic simulations.
- `PowerElectronics`: examples describing microgrid simulations with GridKit.
- `PowerFlow`: power flow steady-state simulations with GridKit and Matpower file parser.

These categories will help users and developers find all examples relevant for their work more easily. Typically users and developers would need to work with only one of those example families.

 ## Checklist
 
 
- [x] All tests pass.
- (N/A) There are unit tests for the new code.
 
 ## Further comments
 
None


abirchfield (approved):
I think it is a much better structure

PR #106, opened 2025-05-13
Fixes to assertions and overrides

nkoukpaizan:
## Description
 
Building in `debug` mode revealed a few `assert`s that were failing

One failing assertion was due to a member function name change:
```
GridKit/src/Solver/Optimization/DynamicConstraint.cpp:31:22: error: no member named 'size_quad' in 'GridKit::Model::Evaluator<double, long>'
   31 |       assert(model_->size_quad() == 1);
      |              ~~~~~~  ^
```

Another failing assertion was due to a comparison between two floating point numbers:
```
Assertion failed: (t == other.t), function operator-=, file example2.cpp, line 42.
```

Additionally, LLVM tends to issue more warnings, such as overrides
```
GridKit/tests/UnitTests/Solver/Dynamic/IdaTests.hpp:113:29: warning: 'y' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  113 |       std::vector<ScalarT>& y()
      |
```
 ## Proposed changes
- `size_quad()` --> `sizeQuadrature()`
- Use a tolerance to compare the floating point values in example2. Also put the data within a namespace.
- Adding missing overrides
 
 ## Checklist 
- [X] All tests pass.
 
 ## Further comments
 I suggest ignoring whitespaces for the diff for the `example*hpp` files.


pelesh (approved):
Tests pass when built with GCC and Clang. Good to merge.

examples/PhasorDynamics/Example2/example2.cpp
```diff
@@ -39,7 +39,7 @@ struct OutputData
 
   OutputData& operator-=(const OutputData& other)
   {
-    assert(t == other.t);
+    assert((t - other.t) < Example2::reference_tol);
     gen2speed -= other.gen2speed;
```
pelesh: I suggest using GridKit's comparison for floating point numbers:
```c++

...

assert(isEqual(t, other.t));
```

PR #115, opened 2025-05-27
Create phasor dynamics model data structures

pelesh:
## Description
 
@abirchfield @lukelowry @alexander-novo @reid-g @abdourahmanbarry 
 
 ## Proposed changes
 
 
 ## Checklist
 
 
- [ ] All tests pass.
 
 ## Further comments


superwhiskers:
i don't think this is related to any change made in this pull request, but there's a segfault in the example `AdjointSens` in the destructor `AnalysisManager::Sundials::Ida::~Ida` caused by `Ida::deleteAdjoint()` calling `IDAAdjFree(solver_)` and then `Ida::deleteSimulation()` calling `IDAFree(&solver_)` which calls `IDAAdjFree(...)` as well. from what i can tell, the best way to solve this would be to remove the call to `Ida::deleteAdjoint()` in the destructor, but i could be wrong about this.

pelesh:
> i don't think this is related to any change made in this pull request, but there's a segfault in the example `AdjointSens` in the destructor `AnalysisManager::Sundials::Ida::~Ida` caused by `Ida::deleteAdjoint()` calling `IDAAdjFree(solver_)` and then `Ida::deleteSimulation()` calling `IDAFree(&solver_)` which calls `IDAAdjFree(...)` as well. from what i can tell, the best way to solve this would be to remove the call to `Ida::deleteAdjoint()` in the destructor, but i could be wrong about this.

This issue was supposed to be fixed in SUNDIAL PR [704](https://github.com/LLNL/sundials/pull/704). For now, we need to use SUNDIALS from `develop` branch.

pelesh:
Closing this in favor of #116.

PR #116, opened 2025-05-27
Create phasor dynamics model data structures (take two)

pelesh:
## Description

Create data structures to store model data for each component model type, e.g. an electrical machine data structure. These data structures will contain a superset of parameters for each component model type.
 
@abirchfield @lukelowry @alexander-novo @reid-g @abdourahmanbarry 
 
 ## Proposed changes
 
The purpose of these data structures is to decouple data input from component models, so that:
- Multiple inputs (GUI, file input, etc.) can be used to enter system model data in GridKit.
- Multiple component model implementations can use the same data structures to instantiate the component.
- Analysis can be performed on these data structures to compute sparsity pattern of system Jacobian without even instantiating the component or system models.

The changes proposed in the  PR are the first step in this direction.
 
 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/src/Model/PhasorDynamics/Branch/Branch.hpp
+++ b/src/Model/PhasorDynamics/Branch/Branch.hpp
@@ -15,17 +15,11 @@ namespace GridKit
 {
   namespace PhasorDynamics
   {
-    template <typename RealT, typename IdxT>
-    struct BranchData;
-  }
-} // namespace GridKit
-
-namespace GridKit
-{
-  namespace PhasorDynamics
-  {
     template <class ScalarT, typename IdxT>
     class BusBase;
+
+    template <typename RealT, typename IdxT>
+    struct BranchData;
   }
 } // namespace GridKit
 
@@ -154,8 +148,8 @@ namespace GridKit
       real_type  X_;
       real_type  G_;
       real_type  B_;
-      const IdxT bus1ID_;
-      const IdxT bus2ID_;
+      const IdxT bus1_id_;
+      const IdxT bus2_id_;
     };
 
   } // namespace PhasorDynamics
--- a/src/Model/PhasorDynamics/Branch/Branch.hpp
+++ b/src/Model/PhasorDynamics/Branch/Branch.hpp
@@ -20,7 +20,7 @@ namespace GridKit
 
     template <typename RealT, typename IdxT>
     struct BranchData;
-  }
+  } // namespace PhasorDynamics
 } // namespace GridKit
 
 namespace GridKit
--- a/src/Model/PhasorDynamics/Branch/Branch.hpp
+++ b/src/Model/PhasorDynamics/Branch/Branch.hpp
@@ -13,7 +13,7 @@
 // Forward declarations.
 namespace GridKit
 {
-  namespace PowerSystemData
+  namespace PhasorDynamics
   {
     template <typename RealT, typename IdxT>
     struct BranchData;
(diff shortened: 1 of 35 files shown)


src/Model/PhasorDynamics/Branch/Branch.hpp
```diff
@@ -51,21 +45,21 @@ namespace GridKit
...
-      using Component<ScalarT, IdxT>::ypB_;
-      using Component<ScalarT, IdxT>::fB_;
-      using Component<ScalarT, IdxT>::gB_;
-      using Component<ScalarT, IdxT>::param_;
+      // using Component<ScalarT, IdxT>::g_;
+      // using Component<ScalarT, IdxT>::yB_;
+      // using Component<ScalarT, IdxT>::ypB_;
+      // using Component<ScalarT, IdxT>::fB_;
+      // using Component<ScalarT, IdxT>::gB_;
+      // using Component<ScalarT, IdxT>::param_;
```
superwhiskers: why are these just commented out, is this incomplete? if so, it'd be better to add a note as to why. if they're unnecessary, then they should be deleted.

src/Model/PhasorDynamics/Bus/BusInfinite.cpp
```diff
@@ -66,10 +66,10 @@ namespace GridKit
...
      */
     template <class ScalarT, typename IdxT>
-    BusInfinite<ScalarT, IdxT>::BusInfinite(BusData& data)
-      : BusBase<ScalarT, IdxT>(data.bus_i),
-        Vr_(data.Vm * cos(data.Va)),
-        Vi_(data.Vm * sin(data.Va))
+    BusInfinite<ScalarT, IdxT>::BusInfinite(DataT& data)
+      : BusBase<ScalarT, IdxT>(data.bus_id),
+        Vr_(data.Vr0),
+        Vi_(data.Vi0)
```
superwhiskers, line 72: so we're replacing the usage of `PowerSystemData` with `BusData` but `PowerSystemData` still exists, right? and what was there previously (i'm thinking with regard to the `Vr_(data.Vm * cos(data.Va))` and `Vi_(data.Vm * sin(data.Va))`) is equivalent to what is replacing it (`Vr_(data.Vr0)`, `Vi_(data.Vi0)`). if these are in fact equivalent, then isn't this duplicating input parameters and potentially introducing another place where an error could stem from? maybe this isn't that big of a deal but without an explanation for the change here, it's not clear to me if this was just a placeholder value.

src/Model/PhasorDynamics/Bus/Bus.cpp
```diff
@@ -61,10 +61,10 @@ namespace GridKit
...
      */
     template <class ScalarT, typename IdxT>
-    Bus<ScalarT, IdxT>::Bus(BusData& data)
-      : BusBase<ScalarT, IdxT>(data.bus_i),
-        Vr0_(data.Vm * cos(data.Va)),
-        Vi0_(data.Vm * sin(data.Va))
+    Bus<ScalarT, IdxT>::Bus(DataT& data)
+      : BusBase<ScalarT, IdxT>(data.bus_id),
+        Vr0_(data.Vr0),
+        Vi0_(data.Vi0)
```
superwhiskers, line 67: same comment as the one for `BusInfinite`.

src/Model/PhasorDynamics/Load/Load.hpp
```diff
@@ -33,16 +36,17 @@ namespace GridKit
...
-      using Component<ScalarT, IdxT>::ypB_;
-      using Component<ScalarT, IdxT>::fB_;
-      using Component<ScalarT, IdxT>::gB_;
-      using Component<ScalarT, IdxT>::param_;
+      // using Component<ScalarT, IdxT>::g_;
+      // using Component<ScalarT, IdxT>::yB_;
+      // using Component<ScalarT, IdxT>::ypB_;
+      // using Component<ScalarT, IdxT>::fB_;
+      // using Component<ScalarT, IdxT>::gB_;
+      // using Component<ScalarT, IdxT>::param_;
```
superwhiskers: same comment as the one on `Branch.hpp`.

superwhiskers (changes requested):

src/Model/PhasorDynamics/Bus/BusInfinite.cpp
pelesh: Thanks for pointing out the confusion in the naming of GridKit objects. `PowerSystemsData` contains data for power flow models only, that is why it is moved to `src/Model/PowerFlow`.

Changes you pointed to refer to phasor domain dynamics models, which require different data structure.

pelesh: Renamed `PowerSystemData` -> `PowerFlowData` in [042ebcd](https://github.com/ORNL/GridKit/pull/116/commits/042ebcdf911e05120837bdb97ee89567f2c2748e). I think it is less confusing now.

pelesh: For phasor domain dynamics, we use rectangular instead of polar coordinates.

src/Model/PhasorDynamics/Bus/Bus.cpp
pelesh: See above.

src/Model/PhasorDynamics/Load/Load.hpp
pelesh: Addressed in [8fcb325](https://github.com/ORNL/GridKit/pull/116/commits/8fcb3255e3aced03440edce7b27c59fe30b7e357).

src/Model/PhasorDynamics/Branch/Branch.hpp
pelesh: Addressed in [8fcb325](https://github.com/ORNL/GridKit/pull/116/commits/8fcb3255e3aced03440edce7b27c59fe30b7e357).

superwhiskers (approved):

PR #117, opened 2025-05-28
fix a memory leak, add asan / ubsan support to the cmake configuration

superwhiskers:
as the title says

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,22 @@ option(GRIDKIT_ENABLE_SUNDIALS_SPARSE "Enable SUNDIALS sparse linear solvers" ON
 # Enzyme support is disabled by default
 option(GRIDKIT_ENABLE_ENZYME "Enable automatic differentiation with Enzyme" OFF)
 
+# Enable the address sanitizer
+option(GRIDKIT_ENABLE_ASAN "Enable the address sanitizer" OFF)
+
+if(GRIDKIT_ENABLE_ASAN)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+endif()
+
+# Enable the undefined behavior sanitizer
+option(GRIDKIT_ENABLE_UBSAN "Enable the undefined behavior sanitizer" OFF)
+
+if(GRIDKIT_ENABLE_UBSAN)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -37,6 +37,22 @@ option(GRIDKIT_ENABLE_SUNDIALS_SPARSE "Enable SUNDIALS sparse linear solvers" ON
...
+endif()
+
+# Enable the undefined behavior sanitizer
+option(GRIDKIT_ENABLE_UBSAN "Enable the undefined behavior sanitizer" OFF)
+
+if(GRIDKIT_ENABLE_UBSAN)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined -fno-omit-frame-pointer")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=undefined")
+endif()
+
```
pelesh, line 56: Consider making these flags part of an interface target, then link that target to GridKit libraries.

When we are at it, we could add a "developer mode" compiler flags that include `-Wall -Wextra -Wconversion -pedantic`.

pelesh (approved):
Thanks for addressing this. I made some optional suggestions.

examples/PowerElectronics/DistributedGeneratorTest/DGTest.cpp
```diff
@@ -89,5 +89,7 @@ int main(int /* argc */, char const** /* argv */)
   }
   std::cout << "Test successful!\n";
 
+  delete dg;
```
alexander-novo: I personally would prefer a smart pointer here. If we ever push linting through, this will be a lint warning.

superwhiskers: absolutely. this was just the minimal necessary change to make

Steven-Roberts: Can it just be created on the stack?
```
GridKit::DistributedGenerator<double, size_t> dg(0, parms, true);
```

pelesh: Yes, it can. For the examples, that would be my preferred way to go about it.

superwhiskers: i've moved it to the stack

CMakeLists.txt
superwhiskers: users may want to enable address sanitizer or ubsan independent of the target, so i don't think it's worthwhile to do that. however, i have added those compile flags you suggested to the default list of compile flags and fixed all warnings resulting from them in the current tree

pelesh (approved):
lgtm

PR #123, opened 2025-06-02
fix the remaining warnings in the codebase

superwhiskers:
this pull request fixes the remaining warnings resulting from the addition of `-Wall -Wextra -Wconversion -Wpedantic` to the default cxxflags. these were missed in the initial pull request due to the source files they're in being gated behind the `GRIDKIT_ENABLE_IPOPT` define.

this pull request also removes some unnecessary commented out flags in the root `CMakeLists.txt` file due to them being superseded by the default flag list and also being covered by the `RelWithDebInfo` and `Debug` `CMAKE_BUILD_TYPE` values (iirc).

### DIFF
--- a/examples/PowerElectronics/RLCircuit/RLCircuit.cpp
+++ b/examples/PowerElectronics/RLCircuit/RLCircuit.cpp
@@ -37,7 +37,7 @@ int main(int /* argc */, char const** /* argv */)
   //  input
   induct->setExternalConnectionNodes(0, 1);
   // output
-  induct->setExternalConnectionNodes(1, -1);
+  induct->setExternalConnectionNodes(1, static_cast<size_t>(-1));
   // internal
   induct->setExternalConnectionNodes(2, 2);
   // add component
@@ -59,7 +59,7 @@ int main(int /* argc */, char const** /* argv */)
   GridKit::VoltageSource<double, size_t>* vsource = new GridKit::VoltageSource<double, size_t>(idoff, vinit);
   // Form index to node uid realations
   // input
-  vsource->setExternalConnectionNodes(0, -1);
+  vsource->setExternalConnectionNodes(0, static_cast<size_t>(-1));
   // output
   vsource->setExternalConnectionNodes(1, 0);
   // internal
--- a/src/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
+++ b/src/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
@@ -189,14 +189,15 @@ inline std::vector<IdxT> COO_Matrix<ScalarT, IdxT>::getCSRRowData()
 {
   if (!this->isSorted())
     this->sortSparse();
-  std::vector<IdxT> row_size_vec(this->rows_size_ + 1, 0);
+  std::vector<IdxT> row_size_vec(static_cast<size_t>(this->rows_size_ + 1), 0);
   IdxT              counter = 0;
   for (IdxT i = 0; i < static_cast<IdxT>(row_size_vec.size() - 1); i++)
   {
-    row_size_vec[i + 1] = row_size_vec[i];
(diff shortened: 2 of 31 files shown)


superwhiskers:
managed to unearth some more warnings, fixing those first

superwhiskers:
this should cover all of them. all tests pass. no warnings are shown anymore

src/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
```diff
@@ -189,14 +189,15 @@ inline std::vector<IdxT> COO_Matrix<ScalarT, IdxT>::getCSRRowData()
 {
   if (!this->isSorted())
     this->sortSparse();
-  std::vector<IdxT> row_size_vec(this->rows_size_ + 1, 0);
+  std::vector<IdxT> row_size_vec(static_cast<size_t>(this->rows_size_ + 1), 0);
   IdxT              counter = 0;
   for (IdxT i = 0; i < static_cast<IdxT>(row_size_vec.size() - 1); i++)
```
alexander-novo, line 194: I think as discussed in the original issue, these should just be changed to `size_t` rather than `static_cast`ing them everywhere they're used. `IdxT` should be used for sparse indices rather than for only indexing stl containers

src/Model/PowerElectronics/MicrogridLoad/MicrogridLoad.cpp
```diff
@@ -44,9 +44,9 @@ namespace GridKit
   template <class ScalarT, typename IdxT>
   int MicrogridLoad<ScalarT, IdxT>::allocate()
   {
-    y_.resize(size_);
-    yp_.resize(size_);
-    f_.resize(size_);
+    y_.resize(static_cast<size_t>(size_));
+    yp_.resize(static_cast<size_t>(size_));
+    f_.resize(static_cast<size_t>(size_));
```
pelesh, line 49: Let's make it a little less cluttered. Consider something like this here and in similar places throughout the code.
```suggestion
    auto size = static_cast<size_t>(size_);
    y_.resize(size);
    yp_.resize(size);
    f_.resize(size);
```

examples/PowerElectronics/RLCircuit/RLCircuit.cpp
```diff
@@ -37,7 +37,7 @@ int main(int /* argc */, char const** /* argv */)
   //  input
   induct->setExternalConnectionNodes(0, 1);
   // output
-  induct->setExternalConnectionNodes(1, -1);
+  induct->setExternalConnectionNodes(1, static_cast<size_t>(-1));
```
nkoukpaizan, line 40: Is this conversion meaningful? This will store the largest integer, right? It works because the same casting is done for the variable this is compared to, e.g., https://github.com/ORNL/GridKit/blob/12fa7be8483f8738af968fb3778ed69d2fa195b3/src/Model/PowerElectronics/SystemModelPowerElectronics.hpp#L177

https://github.com/ORNL/GridKit/blob/12fa7be8483f8738af968fb3778ed69d2fa195b3/src/Model/PowerElectronics/SystemModelPowerElectronics.hpp#L333

but I don't think it is the intended conceptual purpose of using -1.
@reid-g should chime in here.

superwhiskers: it should just store the largest integer `size_t` can hold. i went with this as i wasn't totally sure how invasive we should be here (same applies to the other comments). i would much rather
- either set it to the maximum `size_t`
- or adjust the api to provide another way to communicate the same intent (`std::optional` or something)

pelesh: I think this is okay as is. The index -1 stands for the ground node. It only needs to be a very distinctive integer. 

CC @reid-g

superwhiskers:
moved to #125

PR #126, opened 2025-06-04
Reorganization in preparation for Enzyme automatic differentiation PR

nkoukpaizan:
## Description
 
Reorganization in preparation for Enzyme automatic differentiation PR.
 
 ## Proposed changes
 
 - Moved `src/LinearAlgebra/SparsityPattern/` to `src/AutomaticDifferentiation/DependencyTracking/`
 - Placed `COO_Matrix` into the `GridKit::LinearAlgebra` namespace.
 - Added `~enzyme` spec to CI. Since Enzyme is an optional dependency for now, we need to ensure we can still build without it.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

This PR should not be a heavy load to review.
This is an attempt to break up #124 into more manageable PRs. I am working on disabling Enzyme derivatives and related tests when `GRIDKIT_ENABLE_ENZYME=Off`.


src/Model/Evaluator.hpp
```diff
@@ -83,8 +83,8 @@ namespace GridKit
       virtual std::vector<ScalarT>&       getResidual()       = 0;
       virtual const std::vector<ScalarT>& getResidual() const = 0;
 
-      virtual COO_Matrix<ScalarT, IdxT>&       getJacobian()       = 0;
-      virtual const COO_Matrix<ScalarT, IdxT>& getJacobian() const = 0;
+      virtual GridKit::LinearAlgebra::COO_Matrix<ScalarT, IdxT>&       getJacobian()       = 0;
+      virtual const GridKit::LinearAlgebra::COO_Matrix<ScalarT, IdxT>& getJacobian() const = 0;
```
pelesh: This should really not be in the `Evaluator` class and we should not use `COO_Matrix` outside the power electronics module. I suggest just adding a TODO note here.

pelesh: I think we should use Ipopt approach here and have this function set pointers to matrix data, matrix size and NNZ rather than returning an object. Alternatively, we could create a lightweight matrix object similar to what we've done in Re::Solve.

pelesh (changes requested):
Looks good! Just rebase and fix any compiler warnings. I also suggest to add a todo note in `Model/Evaluator.hpp`

pelesh (approved):
Builds cleanly, tests pass, good to go.

PR #134, opened 2025-06-10
fix several memory leaks

superwhiskers:
this fixes several memory leaks i found in the progress of reviewing an unrelated pull request. my bandage solution to fixing the leaked memory caused by repeated calls to `initializeBackwardSimulation` may not be the best way of fixing that problem, so alternative solutions are welcome

### DIFF
--- a/examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
+++ b/examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
@@ -20,14 +20,14 @@ int main()
   using namespace GridKit::Testing;
 
   // Create an infinite bus
-  std::unique_ptr<BaseBus<double, size_t>> bus = std::make_unique<BusSlack<double, size_t>>(1.0, 0.0);
+  BusSlack<double, size_t> bus(1.0, 0.0);
 
   // Attach a generator to that bus
-  Generator4<double, size_t> gen(bus.get());
+  Generator4<double, size_t> gen(&bus);
 
   // Create a system model
   SystemModel<double, size_t> model;
-  model.addBus(bus.get());
+  model.addBus(&bus);
   model.addComponent(&gen);
 
   // allocate model components
--- a/examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
+++ b/examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
@@ -1,6 +1,6 @@
-
 #include <iomanip>
 #include <iostream>
+#include <memory>
 
 #include <IpIpoptApplication.hpp>
 #include <IpSolveStatistics.hpp>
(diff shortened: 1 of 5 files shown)


src/Solver/Dynamic/Ida.cpp
```diff
@@ -426,6 +426,17 @@ namespace AnalysisManager
...
+
+      if (JacobianMatB_ != nullptr || linearSolverB_ != nullptr)
+      {
+        // guard against repeated initialization
+        SUNLinSolFree(linearSolverB_);
+        SUNMatDestroy(JacobianMatB_);
+
+        linearSolverB_ = nullptr;
+        JacobianMatB_  = nullptr;
+      }
```
pelesh: Must make these separate checks.

CC @Steven-Roberts

src/Solver/Dynamic/Ida.hpp
```diff
@@ -150,9 +150,9 @@ namespace AnalysisManager
       void*           solver_{};
       SUNContext      context_{};
       SUNMatrix       JacobianMat_{};
-      SUNMatrix       JacobianMatB_{};
+      SUNMatrix       JacobianMatB_ = nullptr;
       SUNLinearSolver linearSolver_{};
-      SUNLinearSolver linearSolverB_{};
+      SUNLinearSolver linearSolverB_ = nullptr;
```
pelesh: Either make all of these `nullptr` or make no changes.

pelesh (changes requested):
Thanks for catching this. A couple of minor changes requested.

src/Solver/Dynamic/Ida.cpp
superwhiskers: if i've interpreted the surrounding context correctly, if the jacobian matrix is not a nullptr then we'd need to reset both the linear solver and the jacobian matrix, but if only the linear solver is allocated, then we can do that individually, correct?

src/Solver/Dynamic/Ida.hpp
superwhiskers: assuming i interpreted what you meant here correctly, this should be resolved now

Steven-Roberts: The brace initialize and `= nullptr` are essentially equivalent. I think for consistency here and with KINSOL, we can just leave this as it was.

superwhiskers: done. removed the `= nullptr` bit and tested it to ensure it worked.

src/Solver/Dynamic/Ida.cpp
pelesh: The two are allocated independently, so it is possible to have one and not the other.

pelesh: I also believe there are SUNDIALS methods to check if the matrix and/or linear solver are instantiated. @Steven-Roberts would know more about that than me.

Steven-Roberts: There is a `deleteBackwardSimulation` function which can free these objects. I think this is intended to be called after `initializeBackwardSimulation` and the adjoint solution has been computed, but correct me if I'm wrong @pelesh. If we don't want to rely on user code calling `deleteBackwardSimulation`, we can call that function at the top of `initializeBackwardSimulation`.

superwhiskers: i'm pretty sure i tried this already and since it frees three other vectors, i got a null pointer dereference or something

pelesh: After looking at the code, this is not an easy fix. I suggest creating a separate issue and removing this change altogether from this PR.
```suggestion
```

pelesh: Thanks, @superwhiskers, for identifying this issue.

Steven-Roberts: >I also believe there are SUNDIALS methods to check if the matrix and/or linear solver are instantiated. @Steven-Roberts would know more about that than me.

I'm not aware of one. Checking if it is null is fine, although if it's going to be freed if non-null, you can directly call the SUNDIALS free/destroy method without wrapping it in an if statement.

> i'm pretty sure i tried this already and since it frees three other vectors, i got a null pointer dereference or something

I see. We could merge the `configureAdjoint` and `initializeBackwardSimulation` functions to fix that. Another option is not to free and reallocate the matrix and linear solver each backward integration. In `initializeBackwardSimulation` we can only create the matrix and linear solver if they are null.

examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
```diff
@@ -20,49 +20,49 @@ int main()
   using namespace GridKit::Testing;
 
   // Create an infinite bus
-  BaseBus<double, size_t>* bus = new BusSlack<double, size_t>(1.0, 0.0);
+  std::unique_ptr<BaseBus<double, size_t>> bus = std::make_unique<BusSlack<double, size_t>>(1.0, 0.0);
```
Steven-Roberts: Can we just initialize this on that stack?

src/Solver/Dynamic/Ida.cpp
pelesh: I think this is the best solution in this case.

Steven-Roberts: It would be handy to have an example that performs several backward integrations so we can settle on what functions get called in user code and where objects need to be allocated/freed.

pelesh: I think this will do it in the short term:
```c++
      if (JacobianMatB_ == nullptr)
      {
        JacobianMatB_ = SUNDenseMatrix(static_cast<sunindextype>(model_->size()),
                                       static_cast<sunindextype>(model_->size()),
                                       context_);
        checkAllocation((void*) JacobianMatB_, "SUNDenseMatrix");
      }
      if (linearSolverB_ == nullptr)
      {
        linearSolverB_ = SUNLinSol_Dense(yyB_, JacobianMatB_, context_);
        checkAllocation((void*) linearSolverB_, "SUNLinSol_Dense");
      }
```

pelesh: In [29d3221](https://github.com/ORNL/GridKit/pull/134/commits/29d322119c87f190cfcca44f32faef2dd66fcda4).

examples/Experimental/GenInfiniteBus/GenInfiniteBus.cpp
pelesh: Changed in [15546ee](https://github.com/ORNL/GridKit/pull/134/commits/15546eea1793490db8abafb4bd749101d883b79a).

pelesh (approved):

PR #150, opened 2025-06-23
add a json parser for the grid dynamics case format

superwhiskers:
this pull request adds a parser for the grid dynamics case format
 
 # checklist

 
- [x] All tests pass.


superwhiskers:
i've written the dummy test and it appears to parse. i still need to poke the parsed data a little closer but it appears to work correctly

superwhiskers:
supplanted by #159

PR #159, opened 2025-06-26
add a json parser for the grid dynamics case format (nlohmann/json version)

superwhiskers:
this pull request adds a parser for the grid dynamics case format using https://github.com/nlohmann/json instead of rapidjson, so this supplants #150.

 ## checklist
 
- [x] All tests pass.


src/Model/PhasorDynamics/INPUT_FORMAT.md
```diff
@@ -113,7 +113,7 @@ are specified:
   --------------|------------------------------------------------------|-----------------------------------------------|---------------------------- | -------------------------
   `branch`      | a basic algebraic pi model for a line or transformer | `bus1`, `bus2`                             | `R`, `X`, `G`, `B` | `ir1`, `ii1`, `im1`, `p1`, `q1`, `ir2`, `ii2`, `im2`, `p2`, `q2`
   `static_load` | a basic static ZIP load                              | `bus`                                      | `Pz`, `Qz`, `Pi`, `Qi`, `Pp`, `Qp` | `ir`, `ii`, `p`, `q`
-  `GENROU`      | 6th order machine model                              | `bus`, `exciter_signal`\*, `governor_signal`\* | `p0`, `q0`, `H`, `D`, `Ra`, `Tdop`, `Tdopp`, `Tqopp`, `Tqop`, `Xd`, `Xdp`, `Xdpp`, `Xq`, `Xqp`, `Xqpp`, `Xl`, `S10`, `S12` | `ir`, `ii`, `p`, `q`, `delta`, `omega`
+  `GENROU`      | 6th order machine model                              | `bus`, `exciter_signal`\*, `governor_signal`\* | `unit_id`, `p0`, `q0`, `H`, `D`, `Ra`, `Tdop`, `Tdopp`, `Tqopp`, `Tqop`, `Xd`, `Xdp`, `Xdpp`, `Xq`, `Xqp`, `Xqpp`, `Xl`, `S10`, `S12` | `ir`, `ii`, `p`, `q`, `delta`, `omega`
```
abirchfield: `unit_id` is not a parameter. It is the same as the `id` that all components have.

src/Model/PhasorDynamics/INPUT_FORMAT.md
```diff
@@ -123,25 +123,25 @@ connected to a constant value. This list is subject to change.
...
-        { "number": 1, "class": "bus", "name": "Bus 1", "init": {"Vr":0.994988, "Vi":0.099997}, "vbase": 115e3, "mon": ["Vr", "Vi"] },
-        { "number": 2, "class": "infinite_bus", "name": "Bus 2", "init": {"Vr":1.0, "Vi":0}, "vbase": 115e3 }
-    ],
-    "devices": [
-        { "class": "branch", "ports": {"bus1":1, "bus2":2}, "id": "1", "params": {"R":0, "X":0.1, "G":0, "B":0} },
-        { "class": "GENROU", "ports": {"bus":1}, "id": "1", "params": {"p0":1, "q0":0.05013, "H":3, "D":0, "Ra":0, "Tdop":7, "Tdopp":0.04, "Tqopp":0.05,
-               "Tqop":0.75, "Xd":2.1, "Xdp":0.2, "Xdpp":0.18, "Xq":0.5, "Xqpp":0.18, "Xl":0.15, "S10":0, "S12":0}, "mon": ["delta", "omega"] }
-        { "class": "bus_fault", "ports": {"bus":1}, "id": "1", "params": {"state0":0, "R":0, "X":1e-3} }
-    ]
+   "header": {
```
abirchfield, line 126: In my opinion, all component parameters should be scalar and presumed to be scalar/double. A user is not going to remember to write `60.0` instead of `60` to make sure it is not interpreted as an integer. And it is very typical to use 1 and 0 to specify binary parameter choices rather than true and false.

src/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -6,10 +6,19 @@
  */
 #pragma once
 
+#include <bitset>
+#include <optional>
+#include <string>
+#include <type_traits>
+
+#include <nlohmann/json.hpp>
+
```
pelesh, line 15: The data file should be agnostic of the parser. Ideally, it should allow different parsers for different file formats to be deployed and populate the same data structures. It might be a good idea to separate this into the component data file and the JSON parser for that component. The downside of this suggestion is that we may have a proliferation of many files needed to implement a single component.

src/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -23,13 +32,157 @@ namespace GridKit
...
+        IM2,
+        P2,
+        Q2,
+        MAXIMUM,
+      };
+
+      /// Set indicating the variables being monitored
+      std::bitset<static_cast<
+          std::underlying_type_t<MonitorableVariables>>(MonitorableVariables::MAXIMUM)>
+          monitored_variables;
```
pelesh, line 67: I am not sure how this is supposed to be used but we can implement one usage example or a unit test where this feature is used.

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = buildsystem/spack/spack
 	url = https://github.com/nkoukpaizan/spack.git
 	branch = Gridkit-package-dev+enzyme
+[submodule "third-party/nlohmann-json"]
+	path = third-party/nlohmann-json
+	url = https://github.com/nlohmann/json
```
abirchfield, line 7: Can you provide instructions for installing this third party package? Right now I cannot build gridkit.

src/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -23,13 +32,157 @@ namespace GridKit
...
     };
+
+    template <typename RealT, typename IdxT>
+    void from_json(const json& j, BranchData<RealT, IdxT>& bd)
+    {
+      for (auto& raw_parameter : j.at("params").items())
+      {
```
pelesh, line 77: I think it would be good to document in an excessive detail what these JSON parser functions do. For example:
- What is the return type of `items` method?
- What is the return type/signature of `at` method.
- Do we know what is time complexity of `at`? Can we use `[]` operator instead?

It would be good to have comments in the code explaining what these functions do and perhaps having a README file providing some guidelines for usage and maintenance of the parser.

src/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -23,13 +32,157 @@ namespace GridKit
...
+          raw_parameter.value().get_to(bd.R);
+        }
+        else if (raw_parameter.key() == "X")
+        {
+          raw_parameter.value().get_to(bd.X);
+        }
+        else if (raw_parameter.key() == "G")
+        {
+          raw_parameter.value().get_to(bd.G);
+        }
```
pelesh, line 89: Is it possible with this parser to do something like this:
```c++
// Copy value of parameter G to the data structure bd
j["params"]["G"].get_to(bd.G);
```

src/Model/PhasorDynamics/Bus/BusData.hpp
```diff
@@ -23,18 +32,125 @@ namespace GridKit
...
-        INVALID = 0,
         DEFAULT,
-        SLACK
+        SLACK,
       };
 
-      RealT Vr0{0.0}; ///< Initial value for real bus voltage
-      RealT Vi0{0.0}; ///< Initial value for imaginary bus voltage
+      BusType bus_type{BusType::DEFAULT}; ///< The kind of bus this data is for
+
```
pelesh: The default value for the bus type should be `INVALID`. We want code to crash as soon as possible if the bus type was not decided by the user.

abirchfield (commented):
Overall I think this is a step in the right direction for file I/O in gridkit. I left a few comments below. I did finally get the JSON parser to work after cloning the source and doing sudo make install, but I don't know if that is the recommended way.

I have another general comment. I would repeat my preference that there be a common ComponentData object that can support data from all types of components, which is passed to the instructor of each individual component. And then we delete all these BranchData, GenrouData, LoadData, etc. Otherwise a lot of the code for parsing the JSON is going to be essentially repeated for each of the component data classes. For hundreds of component classes (in the future) this is going to be quite cumbersome and error-prone. Instead there should be one ComponentData class which stores a map<string, double> for the params, a vector<string> for the monitored variables, etc. But this is something that can be discussed or decided in the future.

.gitmodules
superwhiskers: you need to run `git submodule update --init` to clone the submodule. you don't need to install it with `make install` or anything

src/Model/PhasorDynamics/INPUT_FORMAT.md
superwhiskers: `unit_id` is a new parameter that is separate from the `id` disambiguator. it's a unique identifier of a genrou component across all genrou components and just using `id` would not be enough as that 1) may not be unique across all genrou components and 2) is a string. @pelesh was the one who requested this iirc

pelesh (commented):
This is very nice first stab at the input file parser. My main suggestion is to document parser methods in detail and perhaps add a README file with instructions how to use it. 

Also consider putting parser files and documentation in a subdirectory in `PhasorDynamics`.

pelesh:
> I would repeat my preference that there be a common ComponentData object that can support data from all types of components, which is passed to the instructor of each individual component. And then we delete all these BranchData, GenrouData, LoadData, etc. Otherwise a lot of the code for parsing the JSON is going to be essentially repeated for each of the component data classes. For hundreds of component classes (in the future) this is going to be quite cumbersome and error-prone. Instead there should be one ComponentData class which stores a `map<string, double>` for the params, a vector for the monitored variables, etc. But this is something that can be discussed or decided in the future.

This implementation is deliberately redundant to give us more flexibility. The main objective is for component models to be insulated from the specifics of the I/O file parser. The next is to have a thin and generic abstraction layer between the parser and the components, along the lines you described.

src/Model/PhasorDynamics/INPUT_FORMAT.md
pelesh: Question for @abirchfield: Do you think each component (generator, load, etc.) should have a unique ID or there should be component unique ID and in addition to that a unique generator ID? The latter could be a name of the facility or an integer ID as far as I understand. 

If you think this is redundant at this point in development, I am fine with dropping it.

src/Model/PhasorDynamics/INPUT_FORMAT.md
pelesh: I think this is how it works now. My understanding is the data is converted to the right format by overloading function `get_to`. The conversion to the right data type will be from string to the type of the argument of `get_to`, regardless of how the value is stored in JSON file (e.g with or without decimal point).

src/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: the unit test i added uses it already. what do you have in mind for a more specific one?

src/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: `.at` does bounds checking, whereas `operator[]` does not. all of this is available on the documentation website for this json parser; adding comments here would clutter the source i think

the api documentation for `.items()` is here, for example: https://json.nlohmann.me/api/basic_json/items/

src/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: yes but that wouldn't have bounds checking. we could just replace everything with that sort of thing, but i wanted to throw an error if an unrecognized parameter was found. if that isn't the behavior we want, though, i can change the approach i used.

src/Model/PhasorDynamics/Bus/BusData.hpp
superwhiskers: why is that bus type called `DEFAULT`, then

src/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: i think co-locating the parser handling with the data itself is more desirable. if what you have in mind is removing the include if we want to disable json parsing, then we can add a define here but i don't see any real downside to doing things the way we have right now.

src/Model/PhasorDynamics/INPUT_FORMAT.md
abirchfield: Yes, I think we can drop unit_id for now. The main purpose in other programs is to disambiguate multiple components at a bus.

.gitmodules
abirchfield: OK thanks

src/Model/PhasorDynamics/INPUT_FORMAT.md
superwhiskers: we should probably drop it from the `Genrou` class itself as well, but i've removed it and filled in placeholder values in the constructor for `Genrou` that uses the `GenrouData` structure

src/Model/PhasorDynamics/INPUT_FORMAT.md
superwhiskers: we can't use `0`/`1` for a boolean but everything else does work as you would like from what i can tell

superwhiskers:
i think this should be good to merge now

src/Model/PhasorDynamics/Branch/BranchData.hpp
pelesh: Please address this comment. Having a README file with a glossary of parser methods with explanations why they are used and links to the upstream documentation would be helpful.

pelesh (changes requested):
Doxygen comments and some documentation explaining features of the JSON parser we use here are needed.

Otherwise, it is ready to merge.

src/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: i put this in a comment already within `SystemModelData.hpp`. i'll move it out though

superwhiskers: done. i think this should resolve it.

pelesh (approved):

superwhiskers:
rebased

PR #105, opened 2025-05-09
SUNDIALS updates

Steven-Roberts:
## Description
 
Replaces some deprecated SUNDIALS calls, adds a missing `N_VDestroy`, fixes uninitialized member variables that could cause issues when freed, and simplifies some of the code in the Ida and Kinsol classes. There's also some TODOs I left with questions @pelesh 

 ## Checklist
 
- [x] All tests pass.
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [ ] There are unit tests for the new code.
- [x] The new code is documented.
- [ ] The feature branch is rebased with respect to the target branch.

PR #160, opened 2025-06-27
Update to SUNDIALS 7.4.0

Steven-Roberts:
## Description
 
Updates to latest version of SUNDIALS which includes a bug fix for IDA that was previously causing a seg fault in #105
 
 ## Checklist
 
- [ ] All tests pass.

### DIFF
--- a/.github/workflows/spack_default_build.yaml
+++ b/.github/workflows/spack_default_build.yaml
@@ -142,7 +142,7 @@ jobs:
               mumps:
                 require: ~mpi~openmp
               sundials:
-                require: '@develop'
+                require: '@7.4.0'
               llvm:
                 externals:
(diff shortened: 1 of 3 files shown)


superwhiskers (approved):
i've tested this on my machine with sundials 7.4.0. i'm unable to run the enzyme tests as it looks like i need to mess around with enzyme a little to get it to work right but all of the tests i am able to run pass without issue. this should be fine to merge.

Steven-Roberts:
I'll need to wait to update
https://github.com/ORNL/GridKit/blob/fd3dfc608cf9fa338507cede5df2c95297bf1944/.github/workflows/spack_default_build.yaml#L144-L145
until the SUNDIALS spack package gets updated, hopefully in a couple days.

Issue #147, opened 2025-06-19
Organize phasor dynamics examples


pelesh:
### Summary

Provide more intuitive names and directory structure of phasor dynamics examples.

### Rationale

Current structure has examples listed as Example1, Example2, ...., which does not say much to the user. One has to open the example and look at the code to understand what the example is about.

### Description

Consider creating subdirectories like:
- Toy problems
- IEEE systems
- Scalable cases
- etc.

Inside we can have names like Case3Bus, Ieee9Bus, etc.

The directory structure and naming scheme should be aligned by domain conventions.

pelesh:
Looking at #162, let me make some suggestions to standardize example names and the names of subdirectories in `examples` directory:
- Use camel format, e.g. `TenGenClassical` instead of a mixture of different formats like `TenGenClassical`.
- I would avoid using "example" in example names and generally avoid repeating information that is obvious from the names of directories where the examples reside.
- I would use the same name for the executable as for the source code. For example, CMake target for `TenGenClassical.cpp` should be `TenGenClassical.exe`.

pelesh:
Fixed in #162.

PR #162, opened 2025-06-27
Organizing Phasor Dynamics Examples

abirchfield:
## Description
 
Organizing existing phasor dynamics examples with more intuitive directory structure and descriptive names, with lots of room for expansion adding more examples.
 
 _Closes #147 
  
 ## Proposed changes
 
Four major categories of cases: Tiny, Small, Medium, and Large.  Inside each there should be a directory for each system, within which there will be specific examples. 

Right now in Tiny we have TwoBus and ThreeBus, which previously were Example1 and Example2.

Right now in Small we have TenGen.

Right now we do not have any Medium or Large examples, but will be adding them soon. We need to get the JSON format first before that will be worth it.

This PR just moves the files and renames things.

Out of scope: adding more documentation to prior examples, new examples, validation data for the TenGen case.
 
 ## Checklist
 
- [X] All tests pass.


abirchfield:
(rebasing)

examples/PhasorDynamics/Small/TenGen/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(TenGen_GenClassical)
+add_subdirectory(TenGen_Genrou)
```
pelesh: Consider less redundant directory names such as
- `examples/PhasorDynamics/Small/TenGen/Classical` and
- `examples/PhasorDynamics/Small/TenGen/Genrou`

The directory names are as informative but shorter and without underscores.

examples/PhasorDynamics/Small/TenGen/TenGen_GenClassical/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
...
+add_executable(phasordynamics_example_tengen_genclassical Example_TenGen_GenClassical.cpp)
+target_link_libraries(phasordynamics_example_tengen_genclassical
 	GRIDKIT::phasor_dynamics_bus 
 	GRIDKIT::phasor_dynamics_bus_fault 
 	GRIDKIT::phasor_dynamics_branch
 	GRIDKIT::phasor_dynamics_gen_classical
 	GRIDKIT::phasor_dynamics_load
 	GRIDKIT::solvers_dyn)
-install(TARGETS phasordynamics_example3_genclassical RUNTIME DESTINATION bin)
+install(TARGETS phasordynamics_example_tengen_genclassical RUNTIME DESTINATION bin)
```
pelesh: By the same measure, I think we can shorten target names without losing information:
```suggestion
install(TARGETS phasordynamics_tengen_classical RUNTIME DESTINATION bin)
```
The entire directory is with example only, so I would drop "example" from example names.

pelesh: Similar applies to a few other places in the PR.

examples/PhasorDynamics/Tiny/ThreeBus/ThreeBus_Basic/Example_ThreeBus_Basic.cpp
```diff
@@ -8,7 +8,7 @@
  * compares results with data generated for the same system by Poweworld.
  *
  */
-#include "example2.hpp"
+#include "Example_ThreeBus_Basic.hpp"
```
pelesh: Consider dropping "Example" from the example name. I would also use camel case and drop underscores.
```suggestion
#include "ThreeBusBasic.hpp"
```

examples/PhasorDynamics/Tiny/ThreeBus/ThreeBus_Basic/Example_ThreeBus_Basic.cpp
```diff
@@ -86,7 +86,7 @@ int main()
 
   auto error_allowed = static_cast<real_type>(1e-4);
 
-  std::cout << "Example 2 version 1\n";
+  std::cout << "Example_ThreeBus_Basic\n";
```
pelesh: For the output we could do something like this:
```suggestion
  std::cout << "Example: ThreeBusBasic\n";
```

examples/PhasorDynamics/Tiny/ThreeBus/ThreeBus_Basic/Example_ThreeBus_Basic.hpp
```diff
@@ -9,7 +9,7 @@
 
 // Columns:
 // Time, Gen 2 #1 Speed, Gen 3 #1 Speed, Bus 1 V pu, Bus 2 V pu, Bus 3 V pu
-namespace Example2
+namespace Example_ThreeBus_Basic
```
pelesh: I don't think we need namespaces in the examples. They are standalone application, which will not be called by other code.

examples/PhasorDynamics/Tiny/ThreeBus/ThreeBus_Basic/Example_ThreeBus_Basic.cpp
pelesh: I would suggest standardizing on camel case for all file names. In this case, we have a mix of camel case and underscores.

examples/PhasorDynamics/Tiny/TwoBus/TwoBus_TGov1/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
-add_executable(phasordynamics_example1_tgov1 example1_tgov1.cpp)
-target_link_libraries(phasordynamics_example1_tgov1 
+add_executable(phasordynamics_example_twobus_tgov1 Example_TwoBus_TGov1.cpp)
+target_link_libraries(phasordynamics_example_twobus_tgov1
```
pelesh: Target names need to be detailed so we don't get target name clashes. But in this case we could do something like this:
```suggestion
add_executable(TwoBusTGov1.exe TwoBusTGov1.cpp)
target_link_libraries(TwoBusTGov1.exe
```
I understand we inherited untidy collection of names from before, but as we are reorganizing, it might be a good idea to set some rules that we will follow going forward.

pelesh (approved):
This is good to merge imo.

I only make some nitpicking suggestions how to simplify (and hopefully standardize) naming of examples.

abirchfield:
Think I resolved everything with the new names

Issue #137, opened 2025-06-11
Testing issue: EnzymePowerElectronicsCheck test fails on arm arch


pelesh:
### Failed test outputs or steps to reproduce the failure(s)

`EnzymePowerElectronicsCheck` test fails on arm architecture with the error message below:

<details>

```
Test project /Users/55y/src/gridkit/build-clang
    Start 21: EnzymePowerElectronicsCheck
1/1 Test #21: EnzymePowerElectronicsCheck ......***Failed    0.01 sec
Result incorrect at line = 10, column = 11, obtained = -2.06027e-14, reference = 0, difference = 2.06027e-14
Result incorrect at line = 11, column = 10, obtained = 2.06027e-14, reference = -0, difference = 2.06027e-14
Dense matrix: Autodiff Jacobian
-1 -0 -0 -0 -9.4e-05 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0 
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 
-1 -0 -0 -0 -9.4e-05 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
0 0 0 0 -31.41 0 0 0 0 0 0 0 0 0 0 0 
0 0 0 0 0 -31.41 0 0 0 0 0 0 0 0 0 0 
-0 -0 -0 -0 -0 -0.0013 -0 -0 -0 -0 -0 -0 -1 -0 -0 -0 
-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -1 -0 -0 
0 0 0 0 0 -0.00013 420 0 0 0 -1 0 -0.1 -0.015708 0.75 0 
0 0 0 0 0 0 0 420 0 0 0 -1 0.015708 -0.1 0 0.75 
0 0 0 0 0 -1.44444 4.66667e+06 0 1.48148e+07 0 -11185.2 2.06027e-14 -1851.85 -174.533 8333.33 0 
0 0 0 0 0 0 0 4.66667e+06 0 1.48148e+07 -2.06027e-14 -11185.2 174.533 -1851.85 0 8333.33 
0 0 0 0 0 0 0 0 0 0 20000 0 0 314.159 -20000 0 
-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 20000 -314.159 -0 -0 -20000 
[...]

PR #163, opened 2025-06-27
fix warnings (again)

superwhiskers:
there's nothing much here. this just removes the warnings from the repository


pelesh (approved):
No warnings appear in clean builds with GCC and Clang. All tests pass except for the failure documented in #137.

PR #168, opened 2025-07-01
Fix sparse Jacobian allocation

pelesh:
## Description
 
In `Ida` class, sparse Jacobian is allocated to NxN elements (N being the size of the matrix) instead of number of nonzeros in the matrix. This leads to unnecessary memory complexity and sluggish performance for large systems. 
 
 
 ## Proposed changes
 
Allocate sparse Jacobian in `Ida` to NNZ.
 
 ## Checklist
 

- [x] All tests pass.
 
 ## Further comments


superwhiskers (approved):
i was just about to do this so i don't see anything wrong with it

PR #173, opened 2025-07-04
Signal Bus Implementation

### LINKED ISSUE #154: Add signal bus model to phasor dynamics family
### Summary

Create a signal bus model that stores a scalar value, which can be overwritten by one (and only one) component model and which can be read by multiple components.

CC @lukelowry @reid-g 

### Rationale

Signal bus is needed to abstract connections between electrical machines and control devices such as governors and exciters. Machine would update sensor readings (e.g. machine speed), which would be read by control devices (e.g. governor). Control devices would update actuator signals (e.g. mechanical power required), which would be read by a machine.

Both the machine and the controller would only have access to the signal bus and would not need to know about each other internal properties.

### Description

I think signal bus should inherit directly from `Model::Evaluator` class. It should store (memory address of) a single scalar value. Unlike electrical bus, it should discriminate between inlet and outlet. Inlet should be a non-constant accessor to the scalar value and should allow only one connection. Outlet is a const accessor.

To be able to test signal model, we should create constant signal source and signal reader blocks.

### Additional information

_No response_

lukelowry:
## Description
 
A preliminary BusSignal implementation that is a stepping point to a fully developed Bus Hierarchy. Supported signals can be specified by the Data structure input.
 
 Closes Issue #154 

**An exciter model implementation and validated example (see branch [IEEET1](https://github.com/ORNL/GridKit/tree/lukel/ieeet1_dev)) are ready for PR once this is merged**
 
 
 ## Proposed changes
 
- BusSignal Class inherits directly from BusBase, which still has virtual voltage/current functions. This is retained only because creating a seperate base grows the scope of this PR beyond what it was intended for. We can segment these models in the next PR.
- Signal connections can be specified in (Model)Data struct.
- SystemModel now auto implements governor.
- Component class has helper functions that implement 'safe' versions of signal read/write/init to make model implementation easier.
- Removes the GovernorBase class which was only added as a temporary solution.
 
 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments
 
The cmake of many filles had to be changed to include the governor program since I added it to the SystemModel. Is there a better way to have files that use SystemModel to not require all the same programs?

Future PR will need to create seperate bases: BusElectricalBase and BusSignalBase. This is a temporary solutions before these new bases are formed, because that task is large to undertake.
 
Forgive the large quantity of commits. I have a VM in addition to local WSL, so I am making frequent commits to the origin.


abirchfield (approved):

src/Model/PhasorDynamics/Bus/BusSignal.cpp
```diff
@@ -0,0 +1,82 @@
...
+  namespace PhasorDynamics
+  {
+
+
+    template <class ScalarT, typename IdxT>
+    BusSignal<ScalarT, IdxT>::BusSignal(const DataT& data)
+        : BusBase<ScalarT, IdxT>(data.bus_id)
+    {
+      size_ = 1;
+    }
```
pelesh: I don't think `BusSignal` should inherit from `BusBase` as there is almost no overlap in functionality. I suggest `BusSignal` to inherit from `Model::Evaluator` directly. This would make signal bus code leaner and would also simplify code for electrical buses in this PR.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
```diff
@@ -0,0 +1,129 @@
...
+       /**
+       * @brief A write operation only available
+       * to the source signal
+       * 
+       * @param value The value of the signal
+       */
+      void send(ScalarT& value) override
+      {
+        f_[0] += value;
+      }
```
pelesh: This should be assignment rather than assignment addition:
```suggestion
      void send(ScalarT& value) override
      {
        f_[0] = value;
      }
```
There is only one source that sends the signal.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
```diff
@@ -0,0 +1,129 @@
...
+        return y_[0];
+      }
+
+       /**
+       * @brief A write operation only available
+       * to the source signal
+       * 
+       * @param value The value of the signal
+       */
+      void send(ScalarT& value) override
```
pelesh: I would suggest either read/write or send/receive names for signal bus public methods.

pelesh (changes requested):
In this first cut I suggest we implement signal bus as a standalone class inheriting directly from `Model::Evaluator`.

src/Model/PhasorDynamics/Bus/BusSignal.cpp
abirchfield: We did discuss this internally, and there are two big reasons that having `SignalBus` inherit from `BaseBus` greatly simplifies this PR.

1. Storage in the SystemModel class. Right now the system stores two vectors of system objects: buses_ and components_. We could add a new signal_buses_ internal vector, but this affects the creation of the model (add_bus and add_component) and has potentially other side effects. In general, the SystemModel doesn't need to know the difference between the signal and electrical buses. Hence it is helpful to have a common base class.

2. System evaluation. Right now in the initialization process, the system residual process, and potentially other places in the system evaluation, the system has to iterate through both all the buses and all the components. If we add a third category for signal buses, we have to make sure that this is handled correctly in all those places, which will require significant revision to the system evaluator and may cause other side effects. I do not see any reason that these should be separated from the system point of view.

As it is, this PR requires very few changes to the SystemModel class.

We made the decision in the group discussion that we would have the signal objects own the variables associated with their values. Hence they are very similar to the electrical buses in the way they need to be handled by the system evaluator. I do think that in the future that we should have BaseElectricalBus and BaseSignal bus that separates the differing functions from a component point of view. (Both of these would need to inherit from BaseBus from a system point of view.) But this change would have additional side effects, and is not strictly necessary at the moment.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
abirchfield: We made the decision in the group discussion that we would have the signal objects own the variables associated with their values. Based on this paradigm, the residual equation for that value would be `f[i] = <calculated value> - x[i]`.  Following the pattern of the electrical buses, the residual calculation of the signal bus first sets the residual equal to `-x[i]`. Then the writing component should add in the calculated value. That is why it must be an addition. 

However, it is important that this function is not called more than once in a given residual calculation. In a future PR, we should add a flag internal to the signal bus to throw an exception if this occurs.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
abirchfield: Yes, I agree. I prefer read/write. "read" more clearly indicates a passive action that can be done more than once. As in the comment above, we need to make sure "write" is not done more than once.

src/Model/PhasorDynamics/Bus/BusSignal.cpp
lukelowry: Per @abirchfield reasons, I will keep the existing inheritance structure. This PR is too large for more changes without some more group discussion.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
lukelowry: We can change this in a later PR, but right now it is needed for the reasons @abirchfield discussed.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
lukelowry: I have refactored all uses of 'send' to 'write' for BusSignal. This also changed safeSend to safeWrite.

lukelowry:
Okay, I have performed the following:
- Rebased the branch
- Refactor 'send' to 'write'
- Use the new Port data structure to input the BusSignal index

After this merge, I will make the following Issues:
- Electrical and Signal Bus Base Classes
- Implement Governor with new Data Structure (it is currently using old struct)

Seeking further review @pelesh and @abirchfield . Hoping we unlock doors with this PR!

src/Model/PhasorDynamics/Bus/BusSignal.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+        y_[0] = value;
+        // if(is_initialized_)
+        //{
+        //   std::cout << "ERROR!";
+        // }
+        // else{
+        //   y_[0] = value;
+        //   is_initialized_ = true;
+        // }
```
superwhiskers, line 58: why is this commented out? it appears to implement the intended behavior but instead we always set `y_[0]` to the value provided.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
```diff
@@ -0,0 +1,127 @@
...
+      virtual int evaluateResidual() override;
+      virtual int evaluateJacobian() override;
+
+      /**
+       * @brief A one-time initialization function
+       * that can be called by any.
+       *
+       * @return state value of signal
+       */
+      void initial_value(ScalarT value) override
```
superwhiskers, line 47: this method should be called `initialValue` to follow the naming conventions we have.

src/Model/PhasorDynamics/Component.hpp
```diff
@@ -20,12 +21,63 @@ namespace GridKit
...
+      {
+        if (signal)
+        {
+          return signal->read();
+        }
+        else
+        {
+          return default_val;
+        }
+      }
```
superwhiskers, line 79: i'm not sure these would belong on the `Component` class to be totally honest. they seem to exclusively be helper functions with no need to access any private data present on a `Component`; i think they would make more sense as standalone functions in a `Utility.hpp` file or something present within `src/Model/PhasorDynamics`.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouData.hpp
```diff
@@ -39,6 +39,8 @@ namespace GridKit
     enum class GenrouPorts
     {
       bus,             ///< Unique ID of the connecting bus
+      pmech_signal,    ///< Unique ID of the governor Pmech signal.
+      speed_signal,    ///< Unique ID of the speed signal of the machine.
```
superwhiskers, line 43: we should also modify `src/Model/PhasorDynamics/INPUT_FORMAT.md` if we're adding new ports

src/Model/PhasorDynamics/SystemModel.hpp
```diff
@@ -124,12 +125,37 @@ namespace GridKit
...
+          else{
+            bus_index = gendata.ports.at(GenrouData<ScalarT, IdxT>::Ports::bus);
+            
+            auto* gen = new Genrou<ScalarT, IdxT>(
+              getBus(bus_index), 
+              nullptr,
+              nullptr,
+              gendata);
+            addComponent(gen);
           }
```
superwhiskers, line 158:
```suggestion
          bus_type* bus   = nullptr;
          bus_type* pmech = nullptr;
          bus_type* speed = nullptr;

          if (gendata.ports.contains(GenrouData<ScalarT, IdxT>::Ports::bus))
          {
            bus = getBus(gendata.ports.at(GenrouData<ScalarT, IdxT>::Ports::bus));
          }

          if (gendata.ports.contains(GenrouData<ScalarT, IdxT>::Ports::pmech_signal))
          {
            pmech = getBus(gendata.ports.at(GenrouData<ScalarT, IdxT>::Ports::pmech_signal));
          }

          if (gendata.ports.contains(GenrouData<ScalarT, IdxT>::Ports::speed_signal))
          {
            speed = getBus(gendata.ports.at(GenrouData<ScalarT, IdxT>::Ports::speed_signal));
          }

          auto* gen = new Genrou<ScalarT, IdxT>(bus, pmech, speed, gendata);
          addComponent(gen);
```
you can make this much cleaner

src/Model/PhasorDynamics/SystemModel.hpp
```diff
@@ -144,6 +170,19 @@ namespace GridKit
...
 
+        // Add governors
+        for (const auto& govdata : data.gov)
+        {
+
+          /* Need to add signals to gendata class */
+          auto* gov = new Governor::Tgov1<ScalarT, IdxT>(
+              getBus(govdata.signal_pmech),
+              getBus(govdata.signal_speed),
+              govdata);
```
superwhiskers, line 181: is it the case with `Tgov1` that these signals are non-optional? (that sounds correct to me but i want to make sure as it is not documented in the input format)

superwhiskers (changes requested):
some minor suggestions. i don't think i have any major comments.

src/Model/PhasorDynamics/Bus/BusSignal.hpp
```diff
@@ -0,0 +1,127 @@
...
+      /**
+       * @brief A write operation only available
+       * to the source signal
+       *
+       * @param value The value of the signal
+       */
+      void write(ScalarT& value) override
+      {
+        f_[0] += value;
+      }
```
pelesh, line 80: I believe this operation should never be allowed. It is _only the solver_ who should write the new values to the system variables. This operation may alter solution values and consequently system residual and Jacobian evaluation.

A better implementation would be for the source component to have a reference to the signal value and to use it in equations directly. A potential issue with signal values being stored in signal buses is that signal connections are decided at runtime. This means that generator, for example, would have to be able to configure its equations at runtime to use values from signal buses or its local variables based on user-chosen connectivity. This may also complicate computing derivatives of component equations at _compile time_.

pelesh (changes requested):

pelesh:
Closing this for now, but we can revisit this implementation approach later.

PR #175, opened 2025-07-08
fix the means by which cmake finds the suitesparse include directory

superwhiskers:
this is just a really minor fix so that cmake finds the include directory for suitesparse (on my system, at least)

 ## Checklist
 
- [x] All tests pass.

### DIFF
--- a/cmake/FindSuiteSparse.cmake
+++ b/cmake/FindSuiteSparse.cmake
@@ -1,4 +1,4 @@
-# 
+#
 #[[
 
 Finds Sutiesparse include directory and libraries and exports target `Suitesparse`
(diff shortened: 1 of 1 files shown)


superwhiskers:
the test failure does not appear to be related to the change this pull request is for

nkoukpaizan:
> the test failure does not appear to be related to the change this pull request is for

There's a non-deterministic behavior somewhere. I just re-ran the pipeline and it passed. The wrong answer we got when it failed was kind of scary.

superwhiskers:
i saw it fail in #169 as well. same test if i recall correctly, `PhasorDynamicsGenClassicalTest`

pelesh (approved):

PR #178, opened 2025-07-10
address #177

### LINKED ISSUE #177: Do not add magic enum and json parser subdirectories
In current implementation all header magic enumeration and JSON parser libraries are added as project subdirectories to GridKit:
```cmake
add_subdirectory(third-party/nlohmann-json)
add_subdirectory(third-party/magic-enum)
```
Instead, use CMake command `TARGET_INCLUDE_DIRECTORIES` to only include headers from these libraries when building GridKit targets that need those headers.

Other functionality from the two third-party libraries should not be built with GridKit to minimize dependecies on them.

superwhiskers:
closes #177 
 
 ## Checklist
 
- [x] All tests pass.


CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ set(PACKAGE_VERSION_MAJOR "0")
 set(PACKAGE_VERSION_MINOR "1")
 set(PACKAGE_VERSION_PATCH "0")
 
+set(THIRD_PARTY_DIR "${PROJECT_SOURCE_DIR}/third-party")
+
```
pelesh: I believe it should be without quotes:
```suggestion
set(THIRD_PARTY_DIR ${PROJECT_SOURCE_DIR}/third-party)

```
My understanding is with quotes CMake interprets it as a string, without it is interpreted as a path.

tests/UnitTests/Utilities/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_executable(test_case_format runCaseFormatTests.cpp)
-target_link_libraries(test_case_format PRIVATE nlohmann_json::nlohmann_json magic_enum::magic_enum)
+target_include_directories(test_case_format PRIVATE "${THIRD_PARTY_DIR}/nlohmann-json/single_include" "${THIRD_PARTY_DIR}/magic-enum/include")
```
pelesh: Same here:
```suggestion
target_include_directories(test_case_format PRIVATE ${THIRD_PARTY_DIR}/nlohmann-json/single_include ${THIRD_PARTY_DIR}/magic-enum/include)
```

pelesh (approved):
Good to merge. I would suggest dropping quotes around paths, though.

PR #181, opened 2025-07-14
Introduce signal node object to abstract connections with control devices

### LINKED ISSUE #154: Add signal bus model to phasor dynamics family

pelesh:
## Description
 
This is a draft for implementation of signal node (bus) object.
 
 Closes #154 
 
 @superwhiskers 
 
 ## Proposed changes
 
 In this implementation signal node does not own any variable or residual. The `SignalNode` class implements methods to set, initialize and read signal:
1. `void set(ScalarT* p_signal)` method stores the pointer to a component model variable that will be shared with other components as a (control) signal. Only component that owns that variable can invoke the set function. 
2. `void init(ScalarT signal)` writes value of `signal` to the memory location pointed by the pointer stored in the `SignalNode` instance. Both, the component sending and the component receiving signal can initialize the signal.
3. `ScalarT read()` is a const method that reads the value pointed by the pointer stored in the `SignalNode` instance.

Note: When initializing the system, generators need to be initialized before governors because mechanical power sent to a generator by the governor is initialized by the generator rather than the governor.
 
 ## Checklist
  
- [x] All tests pass.
 
 ## Further comments
 
- Current implementation keeps the old interface where generator has pointer directly to the governor and vice versa.
- Still TODO: Create method to select which variable to connect to the signal bus.


pelesh:
Some additional comments on ideas behind the proposal in this PR:
- Components that are source of signals own signal variables. I think this is helpful because:
    - The number of component internal variables is always the same no matter what is the connectivity.
    - Internal variables including signal variables can be stored in a contiguous array to simplify implementing compile-time automatic differentiation.
- Signal node stores address and/or index of the variable shared as a signal. The signal node has minimal functionality and its purpose is to abstract away the source of signal from component receiving the signal.
- Components that receive signal get the address/index of the signal variable so they can use it directly in its residual/Jacobian equations or make a local copy of the signal variable.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/Genrou.hpp
```diff
@@ -47,9 +50,11 @@ namespace GridKit
       using gov_type        = GovernorBase<ScalarT, IdxT>;
```
lukelowry, line 50: The GovernorBase class can be removed. If it is is outside scope I would at least make an Issue as a reminder to remove this class.

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
```diff
@@ -111,6 +114,74 @@ namespace GridKit
...
+        gendata.parameters[GenrouParameters::Xqp]   = 0.5;
+        gendata.parameters[GenrouParameters::Xqpp]  = 0.18;
+        gendata.parameters[GenrouParameters::Xl]    = 0.15;
+        gendata.parameters[GenrouParameters::S10]   = 0.;
+        gendata.parameters[GenrouParameters::S12]   = 0.;
+
+        PhasorDynamics::Bus<ScalarT, IdxT>             bus(busdata);
+        PhasorDynamics::SignalNode<ScalarT, IdxT>      pmech;
+        PhasorDynamics::SignalNode<ScalarT, IdxT>      omega;
+        PhasorDynamics::Genrou<ScalarT, IdxT>          gen(&bus, &omega, &pmech, gendata);
```
lukelowry, line 98: Either as an Issue or in this PR, the SystemModel needs to know how to pass these pointers based on the information in the Data structures. If outside scope of PR, we should atleast add some unqiue ID field and accessor function that can be used to identify each signal. @superwhiskers

src/Model/PhasorDynamics/SignalNode/SignalNode.hpp
```diff
@@ -0,0 +1,304 @@
...
+namespace GridKit
+{
+  namespace PhasorDynamics
+  {
+    /*!
+     * @brief SignalNode model implementation base class.
+     *
+     */
+    template <class ScalarT, typename IdxT>
+    class SignalNode : public Model::Evaluator<ScalarT, IdxT>
```
lukelowry, line 26: I don't see the advantage of making this inherit from Evaluator if it does not functionally implement any of the virtual functions. It doesn't hurt, I suppose.

lukelowry (commented):
Overall I prefer this approach to #173 , but we still need to make the signals definable via data structure soon. Up to @pelesh if we do that in this PR or as next task.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/Genrou.hpp
pelesh: I am keeping governor base class for testing purposes to make sure I am getting the same results with old and new approach. Will remove it once we have sufficiently tested signal buses.

src/Model/PhasorDynamics/SignalNode/SignalNode.hpp
pelesh: Frankly, neither do I. I was keeping it as a part of the draft. We can make signal node a lightweight utility class.

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
pelesh: The idea here is to remove pointers to all buses from the constructors. The component data already contains that information. We have to figure out how to make that transition with minimal interruptions.

lukelowry: Perhaps a generalized signal setter function would be useful addition to the Component class, or if preferred model specific setters (i.e. setPmechSignal, setOmegaSignal). I find the idea of a generalized Component signal setter more appealing, so each model has a vector of indexed signals, rather than named signals. We can even make the ordering of the signals align with the ordering of the External variables in the documentation.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/Genrou.hpp
lukelowry: I'm okay with retaining it for now

src/Model/PhasorDynamics/SignalNode/SignalNode.hpp
nkoukpaizan: I agree with making it lightweight utility class. If I understand this approach correctly, the signal node  will be a "software artifact" to communicate between components. Unlike the electrical bus that own the voltages which are coupling variables and need to be dealt with in the system Jacobian, this signal node does not own any variables. So, for example, there's no reason for it to have evaluateResidual/evaluateJacobian/getJacobian methods.

pelesh: I wouldn't say that signal node is a software artifact, it represents a signal coming from a sensor or an actuator output. But it is true, its software implementation can be very simple.

What I am proposing here is to start with a minimalistic approach, which is an incremental step over what @lukelowry implemented earlier for governor and GENROU models, and update it if needed later. I also don't see a reason (anymore) for the signal bus to have its own residual/Jacobian functions.

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
pelesh: Implementing a generalized signal setter is now part of the issue #189.

src/Model/PhasorDynamics/SignalNode/SignalNode.hpp
pelesh: I suggest we keep inheritance from `Evaluator` class in this PR. We can remove it later, if needed.

nkoukpaizan (approved):

lukelowry (approved):
All looks well. I prefer this over the initial PR for SignalBus #173 

I believe merging this will allow me to submit PR for the exciter model in my feature branch. Hope we can get this approved asap!

PR #184, opened 2025-07-15
Use interface target to link to component models

pelesh:
## Description
 
GridKit requires system model to be linked with each component model library separately. For convenience, I suggest using interface target for phasor dynamics components to which all component models will be linked.

Each GridKit app involving system model would only need to link to that interface as
```cmake
target_link_libraries(myApp GRIDKIT::phasor_dynamics_components)
```

@abirchfield 
  
 ## Proposed changes
 
The CMake configuration file in `src/Model/PhasorDynamics` directory should create the interface target `GRIDKIT::phasor_dynamics_components`. CMake configuration for each component should link the component model library to that target like this:
```cmake
# Link to interface target for all components
target_link_libraries(gridkit_phasor_dynamics_components INTERFACE GRIDKIT::phasor_dynamics_branch)
```
That way, when building new examples or GridKit apps, one needs to link only to the interface library instead of to individual component models.

This PR also adds a common include file for all phasor dynamics component models to simplify include statements when developing GridKit apps.

 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments


nkoukpaizan (approved):
Looks good to me. Tests pass on Frontier. It would be good to push again to check that CI passes as well.

PR #186, opened 2025-07-16
Default is to build GridKit without dependencies

pelesh:
## Description
 
Default GridKit build was with SUNDIALS as dependency. This PR changes that to a build with no dependencies. This can help, for example, work on component models and unit tests, which do not require a solver enabled. 

 ## Proposed changes
 
Made changes to CMake so that GridKit modules and examples are built only if their required dependencies are enabled.
 
 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments


tests/UnitTests/Utilities/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_executable(test_case_format runCaseFormatTests.cpp)
-target_include_directories(test_case_format PRIVATE ${THIRD_PARTY_DIR}/nlohmann-json/single_include ${THIRD_PARTY_DIR}/magic-enum/include)
+target_include_directories(test_case_format PRIVATE ${GRIDKIT_THIRD_PARTY_DIR}/nlohmann-json/single_include ${GRIDKIT_THIRD_PARTY_DIR}/magic-enum/include)
```
lukelowry, line 2: I had to comment out this file for my machine to build. No issues with the json include, but compiler was unable to locate <magic_enum/magic_enum.hpp>. Could be something on my end, but unclear how to fix.

lukelowry (approved):
I agree with need of this PR. One minor issue, details below. Otherwise, builds correctly and all tests passed.

tests/UnitTests/Utilities/CMakeLists.txt
nkoukpaizan: Can you try `git submodule update --init`?

nkoukpaizan:
It would be good to add [GRIDKIT_ENABLE_ENZYME/GRIDKIT_USE_ENZYME](https://github.com/ORNL/GridKit/blob/develop/CMakeLists.txt#L109) to `Definitions.hpp` here. 

And I understand why the other options are there (e.g., `GRIDKIT_USE_RAJA`), but my preference would be to only list options we currently support.

pelesh:
> It would be good to add [GRIDKIT_ENABLE_ENZYME/GRIDKIT_USE_ENZYME](https://github.com/ORNL/GridKit/blob/develop/CMakeLists.txt#L109) to `Definitions.hpp` here.
> 
> And I understand why the other options are there (e.g., `GRIDKIT_USE_RAJA`), but my preference would be to only list options we currently support.

Addressed in [e0e25a7](https://github.com/ORNL/GridKit/pull/186/commits/e0e25a703ad8707690612d66da1ac3d2f3760a27).

nkoukpaizan (approved):

tests/UnitTests/Utilities/CMakeLists.txt
lukelowry: Forgot to thank you for this fix. This was the issue. Is it possible to add this advice to the main README or perhaps a debug page for GridKit? @nkoukpaizan @pelesh

PR #193, opened 2025-07-22
initial attempt at a unified interface to signal buses

### LINKED ISSUE #190: Implement a generalized signal setter function
A generalized signal setter function would be useful addition to the Component class, or if preferred model specific setters (i.e. setPmechSignal, setOmegaSignal). I find the idea of a generalized Component signal setter more appealing, so each model has a vector of indexed signals, rather than named signals. We can even make the ordering of the signals align with the ordering of the External variables in the documentation.

_Originally posted by @lukelowry in https://github.com/ORNL/GridKit/pull/181#discussion_r2208320617_

superwhiskers:
an attempt at providing an interface to signal buses on components

closes #190 

# checklist
 
- [x] All tests pass.


shakedregev (commented):
This seems easier from the user perspective compared to #199. Let's make sure this runs smoothly.

superwhiskers:
this is probably ready for review now. i don't think changing constructors is appropriate just yet. doing that will be more appropriate when they way they're handled in the json format / data model is worked out. all this pull request does is unify the way components handle them internally, in preparation for a more involved change.

pelesh:
> this is probably ready for review now. i don't think changing constructors is appropriate just yet. doing that will be more appropriate when they way they're handled in the json format / data model is worked out. all this pull request does is unify the way components handle them internally, in preparation for a more involved change.

You just need to modify genrou and tgov1 data structures. We can assemble them manually, without a parser in an example. With Genrou and Tgov1 constructors taking in the data structures and initializing signal connections, this PR will be ready to merge. 

Other than that, this is very nice work and the code is clean and readable.

src/Model/PhasorDynamics/Governor/Tgov1/Tgov1.cpp
```diff
@@ -25,21 +24,32 @@ namespace GridKit
...
+
+        // 3 internal variables
+        size_ = 3;
+      }
+
       /**
-       * @brief Constructs a Tgov1 governor model using signal inputs directly.
-       *
-       * Initializes the model parameters and sets the internal model size.
+       * @brief Constructs a Tgov1 governor model from its parameters
```
WiktoriaZielinskaORNL, line 50: You have this same line on line 27. Maybe remove this one if it was accidentally left over.

superwhiskers: these are for two different overloads of the method

pelesh (approved):
lgtm

PR #199, opened 2025-07-24
different implementation of signal assignment

### LINKED ISSUE #190: Implement a generalized signal setter function

superwhiskers:
another attempt at implementing signal assignment on components

closes #190 

 ## checklist
 
- [x] All tests pass.

### DIFF
--- a/src/Model/PhasorDynamics/Branch/Branch.cpp
+++ b/src/Model/PhasorDynamics/Branch/Branch.cpp
@@ -7,11 +7,11 @@
  *
  */
 
-#include "Branch.hpp"
-
+#include <cassert>
 #include <cmath>
 #include <iostream>
 
+#include "Branch.hpp"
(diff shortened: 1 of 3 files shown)


shakedregev (commented):
This code requires the user to memorize relations between enums and integers. There are multiple problems with type safety. I recommend improving #193 instead.

pelesh:
Branch is not the best example to work this out because a branch does not have signal connections. I would use generator and governor as my two test cases.

superwhiskers:
closing in favor of #193

PR #207, opened 2025-08-10
Input Format README Updates

lukelowry:
## Description

Various changes to the INPUT_FORMAT document so that it aligns with the documentation and implementation of models. The README files and small Syntax corrections were made in various models.
 
 ## Proposed changes
 
The `signal_in` and `signal_out` terminology should not be used because a model can/will have multiple input and outputs. Instead the proposed format uses `pmech` and `speed` and `efd` directly named.

All documentation and implementation now uses $\omega$ for speed deviation.
All documentation now  uses $P_m$ for mechanical power.
 

 ## Further comments
 
Draft. Some implementation variables still need to be updated.

### DIFF
--- a/src/Model/PhasorDynamics/Exciter/IEEET1/README.md
+++ b/src/Model/PhasorDynamics/Exciter/IEEET1/README.md
@@ -93,7 +93,7 @@ $k_{sat}$       | [p.u.] | Saturation variable               |
 #### Differential
 Symbol          | Units  | Description                       | Note
 ----------------|--------|-----------------------------------|-------
-$\Delta\omega$  | [p.u.] | Speed Deviation                   | Read from a Machine Model
+$\omega$  | [p.u.] | Speed Deviation                   | Read from a Machine Model
 
 #### Algebraic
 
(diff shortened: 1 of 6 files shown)


abirchfield (commented):
Looks great. Two suggestions.

(1) Move the signals table ahead of the devices table. I think conceptually that the JSON object members are unordered, but of course in a text file they are inherently ordered. Signals are logically prior to devices--that is, devices need to know about the existence of signals but not vice versa. This will help with human readability as well as probably simplifying the parsing.

(2) Below the devices table, add a sentence something like: "See the detailed documentation for each device class for its equations, and for the specification of ports as electrical buses, signal inlets, and signal outlets."

abirchfield (approved):

pelesh (approved):
Looks good.

Question: Why is GridKit spack package updated?

pelesh:
CC @nkoukpaizan

abirchfield:
> Looks good.
> 
> Question: Why is GridKit spack package updated?

I assume it is not intentional. Aren't these submodules separate from our main code?

nkoukpaizan:
Looks unintentional to me as well. Something like a `git add` on a submodule folder that was outdated. The hash that is used in this PR points to an older commit than what we have in develop. I *think* this will need to be undone manually. @lukelowry let me know if you need some guidance on fixing it.

nkoukpaizan:
> Looks unintentional to me as well. Something like a `git add` on a submodule folder that was outdated. The hash that is used in this PR points to an older commit than what we have in develop. I _think_ this will need to be undone manually. @lukelowry let me know if you need some guidance on fixing it.

I went ahead and made the change. Good to merge.

lukelowry:
Yes, the spack changes were unintentional. Thank you for making the changes.

PR #208, opened 2025-08-11
Have Doxygen document private members

alexander-novo:
## Description
 
Modify the `Doxyfile` to make Doxygen document private members. Since the Doxygen documentation is developer-focused, it would be good to have documentation for these members.
 
Also moved the doxygen output folder to be in-tree: the current setup escapes the project and assumes the user has access to their parent folder (and wants to write there).

 
 ## Checklist
 
 _Put an `x` in the boxes that apply. You can also fill these out after creating
 the PR. If you're unsure about any of them, don't hesitate to ask. We're here
 to help! This is simply a reminder of what we are going to look for before
 merging your code._
 
- [x] All tests pass.


Doxyfile
```diff
@@ -56,7 +56,7 @@ PROJECT_LOGO           =
 # entered, it will be relative to the location where doxygen was started. If
 # left blank the current directory will be used.
 
-OUTPUT_DIRECTORY       = ../doc
+OUTPUT_DIRECTORY       = ./doxygen-docs
```
pelesh, line 60: Default location should encourage user/developer to generate docs outside the source tree.
```suggestion
OUTPUT_DIRECTORY       = ../doxygen-docs

```

pelesh (approved):
Okay to merge. I suggest fixing one minor comment.

Doxyfile
alexander-novo: Consider someone using, for instance, the dev container. VS Code will mount the project folder into the container, so that everything inside the folder is shared between host and guest machines. Having documentation out of tree means that it won't be shared between machines, it will take up extra space on the container, and there's no guarantee the user has access to the parent directory in the first place (meaning they won't even be able to generate the docs).

I've added the folder to the `.gitignore` so it won't be committed. Is there another reason to not have docs inside the source folder? This is consistent with documentation tools in other languages like `rustdoc`.

PR #209, opened 2025-08-11
Implement CSR Matrix format for Jacobian computations

alexander-novo:
## Description
 
Currently, Jacobian computations are done purely with the `COO_Matrix` class, which is eventually converted to a CSR format expected by SUNDIALS.
 
 ## Proposed changes
 
We should be able to continue to use the `COO_Matrix` class to calculate sparsity patterns, then convert it to a "template" CSR matrix, which pre-allocates and pre-fills all of the necessary indices, allowing speedup of Jacobian computations.
 
 ## Checklist
 
_CURRENTLY WIP_
 
- [ ] All tests pass.


alexander-novo:
Closing because I don't need review or CI yet

PR #211, opened 2025-08-14
Separate Impl and template instantiation for Branch and Exciter

nkoukpaizan:
## Description
 
This separates the model implementation from template instantiation for a few models. This will be useful when adding the instantiation with Enzyme and testing the Jacobians.
 
CC @lukelowry @abirchfield It would be good to use this implementation approach (separate *Impl.hpp and *cpp) when adding new models. 
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 N/A


pelesh (approved):
Changes in line with the rest of the code. All tests pass on multiple platforms.

PR #215, opened 2025-08-20
Update cmake to indicate which libraries are optional, and not fail when they cannot be configured

### LINKED ISSUE #212: Build issue: SUNDIALS without KLU
### Build error message(s)

```console
 CMake Error at cmake/GridkitAddLibrary.cmake:57 (target_link_libraries):
   Target "solvers_dyn_obj_shared" links to:

     SUNDIALS::sunlinsolklu

     * An ALIAS target is missing.

 Call Stack (most recent call first):
   src/Solver/Dynamic/CMakeLists.txt:8 (gridkit_add_library)

 CMake Error at cmake/GridkitAddLibrary.cmake:69 (target_link_libraries):
   Target "solvers_dyn_shared" links to:

     SUNDIALS::sunlinsolklu

     * An ALIAS target is missing.

 Call Stack (most recent call first):
   src/Solver/Dynamic/CMakeLists.txt:8 (gridkit_add_library)

 Generating done
```

### GridKit™ version

0.1.0

When building with `GRIDKIT_ENABLE_SUNDIALS=ON` and `GRIDKIT_ENABLE_SUNDIALS_SPARSE=OFF`.

### Additional information

We'll need to either require that SUNDIALS is built with KLU or add more checks in the CMake and cpp files to handle the case when it isn't available.

alexander-novo:
## Description
 
Fixes #212.

There's currently an issue with how optional library components and executables are setup in the cmake. These libraries and executables are guarded by if statements checking if their optional components are installed before including their folders. If a library changes which optional components it needs, then all of the if statements guarding its folder and all derivative executables need to change. This is also fragile, since it requires all optional components to be identified ahead of time (see #212 where it was assumed `sunlinsolklu` wasn't optional).
 
 ## Proposed changes
 
Now, the `gridkit_add_library` macro allows an `OPTIONAL` parameter and will check the existence of all linked libraries. If the linked libraries exist - nothing different happens, but if a library which must be linked doesn't exist and the library is marked as `OPTIONAL`, then that target will simply be ignored.

There is a new `gridkit_add_executable` macro as well which does a similar process - if an executable depends on a disabled optional library, then the executable target will just become ignored.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
Only the `PowerElectronics` examples have been configured using the new `gridkit_add_executable` macro, since these were the ones giving me trouble - due to a cmake file looking for the `KLU` target instead of `sunlinsolklu` (which was actually what was needed).

### DIFF
--- a/cmake/GridkitAddExecutable.cmake
+++ b/cmake/GridkitAddExecutable.cmake
@@ -1,0 +1,45 @@
+# add_executable macro loosely based on add_library
+
+macro(gridkit_add_executable target)
+
+  set(options TEST)
+  set(multiValueArgs SOURCES LINK_LIBRARIES COMPILE_OPTIONS)
+
+  # parse arguments
+  cmake_parse_arguments(gridkit_add_executable
+    "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  # Before going any further and configuring this library - make sure we can actually link its requirements
+  foreach(_add_lib ${gridkit_add_executable_LINK_LIBRARIES})
+    if(NOT (${_add_lib} MATCHES "PUBLIC" OR ${_add_lib} MATCHES "PRIVATE" OR ${_add_lib} MATCHES "INTERFACE"))
+      if(NOT TARGET ${_add_lib})
+        get_property(_skipped_libs GLOBAL PROPERTY GRIDKIT_SKIPPED_OPTIONAL_LIBS)
+
+        if(${_add_lib} IN_LIST _skipped_libs)
+          message(STATUS "Disabling GridKit executable ${target} due to disabled optional library: ${_add_lib}")
+          return()
+        else()
+          message(SEND_ERROR "Failed to configure GridKit executable due to missing library: ${_add_lib}")
+        endif()
+      endif()
+    endif()
+  endforeach()
+
+  # -- Create executable --
+
+  add_executable(${target} ${gridkit_add_executable_SOURCES})
+
+  if(gridkit_add_executable_COMPILE_OPTIONS)
+    target_compile_options(${target} PUBLIC ${gridkit_add_executable_COMPILE_OPTIONS})
+  endif()
+  
+  if(gridkit_add_executable_LINK_LIBRARIES)
+    target_link_libraries(${target} ${gridkit_add_executable_LINK_LIBRARIES})
+  endif()
+
+  if(gridkit_add_executable_TEST)
+    add_test(NAME ${target} COMMAND $<TARGET_FILE:${target}>)
+  endif()
+
+  install(TARGETS ${target} RUNTIME DESTINATION bin)
+endmacro()
--- a/cmake/GridkitAddExecutable.cmake
+++ b/cmake/GridkitAddExecutable.cmake
@@ -9,7 +9,7 @@ macro(gridkit_add_executable target)
   cmake_parse_arguments(gridkit_add_executable
     "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
-  # Before going any further and configuring this library - make sure we can actually link its requirements
+  # Before going any further and configuring this library, make sure we can actually link its requirements.
   foreach(_add_lib ${gridkit_add_executable_LINK_LIBRARIES})
     if(NOT (${_add_lib} MATCHES "PUBLIC" OR ${_add_lib} MATCHES "PRIVATE" OR ${_add_lib} MATCHES "INTERFACE"))
       if(NOT TARGET ${_add_lib})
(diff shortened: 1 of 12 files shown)


nkoukpaizan:
Tests pass in CI and on Frontier. I need to take a closer look for edge cases. 

This does not entirely fix #212. It partially addresses the CMake component, but the other part of the issue is that [Ida.cpp](https://github.com/ORNL/GridKit/blob/develop/src/Solver/Dynamic/Ida.cpp) has code for both the dense and the sparse interface. So, it breaks when SUNDIALS is not built with KLU. Marking it as optional altogether is not enough, because there are other dependencies that would also have to be marked optional. I'd say this is out of scope for this PR.
``` 
     22    -- Disabling optional GridKit library solvers_dyn due to missing library: SUNDIALS::sunlinsolklu
  >> 23    CMake Error at cmake/GridkitAddLibrary.cmake:28 (message):
     24      Failed to configure required GridKit library solvers_opt due to missing
     25      library: GRIDKIT::solvers_dyn
     26    Call Stack (most recent call first):
     27      src/Solver/Optimization/CMakeLists.txt:8 (gridkit_add_library)
```

alexander-novo:
That's true, there is a bunch of code in `Ida.cpp` that needs to be revamped. There is also a dependence on IDAS that could potentially be removed as well.

That issue should probably be renamed, because right now it's labeled as a "build issue" and as of right now it's impossible to build `Ida.cpp` without `sunlinsolklu` and without changing the code of `Ida.cpp`. This PR at least fixes cmake attempting to build `Ida.cpp` without `sunlinsolklu`

alexander-novo:
I also wanted to add an example of what it looks like when an optional library is missing its dependency:

```
root@907866574cc8:/workspaces/GridKit/build# cmake .. -DGRIDKIT_ENABLE_SUNDIALS=OFF
-- Disabling optional GridKit library solvers_dyn due to missing library: SUNDIALS::nvecserial
-- Disabling GridKit executable rlcircuit due to disabled optional library: GRIDKIT::solvers_dyn
-- Disabling GridKit executable microgrid due to disabled optional library: GRIDKIT::solvers_dyn
-- Disabling GridKit executable scalemicrogrid due to disabled optional library: GRIDKIT::solvers_dyn
-- Configuring done (1.1s)
```

Steven-Roberts:
> That's true, there is a bunch of code in `Ida.cpp` that needs to be revamped.

I'm curious what you have in mind for this. Perhaps you could make an issue out of this.

cmake/GridkitAddExecutable.cmake
```diff
@@ -0,0 +1,45 @@
...
+macro(gridkit_add_executable target)
+
+  set(options TEST)
+  set(multiValueArgs SOURCES LINK_LIBRARIES COMPILE_OPTIONS)
+
+  # parse arguments
+  cmake_parse_arguments(gridkit_add_executable
+    "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  # Before going any further and configuring this library - make sure we can actually link its requirements
```
shakedregev:
```suggestion
  # Before going any further and configuring this library, make sure we can actually link its requirements.
```

cmake/GridkitAddLibrary.cmake
```diff
@@ -8,14 +8,29 @@
...
-  set(options STATIC_ONLY SHARED_ONLY)
+  set(options STATIC_ONLY SHARED_ONLY OPTIONAL)
   set(oneValueArgs OUTPUT_NAME)
   set(multiValueArgs SOURCES LINK_LIBRARIES INCLUDE_DIRECTORIES COMPILE_OPTIONS)
 
   # parse arguments
   cmake_parse_arguments(gridkit_add_library
     "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
+  # Before going any further and configuring this library - make sure we can actually link its requirements
```
shakedregev:
```suggestion
  # Before going any further and configuring this library, make sure we can actually link its requirements.
```

shakedregev (commented):
Still need to test. Minor comments.

shakedregev (approved):
Tests pass.

alexander-novo:
@pelesh mentioned on slack:

> All external dependencies are optional.

So it might be good to go through after this PR and update all GridKit libraries which depend on external dependencies as optional.

pelesh (changes requested):
The scope of this PR is not quite well defined. Having a macro `gridkit_add_executable` is certainly helpful, but how does it fix #212? I suggest breaking this PR into specific focused PRs addressing well-defined issues.

pelesh:
Closing in favor of #218. I suggest we document and ideas from this PR is corresponding issues and submit more focused PRs once we decide how to proceed.

PR #216, opened 2025-08-20
Update Dev Container Dockerfile

alexander-novo:
- Use correct numbered version of SUNDIALS (previous commit we were using no longer exists)
- Add GDB - a likely necessity when developing GridKit
- Add ssh for git
 
 ## Checklist

**N/A** 

- [ ] All tests pass.


nkoukpaizan (approved):
Looks good to me!

PR #218, opened 2025-08-22
Fix to enable Ida build when SUNDIALS is not configured with KLU

### LINKED ISSUE #212: Build issue: SUNDIALS without KLU

nkoukpaizan:
## Description

Closes #212
 
 ## Proposed changes
 
- Added `configureLinearSolverSparse` and `configureLinearSolverDense` member functions that are called from `configureLinearSolver`. The sparse version is guarded by preprocessor directives.
- Added a check at the CMake level before attempting to link `SUNDIALS::sunlinsolklu`. Once #215 is rescoped, automatically checking for optional external dependencies should simplify this.
- Added CI tests for `~klu~sundials` and `~klu+sundials`.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 I am open to suggestions on a cleaner approach in `Ida.cpp` that relies less on the preprocessor directives (but also does not require too many changes for now).


src/Solver/Dynamic/Ida.cpp
```diff
@@ -110,48 +116,113 @@ namespace AnalysisManager
...
+      {
+        this->configureLinearSolverDense();
+      }
+#else
       if (model_->hasJacobian())
       {
-        sunindextype n   = static_cast<sunindextype>(model_->size());
-        sunindextype nnz = static_cast<sunindextype>((model_->getJacobian()).nnz());
+        /// Todo - Improve error handling capabilities and hasJacobian_ ownership
+        throw std::runtime_error("SUNDIALS is not configured with KLU, but the model has a (sparse) Jacobian.");
```
Steven-Roberts, line 147: Instead of throwing an error, could we just fall back on the dense linear solver?

src/Solver/Dynamic/Ida.cpp
```diff
@@ -110,48 +116,113 @@ namespace AnalysisManager
...
+#ifdef GRIDKIT_ENABLE_SUNDIALS_SPARSE
+      if (model_->hasJacobian())
+      {
+        this->configureLinearSolverSparse();
+      }
+      else
+      {
+        this->configureLinearSolverDense();
+      }
+#else
```
Steven-Roberts, line 143: These ifs and `#ifdef` can be simplified slightly by moving the `#ifdef` inside the if block so the else case doesn't have to be duplicated.

Steven-Roberts (commented):
I can confirm this builds correctly without KLU. Interestingly the value of `GRIDKIT_ENABLE_SUNDIALS_SPARSE` is always `ON` and cannot be set to `OFF` when SUNDIALS has KLU enabled. I'm not sure if that's really so important, though.

nkoukpaizan:
> I can confirm this builds correctly without KLU. Interestingly the value of `GRIDKIT_ENABLE_SUNDIALS_SPARSE` is always `ON` and cannot be set to `OFF` when SUNDIALS has KLU enabled. I'm not sure if that's really so important, though.

I think that's the intended behavior. We only used KLU from SUNDIALS at the moment, and no longer intend for the user to set it. Not exposing it to the user should clarify that.

src/Solver/Dynamic/Ida.cpp
nkoukpaizan: So, there are two things at play here: the _compile-time_ availability of KLU from SUNDIALS and the _runtime_ selection of a user-defined Jacobian (which we currently assume to be sparse). This goes to the suggestion to consolidate the `if` and `ifdef` as well. We want to catch when a user is trying to use a model that has a sparse Jacobian without KLU available. 

This will be a different story when we change the ownership of the `hasJacobian` flag from the models to the dynamic solver. See https://github.com/ORNL/GridKit/blob/develop/src/Model/PhasorDynamics/Component.hpp#L36. In that case, your suggestion would be exactly what we need.

Steven-Roberts (approved):

nkoukpaizan:
Addressed hiding `GRIDKIT_ENABLE_SUNDIALS_SPARSE` from users in [3dcce80](https://github.com/ORNL/GridKit/pull/218/commits/3dcce80c2b099eccc8558a5defb58428024b602b).

src/Solver/Dynamic/Ida.cpp
pelesh: I agree! It is a configuration option of a solver to choose whether estimate Jacobian numerically or use user-supplied Jacobian.

src/Solver/Dynamic/CMakeLists.txt
```diff
@@ -5,12 +5,23 @@
...
+else()
+  gridkit_add_library(solvers_dyn
+    SOURCES
+      Ida.cpp
+    LINK_LIBRARIES
+      PUBLIC SUNDIALS::nvecserial
+      PUBLIC SUNDIALS::idas
+    OUTPUT_NAME
+      gridkit_solvers_dyn)
+endif()
```
pelesh, line 27: This makes me wonder if `gridkit_add_library` macro really helps simplify the cmake code. In this case you have to replicate the entire call to the macro. One cannot do
```cmake
target_link_library(solvers_dyn PUBLIC SUNDIALS::sunlinsolklu)
```
after the call to the macro without linking to `sunlinsolklu`.

pelesh:
> I can confirm this builds correctly without KLU. Interestingly the value of `GRIDKIT_ENABLE_SUNDIALS_SPARSE` is always `ON` and cannot be set to `OFF` when SUNDIALS has KLU enabled. I'm not sure if that's really so important, though.

I believe this is desired behavior. This variable is an output variable and should be hidden from the user outside of the advanced mode.

CMakeLists.txt
```diff
@@ -98,10 +95,10 @@ if(GRIDKIT_ENABLE_SUNDIALS)
...
   message(STATUS "SUNDIALS configuration found: ${SUNDIALS_CONFIG}")
-endif()
-if(TARGET SUNDIALS::KLU)
-  set(GRIDKIT_ENABLE_SUNDIALS_SPARSE ON CACHE BOOL "Sparse solver enabled in SUNDIALS." FORCE)
-  message(STATUS "KLU support enabled in SUNDIALS.")
+  if(TARGET SUNDIALS::sunlinsolklu)
+    set(GRIDKIT_ENABLE_SUNDIALS_SPARSE ON CACHE INTERNAL BOOL "Sparse solver enabled in SUNDIALS.")
+    message(STATUS "KLU support enabled in SUNDIALS.")
+  endif()
 endif()
```
pelesh, line 102:
```suggestion
  if(TARGET SUNDIALS::sunlinsolklu)
    set(GRIDKIT_ENABLE_SUNDIALS_SPARSE ON CACHE INTERNAL BOOL "Sparse solver enabled in SUNDIALS.")
    message(STATUS "KLU support enabled in SUNDIALS.")
  endif()
endif()
mark_as_advanced(GRIDKIT_ENABLE_SUNDIALS_SPARSE)
```

pelesh (approved):
It is functional and addresses #212. I don't particularly like use of preprocessor directives in the code, but that's a matter of taste.

I suggest hiding `GRIDKIT_ENABLE_SUNDIALS_SPARSE` variable in the default mode.

alexander-novo:
I already talked to @nkoukpaizan about this face to face, but I just want to write it here for others:

1) This addresses not being able to build without specifically `SUNDIALS::linsolklu` but all of `SUNDIALS` is considered an optional dependency right? If my build of SUNDIALS doesn't have `idas` or `nvecserial` then the build will still fail because this `if` statement only checks for `sunlinsolklu`. We can surround this library and all executables with an if statements checking `if(TARGET idas AND TARGET nvecserial)` sure, but what happens when the dependencies we need for `solvers_dyn` change? We'll have to hunt down every executable that's built by default and update all of their `if` statements. I guess that's fine, but it kind of seems to me like it would be nice to automate that process. As well one of the problems originally was that the `if` statement was checking the wrong target - it could easily happen again where dependencies are changed and an `if` statement is missed and builds start failing again.

2) I'm also not so a big a fan of the `#ifdef` approach. It requires the least modification of `Ida.cpp` right now but I think it leaves the most future work. What if we add another linear solver we'd like to work with `Ida`? Will we have a 4-way if statement in the cmake and in the code that handles every possibility of linear solver being available? Also as mentioned above, the linear solver used should be a user choice - I think this is why SUNDIALS has separate targets for `idas` and `sunlinsolklu` etc. It seems more forward thinking to have our `Ida.cpp` wrapper make the same choice as SUNDIALS' IDA - have it configurable to work with multiple linear solvers and have the library consumer choose which solver they want to use. That way we don't have to have any particular linear solver library available when trying to link `solvers_dyn` and then we only need to link the exact library the user uses, rather than potentially linking every linear solver library. (Maybe to preserve the current behaviour we could have a "sensible default" helper which picks a linear solver based on the available-to-be-linked ones and some properties of the system, such as whether or not there is a sparse analytic jacobian). (This is part of the changes I was alluding to in the other PR @Steven-Roberts )

EDIT: I forgot to add: this last point will become important when we want to compare linear solvers for the purpose of benchmarking - as it is right now it is quite difficult to re-use a system and simply switch the linear solver being used to solve the system.

nkoukpaizan:
@alexander-novo You're making some great points. Thanks! I've added GridKit issues #220 and #221 to discuss these further.

The idea is to make PRs that have a well-defined scope and address specific issues or enhancements. Since this is functional, it should help verify that improvements don't break anything.

CMakeLists.txt
nkoukpaizan: The `INTERNAL` option already hides the variable from the user, but I don't mind replacing it with `mark_as_advanced`.

pelesh: No need for as long as it is hidden.

nkoukpaizan:
All comments have been addressed. Thanks everyone. Merging.

PR #223, opened 2025-09-04
Remove third party libraries from doxygen documentation

alexander-novo:
## Description
 
Currently third party libraries which are included in-tree as submodules, such as `nlohmann/json` and `magic-enum`, are being documented by doxygen. The JSON library is an especially large library (much much more documentation than Gridkit), so Doxygen spends more time generating documentation for it than for Gridkit.
 
 ## Proposed changes
 
This PR removes all items in the `third-party/` folder from considerations when generating doxygen documentation
 
 ## Checklist
 
N/A
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -920,7 +920,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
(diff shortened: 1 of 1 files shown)


nkoukpaizan (approved):
The change only affects the documentation, and it makes sense to me.

PR #224, opened 2025-09-04
Add CSR Matrix class

alexander-novo:
## Description
 
Add `CSRMatrix` class. Along with the `ScalarT` and `IdxT` template parameters, it includes a `KEEP_SORTED` parameter. I talked with @shakedregev about KLU and found that it doesn't require the CSR matrices it consumes to be sorted (have elements within a row be sorted in ascending order of column), but some algorithms do. I left the parameter in with a default of `false` for now in case in the future we ever start using algorithms which require sorted matrices, it will be easy to switch.

Right now there are only 3 ways of constructing a `CSRMatrix`:

1. As an empty matrix, using a constructor,
2. From a `COO_Matrix` using `CSRMatrix::fromCOO`, and
3. Using a `CSRBuilder`

The `CSRBuilder` class allows you to make a `CSRMatrix` row-by-row and element-by-element, and keep tracks of additional information that ensures that any `CSRMatrix` that results from the builder is valid.

For an example of using a `CSRBuilder` to build a matrix, see [the unit tests](https://github.com/ORNL/GridKit/blob/293516ff207ddfdc9c241d258f69e172aab01f12/tests/UnitTests/LinearAlgebra/SparseMatrix/CSRTests.hpp#L140-L149).

`CSRBuilder` has two additional helpful template parameters:

1. `INCLUDE_DIAGONALS` will ensure that the constructed matrix always keeps track of elements on the diagonal, even if the user doesn't specify a value for them. This is nice when you expect to want to perform some `axpy`-type operation with a diagonal matrix, such as in the case of jacobians for implicit problems.
2. `USE_TEMPLATE`, which will only be true when calling `CSRBuilder::fromTemplate`, and re-uses the row and column index vectors and the value buffer of another matrix. This is handy for when you construct a matrix with same sparsity pattern but different values over and over (such as when computing Jacobians).

When `USE_TEMPLATE` is true, the construction of matrices is extremely fast. For example, this matrix construction:
```C++
auto build = [](auto builder)
{
    builder.row(0).elem(2, 3).elem(3, 4).elem(5, 6);
    builder.row(1).elem(0, 1).elem(2, 2);
    builder.row(3).elem(1, 2).elem(3, 3);
    return builder;
};
```
 
compiles down to just:

```asm
movzx   ecx, byte ptr [rsp]
mov     byte ptr [rsp + 112], cl
movups  xmm0, xmmword ptr [rsp + 8]
movups  xmmword ptr [rsp + 120], xmm0
mov     rcx, qword ptr [rsp + 24]
mov     qword ptr [rsp + 136], rcx
xorps   xmm0, xmm0
movups  xmmword ptr [rax], xmm0
mov     qword ptr [rax + 16], 0
movups  xmm1, xmmword ptr [rsp + 32]
movups  xmmword ptr [rsp + 144], xmm1
mov     rcx, qword ptr [rsp + 48]
mov     qword ptr [rsp + 160], rcx
movups  xmmword ptr [rax + 24], xmm0
mov     qword ptr [rax + 40], 0
movups  xmm1, xmmword ptr [rsp + 56]
movups  xmmword ptr [rsp + 168], xmm1
mov     rcx, qword ptr [rsp + 72]
mov     qword ptr [rsp + 184], rcx
movups  xmmword ptr [rax + 48], xmm0
mov     qword ptr [rax + 64], 0
mov     ecx, dword ptr [rsp + 80]
mov     dword ptr [rsp + 192], ecx
movups  xmm0, xmmword ptr [rax + 80]
movups  xmmword ptr [rsp + 200], xmm0
mov     rax, qword ptr [rax + 96]
mov     qword ptr [rsp + 216], rax
lea     rdi, [rsp + 424]
lea     rsi, [rsp + 112]
```

[(see full example in GodBolt)](https://godbolt.org/z/eTve4GWWW)

As of right now, there is no way to modify a `CSRMatrix` after it has been constructed. I'm anticipating the need for a diagonal `axpy`-like operation for the purposes of using jacobians in implicit problems, but I haven't implemented this yet, as I'm not yet sure what form it will need to look like (and a general-purpose `axpy` is a very  heavy operation that I'm not sure we will need). Please let me know if there are any other types of modification that will be necessary to add.

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
See https://github.com/ORNL/GridKit/pull/225 for an example of a component using the new CSRBuilder to construct its jacobian


shakedregev:
@alexander-novo KLU consumes CSC matrices. If you feed it a CSR matrix, it will read it as the transpose. That said, it does have a transpose solve, so it should be fine.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+     */
+    class MaybeSorted
+    {
+    protected:
+      /// @brief Flag which tracks whether or not the object is sorted
+      bool sorted_;
+
+      /**
+       * @brief Cannot be constructed, except as a base class
+       * @param isSorted Whether or not it should start sorted.
```
shakedregev:
```suggestion
       * @param[in] isSorted - Boolean indicating if it should start sorted.
```

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+      }
+    };
+
+    // Forward-declaration
+    template <class, typename, bool, bool, bool>
+    class CSRBuilder;
+
+    /**
+     * @brief Helper type - derive from this depending on whether your type is always sorted or not
+     * @tparam KEEP_SORTED Whether your type is always sorted
```
shakedregev:
```suggestion
     * @tparam KEEP_SORTED Whether your type is always sorted.
```

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+      sort();
+
+      return CSRMatrix<ScalarT, IdxT, true>(std::move(values_), std::move(row_indices_), std::move(col_indices_), num_cols_, true);
+    }
+
+    /**
+     * @brief Return the value of a particular element.
+     * @param row The row of the element. This is bounds-checked in debug mode,
+     *            but an out-of-bounds access otherwise is UB.
+     * @param col The column of the elements. Bounds-checked always, since it needs to be searched for.
```
shakedregev:
```suggestion
     * @param[in] row - The row of the element. This is bounds-checked in debug mode,
     *            but an out-of-bounds access otherwise is UB.
     * @param[in] col - The column of the elements. Bounds-checked always, since it needs to be searched for.
```

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+
+    /**
+     * @brief A helpful builder utility to help build `CSRMatrix` from their components. When built in debug mode,
+     * enables several checks which ensure that you are building the CSR matrix correctly.
+     * @tparam INCLUDE_DIAGONALS Whether or not we should ensure that empty diagonal elements are created, even if
+     *                           not specified. This way if we axpy with a diagonal matrix later (a common operation
+     *                           with Jacobian matrices) we can do it quickly.
+     * @tparam KEEP_SORTED       Whether or not the sorting of rows should be enforced. If false, then elements in a row
+     *                           can be inserted in arbitrary order.
+     * @tparam USE_TEMPLATE      Whether or not a template matrix should be used. A template matrix should ideally be
```
shakedregev:
```suggestion
     * @param[in] INCLUDE_DIAGONALS - Whether or not we should ensure that empty diagonal elements are created, even if
     *                           not specified. This way if we axpy with a diagonal matrix later (a common operation
     *                           with Jacobian matrices) we can do it quickly.
     * @param[in] KEEP_SORTED  -     Whether or not the sorting of rows should be enforced. If false, then elements in a row
     *                           can be inserted in arbitrary order.
     * @param[in] USE_TEMPLATE -     Whether or not a template matrix should be used. A template matrix should ideally be
```
Also, why are these capital?

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+      {
+        // We may have skipped some rows at the end - finish those entries up
+        skipRows(mat_.numRows());
+
+        // Setup the row index for the row beyond the last row
+        nextRow(mat_.numRows());
+
+        validate();
+        return std::move(mat_);
+      }
```
shakedregev: Why is it good to have something that irreparably finalizes the matrix?

shakedregev (changes requested):
I still need to run this. 
I think the documentation and variable names need some consistency and to follow our protocols.

In terms of major changes: 

1. I think we always want diagonal entries. Numerical solvers may otherwise fail or behave poorly and regularization becomes a big hassle.
2. I think we always want sorted matrices. Any negligible gains in runtime will be wiped out by developer time. These bugs are almost impossible to find.

@pelesh - what's your input on this?

shakedregev:
Tests pass.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+      // Explicit copying
+      CSRMatrix(const CSRMatrix& other)            = default;
+      CSRMatrix& operator=(const CSRMatrix& other) = default;
+
+      template <class, typename, bool>
+      friend class CSRMatrix;
+
+    public:
+      CSRMatrix(CSRMatrix&& other) = default;
+      CSRMatrix(size_t num_rows, IdxT num_cols);
```
Steven-Roberts: It's unclear if this usage of both `size_t` and `IdxT` here and throughout this file is intentional? I think using `IdxT` everywhere might be cleaner.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,737 @@
...
+              sorted_indices.begin() + row_indices_[row],
+              sorted_indices.begin() + row_indices_[row + 1],
+              [&](size_t a, size_t b)
+              { return col_indices_[a] < col_indices_[b]; });
+        }
+
+        // Create the new col_indices_ and values_ array which hold the same elements,
+        // but are instead sorted.
+        std::vector<IdxT>    new_col_indices(col_indices_.size());
+        std::vector<ScalarT> new_values(values_.size());
```
Steven-Roberts: Possible optimization: It should be possible to permute these in place rather than allocating extra vectors.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
alexander-novo: These are not parameters to a function (`param`), they are template parameters (`tparam`). As template parameters, they are essentially constants, which the [developer guidelines say to use screaming snake case](https://github.com/ORNL/GridKit/blob/develop/CONTRIBUTING.md#constants).

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
alexander-novo: This is a good question, which I will document the answer to in the documentation:

The `CSRBuilder` keeps the `CSRMatrix` in an invalid form as it constructs the matrix. When the user wants the matrix, we need to make sure that the `CSRMatrix` is valid and preserves the invariants it needs, otherwise the methods on `CSRMatrix` might start failing.

This function "finishes" the matrix to get it in a usable form, but in doing so "builds" the remainder of the matrix (for example: adding diagonal elements to the remaining rows of the matrix), and so no further calls to e.g. `row()` or `elem()` will be valid.

As well, the matrix being built is *moved* out of the builder - so the builder no longer has any valid data inside of it. Any calls to methods of the builder which modify `mat_` will cause memory issues, and so this needs to be signaled with a method that can only be called on an r-value reference.

I figured it's probably not wanted to be able to partially construct a matrix, use the partially constructed matrix for something, and then resume constructing. Please let me know if this is something that would be desirable.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
alexander-novo: This use is intentional and consistent with the use of `IdxT`: `IdxT` is specifically the contents of the `col_indices_` and `row_indices_` vectors because different libraries will expect different content types of these arrays. Any variable which is meant to by the size of a vector (here `num_rows` is the size of `row_indices_`) is `size_t` and any variable which is meant to be an index or comparable to indices (here `num_cols` which is the maximum value for elements in `col_indices_` and is not otherwise used as the size of a `vector`) is `IdxT`.

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
alexander-novo: I agree... I thought about it, but maybe I can use your comment as an excuse to actually implement it

alexander-novo:
> In terms of major changes:
> 
>   1. I think we always want diagonal entries. Numerical solvers may otherwise fail or behave poorly and regularization becomes a big hassle.

@shakedregev 

 A lot of the matrices being constructed are actually blocks of some larger matrix. Some blocks will be on the diagonal, and we want to contain diagonal elements, but some matrices are off-diagonal blocks and won't end up containing diagonal elements of the larger matrix.

shakedregev:
> > In terms of major changes:
> > 
> > 1. I think we always want diagonal entries. Numerical solvers may otherwise fail or behave poorly and regularization becomes a big hassle.
> 
> @shakedregev
> 
> A lot of the matrices being constructed are actually blocks of some larger matrix. Some blocks will be on the diagonal, and we want to contain diagonal elements, but some matrices are off-diagonal blocks and won't end up containing diagonal elements of the larger matrix.

What is the value of having the blocks in csr if they are meaningless alone? You can keep them in COO until populating the matrix.

alexander-novo:
> > > In terms of major changes:
> > > 
> > > 1. I think we always want diagonal entries. Numerical solvers may otherwise fail or behave poorly and regularization becomes a big hassle.
> > 
> > 
> > @shakedregev
> > A lot of the matrices being constructed are actually blocks of some larger matrix. Some blocks will be on the diagonal, and we want to contain diagonal elements, but some matrices are off-diagonal blocks and won't end up containing diagonal elements of the larger matrix.
> 
> What is the value of having the blocks in csr if they are meaningless alone? You can keep them in COO until populating the matrix.

I believe (and may be mistaken) that we're moving away from using COO at all. We are currently already using COO and converting to CSR when we need it, so this PR isn't really useful if we aren't switching away from using COO for the blocks.

alexander-novo:
@Steven-Roberts can you give me your thoughts on the new `CSRMatrix::sort()`?

alexander-novo:
See #225 for an example of a component using the new `CSRBuilder` to construct its jacobian

src/LinearAlgebra/SparseMatrix/CSRMatrix.hpp
```diff
@@ -0,0 +1,791 @@
...
+          // So prev_elem is the last element in the cycle. Move our previously noted element into it,
+          // and mark it as part of the cycle.
+          col_indices_[prev_elem] = std::move(temp_col);
+          values_[prev_elem]      = std::move(temp_val);
+          is_sorted[prev_elem]    = true;
+        }
+
+        // Make sure to mark this matrix as sorted.
+        MaybeSorted::sorted_ = true;
+      }
```
Steven-Roberts: > @Steven-Roberts can you give me your thoughts on the new CSRMatrix::sort()?

```suggestion
      if constexpr (KEEP_SORTED) {
        return;
      }

      const auto max_nnz_row = std::transform_reduce(
        std::next(row_indices_.cbegin()),
        row_indices_.cend(),
        row_indices_.cbegin(),
        IdxT{0},
        [](auto a, auto b) { return std::max(a, b); },
        std::minus<IdxT>{}  
      );
      std::vector<size_t> sorted_indices(max_nnz_row);
      std::iota(sorted_indices.begin(), sorted_indices.end(), 0);

      for (size_t row_idx = 0; row_idx < numRows(); row_idx++)
      {
        const auto row_start = row_indices_[row_idx];
        const auto nnz_row = row_indices_[row_idx + 1] - row_start;
        std::sort(
          sorted_indices.begin(),
          std::next(sorted_indices.begin(), nnz_row),
          [&](auto a, auto b) {
            return col_indices_[a + row_start] < col_indices_[b + row_start];
        });

        for (size_t col_idx = 0; col_idx < nnz_row; col_idx++) {
          auto cur_col_idx = col_idx;
          while (col_idx != sorted_indices[col_idx]) {
            const auto new_col_idx = sorted_indices[col_idx];
            std::swap(col_indices_[cur_col_idx + row_start], col_indices_[new_col_idx + row_start]);
            std::swap(values_[cur_col_idx + row_start], values_[new_col_idx + row_start]);
            std::swap(sorted_indices[col_idx], sorted_indices[new_col_idx]);
            cur_col_idx = new_col_idx;
          }
        }
      }

      MaybeSorted::sorted_ = true;
```
This is a simpler implementation which replaces the 2 buffers of size NNZ with a single buffer of size max NNZ in a row.

alexander-novo: Thanks, I had a similar idea but didn't think about it enough to convince myself that `is_sorted` isn't necessary. I'll add comments to document this fact.

pelesh:
The proposed design needs more discussion. Closing this PR for now.

PR #225, opened 2025-09-05
Add CSR Jacobian interface to Evaluator

alexander-novo:
## Description
 
This PR adds:

1. An interface in `Evaluator` for evaluating jacobians in CSR form. By default, these methods do nothing and indicate that a component does not have a CSR jacobian available.
2. A mixin for implementing the new interface. Many components will implement the interface in a very similar way, so this mixin reduces boilerplate and the code surface if this common pattern needs to change later.
3. The CSR jacobian of a sample component - `PowerElectronics/Resistor`, using the mixin.
4. A test which verifies the jacobian of a component calculated in COO form is identical to the jacobian calculated in CSR form.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments


alexander-novo:
There was a linking error on the CI build - can someone try building it on their machine, I'm not really sure why it's failing linking?

nkoukpaizan:
> There was a linking error on the CI build - can someone try building it on their machine, I'm not really sure why it's failing linking?

I can reproduce the linking error:
```
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/test_jac.dir/runCSRJacobianTests.cpp.o: in function `GridKit::Mixin::Evaluator::CSRJacobian<double, unsigned long, GridKit::Resistor<double, unsigned long> >::evaluateCSRJacobian()':
/lustre/orion/scratch/nkouk/stf006/Codes/GridKit/src/Model/EvaluatorMixins.hpp:81:(.text._ZN7GridKit5Mixin9Evaluator11CSRJacobianIdmNS_8ResistorIdmEEE19evaluateCSRJacobianEv[_ZN7GridKit5Mixin9Evaluator11CSRJacobianIdmNS_8ResistorIdmEEE19evaluateCSRJacobianEv]+0xcd): undefined reference to `GridKit::LinearAlgebra::CSRMatrix<double, unsigned long, false> GridKit::Resistor<double, unsigned long>::buildCSRJacobian<true, false, true>(GridKit::LinearAlgebra::CSRBuilder<double, unsigned long, true, false, true>)'
make[2]: *** [tests/UnitTests/LinearAlgebra/SparseMatrix/CMakeFiles/test_jac.dir/build.make:98: tests/UnitTests/LinearAlgebra/SparseMatrix/test_jac] Error 1
make[1]: *** [CMakeFiles/Makefile2:4931: tests/UnitTests/LinearAlgebra/SparseMatrix/CMakeFiles/test_jac.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

You may need to create a target for the mixin or declare an instantiation?

src/Model/Evaluator.hpp
```diff
@@ -95,6 +96,34 @@ namespace GridKit
 
       virtual std::vector<ScalarT>&       getAdjointIntegrand()       = 0;
       virtual const std::vector<ScalarT>& getAdjointIntegrand() const = 0;
+
+      using CSRJacobian = LinearAlgebra::CSRMatrix<ScalarT, IdxT, false>;
```
shakedregev: I think our convention would be to use "Csr", rather than "CSR".

src/Model/EvaluatorMixins.hpp
```diff
@@ -0,0 +1,99 @@
...
+       * @code{.cpp}
+       * template <class ScalarT, typename IdxT>
+       * class Component : public virtual Evaluator<ScalarT, IdxT>, public Mixin::Evaluator::CSRJacobian<ScalarT, IdxT, Resistor<ScalarT, IdxT>>
+       * {
+       *  public:
+       *    const static size_t SIZE = 2;
+       *
+       *    template <bool INCLUDE_DIAGONALS, bool KEEP_SORTED, bool USE_TEMPLATE>
+       *    CSRJacobian buildCSRJacobian(LinearAlgebra::CSRBuilder<ScalarT, IdxT, INCLUDE_DIAGONALS, KEEP_SORTED, USE_TEMPLATE> builder);
+       * };
```
shakedregev: Maybe turn this example to a test?

shakedregev (commented):
Fix the conventions, test, and linking error.

src/Model/Evaluator.hpp
alexander-novo: You're right, let me fix

src/Model/EvaluatorMixins.hpp
alexander-novo: It's kind of wonky because this class implements the mixin properly, but doesn't fully implement `Evaluator` so you wouldn't be able to instantiate the class. And because it's a template class, the compiler is perfectly fine for it to be wrong until something tries to actually call the function.

I do have a test with `Resistor` that does a similar thing but with a real component, so it can actually be instantiated.

shakedregev: Is this contributing beyond said test? Maybe just refer to it?

pelesh:
The proposed design needs more discussion. Closing for now ...

Issue #227, opened 2025-09-10
Bug: JSON parser infers data type from input parameter value


pelesh:
### Description how to reproduce the bug

JSON parser infers parameter data type from its value. This among other things will cause that floating point parameter is interpreted as integer if it is entered without decimal point in the input file. Retrieving this parameter as a floating point type will then cause an exception to be throw:
```
terminate called after throwing an instance of 'std::bad_variant_access'
Abort trap: 6
```

Possible solutions are
- to interpret every parameter as the floating point number
- have separate enums for real, integer and boolean parameters.

@superwhiskers 

### GridKit™ version

develop @ b191e6619bfd7d70a47e24b41fac238e3de1b9f3

### System and environment details

Observed on MacOS 15.6.1 with gcc 13.3.0 and clang 16.0.6.

pelesh:
Actually, even adding decimal point at the end does not help. One needs to add at lease one decimal place:
- Setting `Pmax: 1` in JSON gives integer.
- Setting `Pmax: 1.` is interpreted as integer, as well.
- Setting `Pmax: 1.0` is interpreted as real parameter.

pelesh:
CC @abirchfield @lukelowry

superwhiskers:
i'd say this was more underspecified behavior if anything, not a bug.

if we really want implicit conversion to double we can change the parsing code to use the [implicit conversion functionality](https://github.com/nlohmann/json?tab=readme-ov-file#implicit-conversions) of the json library when retrieving values expected to be reals. that would work. then we explicitly state in the format spec that `1` (or any other integer literal) is interpreted as a real if the expected type is a real.

an alternative solution would be to just specify in the format that you need to have the decimal place with a zero at minimum. that would also solve the problem of it being unspecified.

pelesh:
> an alternative solution would be to just specify in the format that you need to have the decimal place with a zero at minimum. that would also solve the problem of it being unspecified.

Per @abirchfield, this would be difficult to enforce and would cause a lot of bug reports down the road.

abirchfield:
To me the simplest solution for now is to just assume every parameter is floating point. Then we can have parameter checks if we need to restrict some to integer values.

superwhiskers:
we don't really need to do that. there's the implicit conversion flag i mentioned above that will automatically convert to floating point if you try to access a value that can be coerced to floating point. iirc the only thing you'd need to do is set a define or something

pelesh:
[...]

PR #228, opened 2025-09-10
Add test to verify reading and configuring signal ports by JSON parser.

pelesh:
## Description
 
Implement tests for JSON parser that verifies signal ports are correctly processed. 
 
 ## Proposed changes
 
This is a step towards completing a prototype for instantiating the entire system automatically from a JSON input file. In this PR we verify that JSON parser correctly reads and sets data for all supported components.

The second part of this would be to update `SystemModel` class to use parsed data to instantiate system model. This will be handled in a separate PR.
 
 ## Checklist
 

- [x] All tests pass.
 
 ## Further comments
 
Notably, there is an issue in parser that will infer type of component model parameters based on how they are entered in the input file (with or without decimal point). See #227.

### DIFF
--- a/src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
+++ b/src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
@@ -64,6 +64,7 @@ namespace GridKit
         }
         else
         {
+          std::cout << raw_port.key() << "\n";
           throw std::runtime_error("Invalid port mapping");
         }
       }
--- a/src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
+++ b/src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
@@ -45,7 +45,7 @@ namespace GridKit
             c.parameters[key.value()] = raw_parameter.value().template get<IdxT>();
           }
           else
-          {            
+          {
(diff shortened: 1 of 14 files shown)


src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
```diff
@@ -51,6 +51,7 @@ namespace GridKit
         }
         else
         {
+          // std::cout << "Key " << raw_parameter.key() << " has no value!\n";
```
nkoukpaizan, line 54: Don't forget to remove.

src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
```diff
@@ -64,6 +65,7 @@ namespace GridKit
         }
         else
         {
+          // std::cout << raw_port.key() << "\n";
```
nkoukpaizan, line 68: Don't forget to remove.

nkoukpaizan (approved):
Looks good to me and test pass.

src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
pelesh: I'll need these for debugging for the next PR.

PR #230, opened 2025-09-24
Use JSON parser in 2-bus example with governor

pelesh:
## Description
 
In 2-bus example with governor, replace manual setting of the `SystemModelData` instance with automated setting from JSON input file, using existing input file parser in GridKit.

 @abdourahmanbarry @alexander-novo @reid-g 
 
 ## Proposed changes
 
For now I suggest adding `TwoBusTgov1Json` example, which uses parser, in addition to `TwoBusTgov1` example, which sets up system model configuration manually. This is for additional verification of the parser and JSON input file format. Once the parser code is more mature, we can remove old example(s).

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp
+++ b/examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp
@@ -1,5 +1,5 @@
 /**
- * @file TwoBusTgov1SignalJson.cpp
+ * @file TwoBusBasic.cpp
  * @author Adam Birchfield (abirchfield@tamu.edu)
  * @author Slaven Peles (peless@ornl.gov)
  * @brief Example running a 2-bus system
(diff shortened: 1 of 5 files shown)


nkoukpaizan (approved):
Looks good to me. I would leave this open until @abirchfield  and @lukelowry have had a chance to review.

I wonder if it might be a good time to have actual `json` files, rather than implementing them within the code.

examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp
```diff
@@ -0,0 +1,231 @@
...
+  using real_type   = double;
+  using index_type  = size_t;
+
+  std::cout << "Example: TwoBusBasicJson\n";
+
+  //
+  // Input file
+  //
+
+  const char input_file[] =
```
lukelowry, line 38: This looks great. I would second @nkoukpaizan in requesting we have actual .json files if possible. I have a draft PR with the file format of a 9-bus case, and other larger cases prepared in the GK format. They are quite large and debugging them or testing them with this temporary string representation may be difficult.

lukelowry (changes requested):
Please let me know if changing this to read a true JSON file is difficult or out of scope of the PR.

examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp
pelesh: I would prefer to do that in a separate PR. I just want to get basic functionality in place and verified in this PR.

pelesh: I need to make some fixes to CMake for examples and components and will create JSON input files repository as a part of that PR.

pelesh:
Closing in favor of #233

PR #236, opened 2025-09-26
Two-Area Case (10 Bus) in JSON File Format

lukelowry:
## Description
 
Larger case in the specified file format of GridKit. This [two-area Kundur case](https://electricgrids.engr.tamu.edu/electric-grid-test-cases/two-area-system/) has been modified to include exciters, governors and machine models for each generator.

A README is included with a one-line diagram and a table with the models used in the case. 
 
 ## Further comments
 
This will provide data to test the SystemModel JSON parser with .json files and larger systems.


abirchfield (approved):

lukelowry:
Added a README with a one-line diagram, description of models used, and a list of events that we should aim to implement with this case. Once the case description README and this JSON are approved, I will add simulated testbench data so that any contributor can work on implementing tests with this case,  without needing external tools to validate. @abirchfield 

@pelesh If you have feedback on useful information you would like to see in these Case file READMEs, please let me know. This is a good time to decide how we want to present the cases built into GridKit. Thank you

pelesh (approved):
This is good to be merged.

Question: The model in the figure has 4 transformers, but we do not have a transformer model implemented in GridKit yet. Do we need to put this on our to-do list?

Nitpicking: There are some characters in the figure not displaying correctly. Please try to fix at your convenience.

pelesh:
CC @PhilipFackler

lukelowry:
@pelesh 

> Question: The model in the figure has 4 transformers, but we do not have a transformer model implemented in GridKit yet. Do we need to put this on our to-do list?

Transformers and Lines are treated identically within the positive sequence phasor paradigm, and that's why we classify them both more generally as Branches. More complicated transformer models in phasor dynamics exist, but w.l.o.g. we treat them the same as lines.

> Nitpicking: There are some characters in the figure not displaying correctly. Please try to fix at your convenience.

I took this straight from the repository website, but I agree that this is bothersome. I will make an Issue and come up with an unenforced 'one-line style guide' in the meantime.

pelesh:
> @pelesh If you have feedback on useful information you would like to see in these Case file READMEs, please let me know. This is a good time to decide how we want to present the cases built into GridKit. Thank you

The README file looks good to me. You might want to add to [`CONRTIBUTING.md`](https://github.com/ORNL/GridKit/blob/develop/CONTRIBUTING.md) a checklist of what a README file for a new example should include.

pelesh:
> Once the case description README and this JSON are approved, I will add simulated testbench data so that any contributor can work on implementing tests with this case, without needing external tools to validate.

I am guessing this data would be specific to a fault and when/where it is introduced in the system.

lukelowry:
> I am guessing this data would be specific to a fault and when/where it is introduced in the system.

Correct. We will need to decide what location/type/duration of fault is ideal for testing before we add it to the repo. 

Also, the issue has been created #238 regarding one-line and documentation. I am going to address this before PR for the larger case (Hawaii) so that we set a good precedent.

PR #240, opened 2025-09-30
Tiny ThreeBus JSON Examples

lukelowry:
## Description
 
Add JSON formatted case for ThreeBus Tiny example, for both Classical and GENROU.
 
 ## Further comments
 
Need to discuss initialization parameters for generators, $p0$ and $q0$, which are static here. For now this PR provides the file format to begin testing JSON parser.


pelesh (approved):
I suggest a follow-up PR where you hard-wire these files as char arrays in the existing examples and use parser to read the data and instantiate the system.

PR #243, opened 2025-10-02
Refactored TinyTwoBus examples to use separate JSON files

### LINKED ISSUE #235: Tiny phasor dynamics examples should not use hard-wired input files
### Summary

Modify "tiny" phasor domain dynamics examples to use external JSON input files instead of hard-wired ones in the example code.

### Rationale

Debugging JSON input files is much easier when done on a standalone file than on char array embedded in the C++ code. For larger examples it is infeasible to have input files embedded in C++ code to begin with.

Perhaps a good first step in creating more generic examples would be to create file I/O for the "tiny" phasor dynamics examples as they are simple and mostly feature complete.

### Description

- Create directory with JSON input files. Follow the subdirectory structure of the `examples` directory.
- Ensure that tiny phasor dynamics examples run correctly when launched from the build and install directory alike.
- Manage input file path installation in CMake.

### Additional information

_No response_

PhilipFackler:
## Description
 
`TinyTwoBus*Json.cpp` examples are refactored to read from a file given on the command line (rather than define the JSON in a raw string literal). These json files are added to the build tree--and in the install prefix (under `share`) using the same directory structure as the `examples` in the source tree.
 
Closes #235 
 
 ## Proposed changes
 
This change (in addition to what is described above) establishes a convention that can easily be generally applied to any other examples that can use JSON files.
 
 ## Checklist
  
- [ ] All tests pass.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -3,6 +3,33 @@
 #    - Slaven Peles <peless@ornl.gov>
 #]]
 
+set(GRIDKIT_EXAMPLES_INSTALL_ROOT "share/gridkit/examples")
+
+#
+# gridkit_example_add_file(<filename>)
+#
+# This macro will add a file (e.g., JSON) that is needed for an example to the
+# build tree and the install tree following the same directory structure as the
+# source tree under `examples`.
+# 
+# <filename> must exist in the CMAKE_CURRENT_SOURCE_DIR where this macro is
+# called.
+#
+macro(gridkit_example_add_file _filename)
+    get_filename_component(__fname ${_filename} NAME)
+    set(_source_filepath ${CMAKE_CURRENT_SOURCE_DIR}/${__fname})
+    if(NOT EXISTS ${_source_filepath})
+        message(FATAL_ERROR "File does not exist: ${_source_filepath}")
+    endif()
+    set(_binary_filepath ${CMAKE_CURRENT_BINARY_DIR}/${__fname})
+    configure_file(${_source_filepath} ${_binary_filepath} COPYONLY)
+    cmake_path(RELATIVE_PATH CMAKE_CURRENT_SOURCE_DIR
+        BASE_DIRECTORY ${CMAKE_SOURCE_DIR}/examples
+        OUTPUT_VARIABLE _rel_path)
+    set(_ex_install_path ${GRIDKIT_EXAMPLES_INSTALL_ROOT}/${_rel_path})
+    install(FILES ${_filename} DESTINATION ${_ex_install_path})
+endmacro()
+
 add_subdirectory(LinearAlgebra)
 
 add_subdirectory(PowerFlow)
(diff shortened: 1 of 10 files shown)


examples/CMakeLists.txt
```diff
@@ -3,6 +3,15 @@
...
 #]]
 
+set(GRIDKIT_INSTALL_EXAMPLES_ROOT "share/gridkit/examples")
+
+macro(gridkit_example_install_path out_var)
+    cmake_path(RELATIVE_PATH CMAKE_CURRENT_SOURCE_DIR
+        BASE_DIRECTORY "${CMAKE_SOURCE_DIR}/examples"
+        OUTPUT_VARIABLE _rel_path)
+    set(${out_var} "${GRIDKIT_INSTALL_EXAMPLES_ROOT}/${_rel_path}")
+endmacro()
```
pelesh: - Do we need a macro for this? It is a one-liner, so macro looks like an overkill.
- Do we need to have literal path? What if we drop the quotation marks around examples install dir?

PhilipFackler: - I think defining the macro makes the code for individual example targets more readable (and more likely to be consistent). I considered having a macro that would handle adding the json file to the build tree and install, and now that I think about it again, that would probably be better.
- I tend to use quotes around paths in case they have spaces (I've been bitten before), but I'm happy to remove them.

PhilipFackler: Something like this:
```cmake
macro(gridkit_example_add_file _filename)
    cmake_path(RELATIVE_PATH CMAKE_CURRENT_SOURCE_DIR
        BASE_DIRECTORY "${CMAKE_SOURCE_DIR}/examples"
        OUTPUT_VARIABLE _rel_path)
    set(${out_var} "${GRIDKIT_INSTALL_EXAMPLES_ROOT}/${_rel_path}")
    configure_file(
        "${CMAKE_CURRENT_SOURCE_DIR}/${_filename}"
        "${CMAKE_CURRENT_BINARY_DIR}/${_filename}"
        COPYONLY)
    install(FILES ${_filename} DESTINATION "${_ex_install_path}")
endmacro()
```

So the example CMakeLists.txt could just do this:
```cmake
gridkit_example_add_file(TwoBusBasic.json)
```

pelesh:
@PhilipFackler please make sure all tests pass.

PhilipFackler:
> @PhilipFackler please make sure all tests pass.

@pelesh It looks like [this job](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243) is  assuming my branch is in the base repo. Should I be working from here instead of my fork?

`origin` is set to the [base repo](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243#step:2:41) and then it tries to access my branch under `origin` [here](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243#step:2:54).

shakedregev:
> > @PhilipFackler please make sure all tests pass.
> 
> @pelesh It looks like [this job](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243) is assuming my branch is in the base repo. Should I be working from here instead of my fork?
> 
> `origin` is set to the [base repo](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243#step:2:41) and then it tries to access my branch under `origin` [here](https://github.com/ORNL/GridKit/actions/runs/18202836600/job/51825826222?pr=243#step:2:54).

We typically just branch off (not fork off), so let's say yes.

PR #244, opened 2025-10-02
JSON Load and Gen Values on MVA Base

lukelowry:
## Description
 
This PR corrects some elements of the JSON cases that have been added thusfar. Loads are added as constant impedance on the system MVA base. Additionally, we correct the generator p0 and q0 on the machine MVA base. This was done for the following JSON case formats:
- Three Bus Basic (Genrou)
- Three Bus Classical
- Two-Area
- Synthetic Hawaii

Lastly, for the three-bus cases, the appropriate bus was set to `infinite_bus` class.

 ## Further comments
 
Once we can parse cases from the JSON format correctly, we need to include a new field for the Machine model MVA Base. Right now all machine parameters are given on each respective generator base, but this base is not specified.


abirchfield (approved):
We discussed these changes offline -- looks good. Expect more changes will be needed once we can parse these and validate, but we know we need to make these changes.

PR #245, opened 2025-10-02
Adding Genrou Test for PhasorDynamics

joshua-x-7:
## Description
 Adding another test for the Genrou model with hard-coded values to test residual in equations.

 ## Proposed changes
- Added function `hard_coded_residual`
- Function hard codes Genrou parameter and states, and verifies that the residual function matches the hard-coded answer key.
 
 ## Checklist
- [x] All tests pass.
 
 ## Further comments
 Any suggestions for my first PR are appreciated. Thank you.


abirchfield (approved):
New test is formulated well, and runs correctly on my build. Looks good. You will need to rebase before it can be merged (and please be careful doing so).

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -101,6 +102,111 @@ namespace GridKit
...
+                                                 1.8896749891587163,
+                                                 1.4000000000000001,
+                                                 0.31000000000000005,
+                                                 2.211,
+                                                 0.85,
+                                                 1.2,
+                                                 0.9846153846153847,
+                                                 -1.8230769230769228,
+                                                 -1.0846153846153845,
+                                                 -0.9269230769230766};
```
pelesh: Selecting answer key like this, with 16 digits, makes it specific to the `double` data type. It would be good to make the answer key more generic, using for example overloaded trigonometric functions, multiples/fractions of pi, or divisions like this `static_cast<ScalarT>(2.)/static_cast<ScalarT>(3.)` instead of `0.6666666666666667`.

You can use similar tricks for input parameters, such as `p0` and `q0`, to make it easier to define answers in a compact generic form.

pelesh (commented):
I suggest trying to make these tests generic with respect to the floating point data type. It is a little tricky but I think it is doable with a reasonable effort.

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
abirchfield: I do not think more than 10 digits of precision would ever be needed for an engineering application of this model. Perhaps the simpler fix would be to set the tolerance above to the max of 1e-10 and 10 * std::numeric_limits<ScalarT>::epsilon()

Also, I think getting generic form answers will end up being quite impractical for the range of models we intend to develop. Even here, we would need quite complex expressions with many sines, cosines, and rather large fractions for some of these equations. If, alternatively, we set trivial or unrealistic values to the parameters we end up triggering unrealistic situations (such as if Xdp = Xl, then we get whole terms with (Xdp-Xl) as disappearing to zero).

pelesh: I agree that for practical applications we don't need more than a 4-5 significant digits. The purpose of these tests is to verify numerics, not the user end accuracy, though. The models perform algebraic operations and there is no reason for them not to produce results accurate to machine precision. If there is a small discrepancy, this means something has not been implemented correctly. These small errors tend to accumulate and distort results in large systems where debugging such issues is difficult. This is why we want to catch them early in unit tests.

pelesh: > Also, I think getting generic form answers will end up being quite impractical for the range of models we intend to develop. Even here, we would need quite complex expressions with many sines, cosines, and rather large fractions for some of these equations. If, alternatively, we set trivial or unrealistic values to the parameters we end up triggering unrealistic situations (such as if Xdp = Xl, then we get whole terms with (Xdp-Xl) as disappearing to zero).

To get around this, we could hard-wire `ScalarT` to `double` type and declare we are testing only for double precision. If and when we decide to use single or quad precision, we could then decide how we want to test those.

For now, @joshua-x-7, please code all numbers in the answer key in scientific notation with 16 significant digits and add a comment that the answer key is available only in double precision and therefore only double precision tests are done at this time. I think that should be good for now.

abirchfield: > I agree that for practical applications we don't need more than a 4-5 significant digits. The purpose of these tests is to verify numerics, not the user end accuracy, though. The models perform algebraic operations and there is no reason for them not to produce results accurate to machine precision. If there is a small discrepancy, this means something has not been implemented correctly. These small errors tend to accumulate and distort results in large systems where debugging such issues is difficult. This is why we want to catch them early in unit tests.

I see your point here. Yes, these tests should be checking the algebraic operations (up to the machine precision we are using).

abirchfield: > > Also, I think getting generic form answers will end up being quite impractical for the range of models we intend to develop. Even here, we would need quite complex expressions with many sines, cosines, and rather large fractions for some of these equations. If, alternatively, we set trivial or unrealistic values to the parameters we end up triggering unrealistic situations (such as if Xdp = Xl, then we get whole terms with (Xdp-Xl) as disappearing to zero).
> 
> To get around this, we could hard-wire `ScalarT` to `double` type and declare we are testing only for double precision. If and when we decide to use single or quad precision, we could then decide how we want to test those.
> 
> For now, @joshua-x-7, please code all numbers in the answer key in scientific notation with 16 significant digits and add a comment that the answer key is available only in double precision and therefore only double precision tests are done at this time. I think that should be good for now.

I think this is a good compromise for now.

This made me interested, and I actually tried to run this test with "float" instead of "double." There were several compile errors in the main code I had to fix by adding instantiations of Genrou and Bus with float, plus some other things. And there were more warnings I did not fix, but ultimately the test passes with no changes to the test code. So that is a good sign. I think it is obvious that quad precision would fail as it is currently written. But perhaps we can cross that bridge if/when we come to it?

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -101,6 +102,113 @@ namespace GridKit
...
+
+        ScalarT Vr1{1.0}; ///< Bus real voltage
+        ScalarT Vi1{0};   ///< Bus imaginary voltage
+
+        PhasorDynamics::Bus<ScalarT, IdxT>    bus(Vr1, Vi1);
+        PhasorDynamics::Genrou<ScalarT, IdxT> gen(&bus, 1, p0, q0, H, D, Ra, Tdop, Tdopp, Tqopp, Tqop, Xd, Xdp, Xdpp, Xq, Xqp, Xqpp, Xl, S10, S12);
+
+        // Answer key is available only in double precision.
+        // Therefore, only double precision tests are done at this time.
+        const std::vector<double> res_answer = {
```
pelesh: The solution vector elements are of type `ScalarT`. In the instantiation of tests, we instantiate only the class with template parameter `ScalarT==double`
```suggestion
        const std::vector<ScalarT> res_answer = {
```

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -101,6 +102,113 @@ namespace GridKit
...
+            -4.2476190476190490e-01,
+            -5.4750000000000000e+01,
+            -9.6000000000000010e+00,
+            2.3786666666666663e+00,
+            2.1000000000000002e-01,
+            -7.0000000000000000e-02,
+            -1.9223748416156686e-01,
+            1.8896749891587163e+00,
+            1.4000000000000001e+00,
+            3.1000000000000005e-01,
```
pelesh: Are these numbers really supposed to be 1.4 and 3.1? It seems to me that the 16th digits here are affected by numerical round off error. Please check.

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -101,6 +102,113 @@ namespace GridKit
...
+        ScalarT Vi1{0};   ///< Bus imaginary voltage
+
+        PhasorDynamics::Bus<ScalarT, IdxT>    bus(Vr1, Vi1);
+        PhasorDynamics::Genrou<ScalarT, IdxT> gen(&bus, 1, p0, q0, H, D, Ra, Tdop, Tdopp, Tqopp, Tqop, Xd, Xdp, Xdpp, Xq, Xqp, Xqpp, Xl, S10, S12);
+
+        // Answer key is available only in double precision.
+        // Therefore, only double precision tests are done at this time.
+        const std::vector<double> res_answer = {
+            -3.7699111843077515e+02,
+            -1.1111111111111112e+00,
```
pelesh:
```suggestion
            -static_cast<ScalarT>(10.)/static_cast<ScalarT>(9.),
```

pelesh: This number is -10/9.

tests/UnitTests/PhasorDynamics/GenrouTests.hpp
```diff
@@ -19,8 +19,9 @@ namespace GridKit
     class GenrouTests
     {
     private:
-      using real_type = typename PhasorDynamics::Component<ScalarT, IdxT>::real_type;
-
+      using real_type               = typename PhasorDynamics::Component<ScalarT, IdxT>::real_type;
+      static constexpr ScalarT tol_ = 10 * std::numeric_limits<ScalarT>::epsilon(); // added this: was not originally there
+      // static constexpr ScalarT tol_ = .00001;
     public:
```
pelesh, line 25: Remove commented out code.

pelesh (changes requested):
More due diligence is needed in determining where numbers in the answer key come from. There are some apparent numerical artifacts in the answer key. Quite a few numbers in the answer key can be represented in a generic form (e.g. 10/9 = 1.1111...).

joshua-x-7:
I messed up the rebase. I can make a clean PR once these changes are satistfactory.

joshua-x-7:
> I messed up the rebase. I can make a clean PR once these changes are satistfactory.

This issue has been resolved. Rebase is now done correctly.

pelesh (approved):
Good job!

pelesh:
Tests were failing due to unrelated bug.

PR #250, opened 2025-10-03
Refactored TinyTwoBus examples to use separate JSON files

### LINKED ISSUE #235: Tiny phasor dynamics examples should not use hard-wired input files

PhilipFackler:
## Description
 
`TinyTwoBus*Json.cpp` examples are refactored to read from a file given on the command line (rather than define the JSON in a raw string literal). These json files are added to the build tree--and in the install prefix (under `share`) using the same directory structure as the `examples` in the source tree.
 
Closes #235 
 
 ## Proposed changes
 
This change (in addition to what is described above) establishes a convention that can easily be generally applied to any other examples that can use JSON files.
 
 ## Checklist
  
- [x] All tests pass.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -3,6 +3,33 @@
 #    - Slaven Peles <peless@ornl.gov>
 #]]
 
+set(GRIDKIT_EXAMPLES_INSTALL_ROOT "share/gridkit/examples")
+
+#
+# gridkit_example_add_file(<filename>)
+#
+# This macro will add a file (e.g., JSON) that is needed for an example to the
+# build tree and the install tree following the same directory structure as the
+# source tree under `examples`.
+# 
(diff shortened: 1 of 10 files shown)


shakedregev (approved):
Looks good. Tests pass and works as advertised.

PR #259, opened 2025-10-08
Change label to "BusFault" to match the input guidelines

PhilipFackler:
## Description

Input guidelines specify "BusFault" as the [device](https://github.com/ORNL/GridKit/blob/develop/src/Model/PhasorDynamics/INPUT_FORMAT.md#device-classes) class name.

 ## Proposed changes
 
This adds the "BusFault" as an additional option, keeping the existing "bus_fault" label in case that's being used.
 
 ## Checklist
- [X] All tests pass.

### DIFF
--- a/examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasic.json
+++ b/examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasic.json
@@ -65,7 +65,7 @@
             "mon": ["delta", "omega"]
         },
         { 
-            "class": "bus_fault",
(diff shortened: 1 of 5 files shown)


nkoukpaizan:
I would remove the non-compliant `bus_fault` label. And replace wherever it was used.

pelesh:
I think this is good to merge but would prefer if @abirchfield could give his blessing first. Just to make sure we are consistent with the _intent_ of our JSON format.

abirchfield (approved):
This is good. More aligns with our other naming conventions.

Issue #234, opened 2025-09-26
Build issue: GenrouEnzyme intermittently fails to build


nkoukpaizan:
### Build error message(s)

I have observed a few instances of the compilation of GenrouEnzyme failing. The compilation usually completes after running `make` again.  I thought this was resolved with `__attribute__((always_inline)) inline` and enzyme@0.0.196, but it still occurs intermittently. 

Error message:
```console
clang++: /lustre/orion/stf006/scratch/nkouk/Codes/GridKit/buildsystem/spack/opt/spack/linux-zen3/llvm-19.1.7-5o343ldg2sog2rxd3ldqqcuxmrqzx5s7/include/llvm/IR/User.h:170: llvm::Value* llvm::User::getOperand(unsigned int) const: Assertion `i < NumUserOperands && "getOperand() out of range!"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
[...]

PR #265, opened 2025-10-09
Model Jacobians with respect to external variables

nkoukpaizan:
## Description

This adds Jacobian evaluation for internal residuals with respect to external variables ($\partial{f}/\partial{w_e}$) with Enzyme.
cc @abdourahmanbarry @reid-g @alexander-novo 

 ## Proposed changes
- Added `we_` and `we_indices_` to store a copy of the external variables and their indices. The indices are initialize to a unique integer identifyier, such that nothing is stored in the Jacobian unless signal nodes are connected and the indices are updated.
- Updated GovernorTgov1Tests to test the derivative with respect to the external variable `omega`. Both the Enzyme and DependencyTracking Jacobians provide one more, off-diagonal, entry.
- Updated `GenrouEnzyme` to compute the additional Jacobian blocks and verified that it behaves as expected. This will be enabled when #234 is fixed.
- Updated `SignalNode` to store a pointer to the variable's global index. 
- Updated `ComponentSignals` with a method to read the index from the signals.

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Added 3, 10, and 37 bus test cases.
 - Updated documentation.
 - Added JSON parsing.
-- Automatic differentiation with enzyme.
+- Automatic differentiation with enzyme (w.r.t. internal, bus and external variables).
 - Added PR and issue templates.
 - Added Genrou class unit test and example.
 - Added Dev container.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Added 3, 10, and 37 bus test cases.
 - Updated documentation.
 - Added JSON parsing.
-- Automatic differentiation with enzyme (w.r.t. internal, bus and external variables).
+- Automatic differentiation with enzyme (w.r.t. internal and external variables).
 - Added PR and issue templates.
 - Added Genrou class unit test and example.
 - Added Dev container.
--- a/src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
+++ b/src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
@@ -38,23 +38,32 @@ namespace GridKit
       };
 
       /**
-       * @brief Residual wrapper around residual methods inside model classes
+       * @brief Template definition for wrapper around residual methods inside model classes
        *
        * @tparam ModelT - model type
        * @tparam MemberFunctions - member function parameter key
        * @tparam ScalarT - scalar data type
        *
-       * @param[in] model - Pointer to the model to be differentiated
-       * @param[in] y - Internal variables
-       * @param[in] yp - Internal variable derivatives
-       * @param[in] wb - Coupling (bus) variables
-       * @param[in] we - External variables
-       * @param[out] f - Internal residual
-       * @param[out] h - Coupling (bus) residual
        */
       template <typename ModelT, MemberFunctions function, typename ScalarT>
       struct Wrapper
       {
+      };
+
+      /**
+       * @brief Residual wrapper partial template specialization for InternalResidual
+       *
+       */        
+      template <typename ModelT, typename ScalarT>
+      struct Wrapper<ModelT, MemberFunctions::InternalResidual, ScalarT>
+      {
+        /**
+         * @param[in] model - Pointer to the model to be differentiated
+         * @param[in] y - Internal variables
+         * @param[in] yp - Internal variable derivatives
+         * @param[in] wb - Bus variables
+         * @param[out] f - Internal residual
+         */
         static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* f)
         {
           model->evaluateInternalResidual(y, yp, wb, f);
@@ -68,9 +77,17 @@ namespace GridKit
       template <typename ModelT, typename ScalarT>
       struct Wrapper<ModelT, MemberFunctions::InternalResidualWithExternal, ScalarT>
       {
-        static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* we, ScalarT* h)
+        /**
+         * @param[in] model - Pointer to the model to be differentiated
+         * @param[in] y - Internal variables
+         * @param[in] yp - Internal variable derivatives
+         * @param[in] wb - Bus variables
+         * @param[in] we - Signal variables
+         * @param[out] f - Internal residual
+         */
+        static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* we, ScalarT* f)
         {
-          model->evaluateInternalResidualWithExternal(y, yp, wb, we, h);
+          model->evaluateInternalResidualWithExternal(y, yp, wb, we, f);
         }
       };
 
@@ -81,6 +98,13 @@ namespace GridKit
       template <typename ModelT, typename ScalarT>
       struct Wrapper<ModelT, MemberFunctions::BusResidual, ScalarT>
       {
+        /**
+         * @param[in] model - Pointer to the model to be differentiated
+         * @param[in] y - Internal variables
+         * @param[in] yp - Internal variable derivatives
+         * @param[in] wb - Bus variables
+         * @param[out] h - Bus residual
+         */
         static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* h)
         {
           model->evaluateBusResidual(y, yp, wb, h);
@@ -94,6 +118,13 @@ namespace GridKit
       template <typename ModelT, typename ScalarT>
       struct Wrapper<ModelT, MemberFunctions::BusResidual11, ScalarT>
       {
+        /**
+         * @param[in] model - Pointer to the model to be differentiated
+         * @param[in] y - Internal variables
+         * @param[in] yp - Internal variable derivatives
+         * @param[in] wb - Bus variables
+         * @param[out] h - Bus residual
+         */
         static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* h)
         {
           model->evaluateBusResidual11(y, yp, wb, h);
@@ -107,6 +138,13 @@ namespace GridKit
       template <typename ModelT, typename ScalarT>
       struct Wrapper<ModelT, MemberFunctions::BusResidual12, ScalarT>
       {
+        /**
+         * @param[in] model - Pointer to the model to be differentiated
+         * @param[in] y - Internal variables
+         * @param[in] yp - Internal variable derivatives
+         * @param[in] wb - Bus variables
+         * @param[out] h - Bus residual
+         */
         static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* h)
         {
           model->evaluateBusResidual12(y, yp, wb, h);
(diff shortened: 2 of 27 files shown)


src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
```diff
@@ -46,16 +47,30 @@ namespace GridKit
        * @param[in] model - Pointer to the model to be differentiated
        * @param[in] y - Internal variables
        * @param[in] yp - Internal variable derivatives
-       * @param[in] w - Coupling (bus) variables
+       * @param[in] wb - Coupling (bus) variables
+       * @param[in] we - External variables
```
pelesh: The documentation is inconsistent with the struct and functions. Why are we separating `we` and `wb`? Both are external variables.

src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
```diff
@@ -92,9 +107,9 @@ namespace GridKit
       template <typename ModelT, typename ScalarT>
       struct Wrapper<ModelT, MemberFunctions::BusResidual12, ScalarT>
       {
-        static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* w, ScalarT* h)
+        static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* h)
         {
-          model->evaluateBusResidual12(y, yp, w, h);
+          model->evaluateBusResidual12(y, yp, wb, h);
         }
```
pelesh, line 151: This is specific to the line model. A generic implementation would be preferred.

pelesh: Similar for `evaluateBusResidual11`, `evaluateBusResidual21` and `evaluateBusResidual22`.

src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
```diff
@@ -259,7 +275,70 @@ namespace GridKit
                          const std::map<IdxT, IdxT>&                        var_indices,
                          ScalarT*                                           y,
                          ScalarT*                                           yp,
-                         ScalarT*                                           w,
+                         ScalarT*                                           wb,
+                         GridKit::LinearAlgebra::COO_Matrix<ScalarT, IdxT>& jac)
+        {
```
pelesh, line 332: We really need to get rid of `COO_Matrix` class and replace it with `CsrMatrix`.

Let's keep it for now and follow with the next PR where we will replace `COO_Matrix` for good.

CC @alexander-novo

src/Model/PhasorDynamics/Branch/Branch.hpp
```diff
@@ -47,7 +47,7 @@ namespace GridKit
       using Component<ScalarT, IdxT>::yp_;
       using Component<ScalarT, IdxT>::tag_;
       using Component<ScalarT, IdxT>::f_;
-      using Component<ScalarT, IdxT>::w_;
+      using Component<ScalarT, IdxT>::wb_;
```
pelesh, line 50: Why this change?

src/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -286,10 +294,11 @@ namespace GridKit
         // Input Variables
         if (signals_.template isAttached<Tgov1ExternalVariables::DELTAOMEGA>())
         {
-          omega_ = signals_.template readExternalVariable<Tgov1ExternalVariables::DELTAOMEGA>();
+          we_[0]         = signals_.template readExternalVariable<Tgov1ExternalVariables::DELTAOMEGA>();
+          we_indices_[0] = signals_.template readExternalVariableIndex<Tgov1ExternalVariables::DELTAOMEGA>();
         }
 
-        evaluateInternalResidual(y_.data(), yp_.data(), w_.data(), f_.data());
+        evaluateInternalResidualWithExternal(y_.data(), yp_.data(), wb_.data(), we_.data(), f_.data());
```
pelesh: Why this name change? The previous name `evaluateInternalResidual` was clear. Internal residual evaluation should include external (coupling) variables.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -272,20 +272,25 @@ namespace GridKit
...
 
       // Resize coupling data
-      w_.resize(2);
+      wb_.resize(2);
       h_.resize(2);
 
+      // Resize external variable data
+      we_.resize(2);
+      we_indices_[0] = static_cast<IdxT>(-1);
+      we_indices_[1] = static_cast<IdxT>(-1);
```
pelesh: Perhaps we should define integer constant like
```c++
constexpr IdxT INVALID_INDEX = static_cast<IdxT>(-1);
```
and use that throughout the code.

@PhilipFackler any suggestions?

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 - Added 3, 10, and 37 bus test cases.
 - Updated documentation.
 - Added JSON parsing.
-- Automatic differentiation with enzyme.
+- Automatic differentiation with enzyme (w.r.t. internal, bus and external variables).
 - Added PR and issue templates.
```
pelesh: Bus variables _are_ external variables. This wording/notation may be confusing. Consider using something like bus and signal variables or external variables (bus) and external variables (signal).

pelesh (changes requested):
There are a few minor issues that need to be addressed.

More concerning is:
- Use of specialized functions like `evaluateBusResidual11`.
- Use of `COO_Matrix` class instead of a CSR matrix class.

These two issues may be out of scope here.

pelesh:
> * Updated `SignalNode` to store a pointer to the variable's global index.

Why is signal node holding pointer instead of the index? Is it because you expect global indexing to change at runtime?

nkoukpaizan:
> > * Updated `SignalNode` to store a pointer to the variable's global index.
> 
> Why is signal node holding pointer instead of the index? Is it because you expect global indexing to change at runtime?

This is to get around the question of when the global indices are set. In a previous version of the implementation, I had to introduce a separate function that sets the signal only after the indices are set (i.e., after and outside the `allocate` methods).

CHANGELOG.md
nkoukpaizan: I agree that `external` is confusing here, since bus variables are also external. These are meant to point to the external variable coming from signals.

src/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
nkoukpaizan: The fundamental difference between `we` and `wb` is how we are using these variables. Buses own the variables (and associated residual terms); Signals don't own the variables, and the residual contributions are part of what the models own. The associated Jacobian terms require different implementations.

CHANGELOG.md
pelesh: Why not `(w.r.t. internal and external variables)`?

nkoukpaizan: Yes, I'll change the wording.

alexander-novo:
Some thoughts I have related to this PR:

Components which are connected to multiple nodes (such as branches) currently have a separate function for every possible combination of input variables and outputs. So, for a `Branch` we have `evaluateBusResidual11`, `evaluateBusResidual12`, `evaluateBusResidual21`, `evaluateBusResidual22`. I know of at least one component we want in the future which will have connections to at least 3 nodes (transformers). Will this have 9 different `evaluateResidual` functions?

Most of the residual functions in `Branch` are fairly identical - it kind of seems to me like it would be easier to maintain a function which is templated on the buses rather than 4 functions with different names.

Also looking forward into the future - will there be any other kinds of nodes besides buses and signal nodes? If so, the API surface will start ballooning worrying about all the different kinds of combinations of nodes a component could or could not be attached to. Maybe a function like `evaluateNodeResidual` which works on any kind of node might be better than separate functions for every kind of node or pair of nodes.

Finally, I'm not sure if pairwise node contribution functions are necessary. Why not have something like `evaluateBusResidual1()` which evaluates the residual for bus 1 based on input variables from *all* connected nodes?

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
PhilipFackler: That makes sense. I've used this convention before to define an "invalid" value for any numeric type:
```c++
template <typename T>
inline constexpr T invalid = std::numeric_limits<T>::max() - static_cast<T>(1);
```
and then you could define
```c++
inline constexpr IdxT INVALID_INDEX = invalid<IdxT>;
```
If you don't anticipate needing the generic functionality, then use the simplified version you're suggesting.

PhilipFackler: (I used "max - 1" there because I used "max" for a "wildcard" value which was applicable in that library. Use whatever value makes sense here.)

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
```diff
@@ -182,14 +182,16 @@ namespace GridKit
...
-        auto tol = 1000 * std::numeric_limits<real_type>::epsilon(); ///< Todo: Resolve mismatch from sigmoid approximation
+        auto tol = 10 * std::numeric_limits<real_type>::epsilon();
         for (size_t i = 0; i < dependency_tracking_residuals.size(); ++i)
         {
           DependencyTracking::Variable                       res           = dependency_tracking_residuals[i];
           const DependencyTracking::Variable::DependencyMap& dependencies  = res.getDependencies();
           success                                                         *= (GridKit::Testing::isEqual(dependencies, enzyme_jacobian[i], tol));
         }
 
+        success.expectFailure(); // TODO: Resolve mismatch from sigmoid approximation
```
PhilipFackler, line 193: nice

src/Model/PhasorDynamics/Branch/Branch.hpp
PhilipFackler: If we're changing this and adding a similar one down the line (`we_`) why not make them more explicit, like `w_bus_` and `w_ext_` (if I'm understanding correctly)?

PhilipFackler:
I agree with the comments about `evaluateBusResidual##`, but they seem out of scope for this PR. Someone should probably create an issue that explains what we'd like to see happen with those. :)

CHANGELOG.md
nkoukpaizan: Addressed in [f30fcd0](https://github.com/ORNL/GridKit/pull/265/commits/f30fcd012276bb4d6e845fc5d5e7737c3f0ed3f4)

src/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
nkoukpaizan: Addressed in [f5f6196](https://github.com/ORNL/GridKit/pull/265/commits/f5f61964d49b4b014cc5586ccd6114cc5469b665).

I made the change when I needed the wrapper to call these with different number of arguments (if the model has signals or not):
- Overloading the method within the wrapper confused the compiler. 
- I did not want to extend the interface for models that don't have signals
The solution I propose in [f5f6196](https://github.com/ORNL/GridKit/pull/265/commits/f5f61964d49b4b014cc5586ccd6114cc5469b665) is to keep the `evaluateInternalResidual` method names and limit the change to the `InternalResidualWithSignals` template parameter.

pelesh:
> Finally, I'm not sure if pairwise node contribution functions are necessary. Why not have something like `evaluateBusResidual1()` which evaluates the residual for bus 1 based on input variables from _all_ connected nodes?

This is a good question and this is how most of the phasor dynamics modeling tools assemble system model. The issue here is that each bus is connected to different number of other buses, so we are not guaranteed to have SIMD parallelism. On the other hand, each branch is connected to 2 buses, so we can evaluate branches contributions in a SIMD manner. This would allow us to run these computations on GPUs or similar accelerator devices.

pelesh:
> So, for a `Branch` we have `evaluateBusResidual11`, `evaluateBusResidual12`, `evaluateBusResidual21`, `evaluateBusResidual22`. I know of at least one component we want in the future which will have connections to at least 3 nodes (transformers). Will this have 9 different `evaluateResidual` functions?

A more general approach would be for each component to provide an array of Jacobian contributions and a mapping to the system Jacobian.

pelesh (approved):

PR #269, opened 2025-10-10
Add more information and context to JSON parser errors

### LINKED ISSUE #263: More helpful error messages in JSON parsers
### Summary

Error messages need to guide the user as much as possible. We should provide more information in the error messages in the JSON parser(s). 

### Rationale

Users providing JSON input won't generally know the implementation details of the code.

### Description

This applies to exceptions thrown in the following files:
- src/Model/PhasorDynamics/SystemModelDataJSONParser.hpp
- src/Model/PhasorDynamics/ComponentDataJSONParser.hpp
- src/Model/PhasorDynamics/Bus/BusDataJSONParser.hpp

Incorrect parameters in JSON files should be reported to the user with as much specificity as possible. Give the parameter name and as much context as possible (_e.g._, device id) so the problem can be located in the JSON file by the user.

### Additional information

I have done two small PRs related to this (#258 and #262 ), but they don't address the larger issue. These error messages should be consistent.

There are probably more more ways that input checks could be done that would give more help. This request is mainly about improving the existing error messages. But once it is done, it should set the precedent for further error reporting.

PhilipFackler:
## Description
 
Error messages gave no helpful information to users.
 
 _Closes #263_
 
 ## Proposed changes
 
Provide relevant information about the error and the context by which to locate it in the JSON file.
 
 ## Checklist
 
- [X] All tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Added signal node connectivity tools.
 - Upgraded to Spack 1.0.0.
 - Added `GenClassical` as supported device class in JSON parser.
+- Added more context information to JSON parser errors.
 
 ## v0.1
 
--- a/src/Model/PhasorDynamics/Bus/BusDataJSONParser.hpp
+++ b/src/Model/PhasorDynamics/Bus/BusDataJSONParser.hpp
@@ -19,6 +19,11 @@ namespace GridKit
     {
       j.at("name").get_to(bd.name);
 
+      std::stringstream error_context;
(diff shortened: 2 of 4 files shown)


pelesh (approved):
The current solution violates xSDK mandatory policy [M11](https://github.com/xsdk-project/xsdk-community-policies/blob/master/package_policies/M11.md). It would be good to manage output in GridKit in a similar way it is done in Re::Solve library. See #270.

PR #273, opened 2025-10-16
Specify `CXX` to avoid unnecessary C language checks

PhilipFackler:
## Description

Since GridKit doesn't use C, we can leave it out of the build configuration.
 
 ## Proposed changes
 
Specify `CXX` as the project language.
 
 ## Checklist
 
- [X] All tests pass.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@
 
 cmake_minimum_required(VERSION 3.13)
 
-project(gridkit)
+project(gridkit CXX)
 
 set(PACKAGE_NAME "GRIDKIT")
 set(PACKAGE_STRING "GRIDKIT 0.1.0")


nkoukpaizan:
Looks good to me. Not sure whether this was intentionally not set. @PhilipFackler was there an issue you wanted to address with this change? Does it save a significant amount of build time?

PhilipFackler:
> Looks good to me. Not sure whether this was intentionally not set. @PhilipFackler was there an issue you wanted to address with this change? Does it save a significant amount of build time?

It didn't cause a problem exactly. I just noticed in my clang build that the compilers were inconsistent. I had specified `CMAKE_CXX_COMPILER` as `clang++-16`, but the C compiler was still the system gcc. That didn't seem right and there are extra checks that cmake has to do for C. So, since it seems like a C compiler isn't actually necessary, I thought it best to exclude it.

nkoukpaizan (approved):

PR #274, opened 2025-10-16
Introduce Logger class with tests

### LINKED ISSUE #129: add a logging faculty
currently, writing to stdout/stderr is handled with plain `std::cout`, `std::cerr`, which doesn't allow for the kind of fine-grained configuration of logging levels and modules that an end-user may want. 

we should add a proper logging faculty and avoid using the standard output/error handles entirely.

### LINKED ISSUE #270: Add output management tool to GridKit
### Summary

Add output management tool to allow GridKit application to output messages at different verbosity level, including an option to suppress all the output.

### Rationale

Having an option to suppress all the output is required to conform to xSDK [standards](https://github.com/xsdk-project/xsdk-community-policies/blob/master/package_policies/M11.md).

### Description

I suggest we take [`Logger`](https://github.com/ORNL/ReSolve/tree/develop/resolve/utilities/logger) class from Re::Solve and adapt it for GridKit. 

### Additional information

_No response_

PhilipFackler:
## Description
 
This introduces a logging tool so that we can move away from using `std::cout`/`std::cerr` directly.

 _Closes #270_ 
 _Closes #129_

 ## Proposed changes
 
The new class and its tests were lifted directly from Re::Solve, see [here](https://github.com/ORNL/ReSolve/tree/develop/resolve/utilities/logger) and [here](https://github.com/ORNL/ReSolve/tree/develop/tests/unit/utilities/logger).
 
 ## Checklist
 
- [X] All tests pass.

### DIFF
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -2,3 +2,5 @@ add_library(Utilities INTERFACE)
 target_include_directories(Utilities INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_library(GRIDKIT::Utilities ALIAS Utilities)
+
+add_subdirectory(Logger)
--- a/src/Utilities/Logger/CMakeLists.txt
+++ b/src/Utilities/Logger/CMakeLists.txt
@@ -1,0 +1,7 @@
+gridkit_add_library(utilities_logger
+  SOURCES
+    Logger.cpp
+  OUTPUT_NAME
+    gridkit_utilities_logger)
+
+target_link_libraries(Utilities INTERFACE GRIDKIT::utilities_logger)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Upgraded to Spack 1.0.0.
 - Added `GenClassical` as supported device class in JSON parser.
 - Added more context information to JSON parser errors.
+- Added `Logger` class.
 
 ## v0.1
 
--- a/src/Utilities/Colors.hpp
+++ b/src/Utilities/Colors.hpp
@@ -1,0 +1,19 @@
+#pragma once
+
+namespace GridKit
+{
+  namespace Utilities
+  {
+    namespace Colors
+    {
+      // must be const pointer and const dest for
+      // const string declarations to pass -Wwrite-strings
+      static const char* const RED    = "\033[1;31m";
+      static const char* const GREEN  = "\033[1;32m";
+      static const char* const YELLOW = "\033[33;1m";
+      static const char* const BLUE   = "\033[34;1m";
+      static const char* const ORANGE = "\u001b[38;5;208m";
+      static const char* const CLEAR  = "\033[0m";
+    } // namespace Colors
+  } // namespace Utilities
+} // namespace GridKit
--- a/src/Utilities/Logger/Logger.cpp
+++ b/src/Utilities/Logger/Logger.cpp
@@ -1,0 +1,187 @@
+/**
+ * @file Logger.cpp
+ * @brief Contains definition of Logger class.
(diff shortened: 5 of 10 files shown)


src/Utilities/Logger.hpp
```diff
@@ -0,0 +1,72 @@
...
+  {
+    // must be const pointer and const dest for
+    // const string declarations to pass -Wwrite-strings
+    static const char* const RED    = "\033[1;31m";
+    static const char* const GREEN  = "\033[1;32m";
+    static const char* const YELLOW = "\033[33;1m";
+    static const char* const BLUE   = "\033[34;1m";
+    static const char* const ORANGE = "\u001b[38;5;208m";
+    static const char* const CLEAR  = "\033[0m";
+  } // namespace colors
```
nkoukpaizan: There is another definition in https://github.com/ORNL/GridKit/blob/develop/src/Utilities/TestHelpers.hpp#L19. Consider consolidating.

src/Utilities/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-add_library(Utilities INTERFACE)
-target_include_directories(Utilities INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
-
-add_library(GRIDKIT::Utilities ALIAS Utilities)
+gridkit_add_library(Utilities
+  SOURCES
+    Logger.cpp
+  OUTPUT_NAME
+    gridkit_utilities)
```
nkoukpaizan: I would use a different target (e.g., `gridkit_logger` or `gridkit_utilities_logger`) if this only has the logger.

src/Utilities/Logger.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * @brief Manages and logs outputs from GridKit code.
+   *
+   * All methods and data in this class are static.
+   *
+   */
+  class Logger
```
nkoukpaizan: Consider putting inside a namespace, e.g., `Utilities` or `IO`. Some of the code that is directly in the `GridKit` namespace is older code that we haven't updated after changing the directory structure.

nkoukpaizan (changes requested):
This works. I have a few suggestions on the target name choices, consolidating some definitions and namespaces.

src/Utilities/Logger.hpp
PhilipFackler: Would it make sense to put this by itself in src/Utilities/Colors.hpp?

src/Utilities/CMakeLists.txt
PhilipFackler: Well, I'm also working on a `CliOptions` class that will go in here. I'm happy to split them apart, but I assumed there would be other things thrown into `utilities` in the future. Let me know what y'all prefer.

nkoukpaizan: Understood. I think it will be useful to split them. A user of these targets may want one and not the other.

PhilipFackler: Sounds good.

PhilipFackler: Alright, I think I have done it according to the convention in PhasorDynamics.

src/Utilities/Logger/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+gridkit_add_library(utilities_logger
+  SOURCES
+    Logger.cpp
+  OUTPUT_NAME
+    gridkit_utilities)
```
nkoukpaizan:
```suggestion
    gridkit_utilities_logger)
```

PhilipFackler: Fixed (thought I had already done that)

nkoukpaizan (approved):

PR #275, opened 2025-10-17
Add basic command-line parser

### LINKED ISSUE #112: Add command line input parser
### Summary

It would be helpful for examples and tests in GridKit to take command line input such as to provide a verbose output or to save output to a file. 

### Rationale

Having a common command line input parser for the entire GridKit would help streamline implementation of examples and tests.

### Description

Consider re-using parser from [Re::Solve](https://github.com/ORNL/ReSolve/blob/develop/resolve/utilities/params/CliOptions.hpp).

### Additional information

_No response_

### LINKED ISSUE #249: Add a command line input parser to GridKit
### Summary

Add a class similar to [ReSolve's](https://github.com/ORNL/ReSolve/blob/develop/resolve/utilities/params/CliOptions.cpp) for command line input.

### Rationale

Right now we just have to pass arguments in order. This becomes cumbersome if we want to skip some arguments, or allow users to input them in arbitrary order (as long as they are marked).

### Description

I think something very similar (or identical) to what ReSolve has would be good. We should also modify the examples to that they use this new format.

### Additional information

_No response_

PhilipFackler:
## Description
  
This gives users the ability to query the command-line arguments for specific expected options.

_Closes #249_
_Closes #112_
 
 ## Proposed changes
 
Lifted `CliOptions` class from Re::Solve and modified to simplify interface (user doesn't have to parse manually or use a `std::unique_ptr<std::pair<std::string, std::string>>`)
 
 ## Checklist
 
- [X] All tests pass.
 
 ## Further comments
 
This provides the same basic functionality as it did for Re::Solve. However, I think it would be nice to expand the design to make using it in apps less cumbersome. Specifics should go in a new issue.

### DIFF
--- a/src/Utilities/TestHelpers.hpp
+++ b/src/Utilities/TestHelpers.hpp
@@ -9,6 +9,7 @@
 
 #define THROW_NULL_DEREF throw std::runtime_error("error")
 
+#include <array>
 #include <cmath>
 #include <iostream>
 #include <limits>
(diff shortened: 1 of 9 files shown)


nkoukpaizan:
Overall looks good to me. Please rebase and apply the structure we settled on in #274.

PhilipFackler:
@nkoukpaizan @pelesh I added a function in TestHelpers.hpp to test for throwing (either a specific exception type or anything). I didn't see a tool for this anywhere else, but if you do already have it, let me know.

src/Utilities/TestHelpers.hpp
```diff
@@ -255,5 +256,88 @@ namespace GridKit
 
     // };
 
+    namespace Detail
+    {
```
pelesh, line 260: Why not use namespace `Testing`?

pelesh (approved):
Looks good to merge provided comments by @nkoukpaizan are addressed.

nkoukpaizan (approved):
Yes, my comments have been addressed. Looks good.

src/Utilities/TestHelpers.hpp
PhilipFackler: It holds the implementation "details" for the `throws` function. The helper structs are not meant to be used directly.

shakedregev (commented):
Changes are sensible. There are a few comments I'd edit for grammar, but we can skip it. Mainly I'm concerned that I can no longer build GridKit. I'll try everything from scratch.

shakedregev (approved):
Looks good post rebase.

PR #286, opened 2025-10-23
Machine MVA Bases

lukelowry:
## Description
 
Add support for machine models to have a different MVA base than the system.
 
 ## Proposed changes
 
- Add `mva_base` parameter to GenClassical and Genrou
- Remove `va_base` and  `freq_base` from the general Component class (every model handles this differently, and it could actually be misleading to offer this as an option for models like exciter and governor)
- Default mva_base is 100 MVA.
- Updated the respective INPUT_FORMAT specifications
 
 ## Checklist
 
- [X] All tests pass.
 
 ## Further comments
 
This is a minor fix and does not need to be included as a change feature.

### DIFF
--- a/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
+++ b/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
@@ -526,7 +526,7 @@ namespace GridKit
       evaluateInternalResidual(y_.data(), yp_.data(), w_.data(), f_.data());
       evaluateBusResidual(y_.data(), yp_.data(), w_.data(), h_.data());
 
-      // Genrou contribution to bus algebraic equations 
+      // Genrou contribution to bus algebraic equations
       Ir() += h_[0];
       Ii() += h_[1];
 
--- a/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
+++ b/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
@@ -307,8 +307,8 @@ namespace GridKit
       /* Initialization tricks -- assuming NO saturation */
       ScalarT vr    = Vr();
       ScalarT vi    = Vi();
-      ScalarT p     = static_cast<ScalarT>(p0_) * 100.0 / mva_base_;
-      ScalarT q     = static_cast<ScalarT>(q0_) * 100.0 / mva_base_;
+      ScalarT p     = static_cast<ScalarT>(p0_) * mva_system_base_ / mva_base_;
+      ScalarT q     = static_cast<ScalarT>(q0_) * mva_system_base_ / mva_base_;
       ScalarT vm2   = vr * vr + vi * vi;
       ScalarT Er    = vr + (Ra_ * p * vr + Ra_ * q * vi - Xq_ * p * vi + Xq_ * q * vr) / vm2;
       ScalarT Ei    = vi + (Ra_ * p * vi - Ra_ * q * vr + Xq_ * p * vr + Xq_ * q * vi) / vm2;
@@ -485,8 +485,8 @@ namespace GridKit
       ScalarT vi  = w[1];
 
       // Current base conversion. Assumes generator and bus are same V base
-      h[0] = (inr - vr * G_ + vi * B_) * mva_base_ / 100.0;
-      h[1] = (ini - vr * B_ - vi * G_) * mva_base_ / 100.0;
+      h[0] = (inr - vr * G_ + vi * B_) * mva_base_ / mva_system_base_;
+      h[1] = (ini - vr * B_ - vi * G_) * mva_base_ / mva_system_base_;
 
       return 0;
     }
--- a/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
+++ b/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
@@ -6,7 +6,6 @@
 
 #include "Genrou.hpp"
 #include <Model/PhasorDynamics/Bus/Bus.hpp>
-#include <Model/PhasorDynamics/Governor/Tgov1/Tgov1.hpp> // <- TODO: Temporary, to be removed.
 #include <Model/PhasorDynamics/SignalNode/SignalNode.hpp>
 #include <Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouData.hpp>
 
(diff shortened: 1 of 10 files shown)


abirchfield (approved):
Looks good. This will be needed for the larger case models.

src/Model/PhasorDynamics/SynchronousMachine/GenClassical/GenClassicalImpl.hpp
```diff
@@ -155,8 +162,8 @@ namespace GridKit
     {
       ScalarT vr    = Vr();
       ScalarT vi    = Vi();
-      ScalarT p     = static_cast<ScalarT>(p0_);
-      ScalarT q     = static_cast<ScalarT>(q0_);
+      ScalarT p     = static_cast<ScalarT>(p0_) * 100.0 / mva_base_;
+      ScalarT q     = static_cast<ScalarT>(q0_) * 100.0 / mva_base_;
```
pelesh: Magic number `100.0` needs to come from somewhere. Please define with the proper type (`RealT`?).

src/Model/PhasorDynamics/SynchronousMachine/GenClassical/GenClassicalImpl.hpp
```diff
@@ -248,8 +255,8 @@ namespace GridKit
     {
       const ScalarT ir = y[3];
       const ScalarT ii = y[4];
-      h[0]             = ir;
-      h[1]             = ii;
+      h[0]             = ir * mva_base_ / 100.0;
+      h[1]             = ii * mva_base_ / 100.0;
```
pelesh: Magic number in residual evaluation. Please define.

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -301,8 +307,8 @@ namespace GridKit
       /* Initialization tricks -- assuming NO saturation */
       ScalarT vr    = Vr();
       ScalarT vi    = Vi();
-      ScalarT p     = static_cast<ScalarT>(p0_);
-      ScalarT q     = static_cast<ScalarT>(q0_);
+      ScalarT p     = static_cast<ScalarT>(p0_) * 100.0 / mva_base_;
+      ScalarT q     = static_cast<ScalarT>(q0_) * 100.0 / mva_base_;
```
pelesh: Flagging magic numbers ...

src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -478,8 +484,9 @@ namespace GridKit
       ScalarT vr  = w[0];
       ScalarT vi  = w[1];
 
-      h[0] = inr - vr * G_ + vi * B_;
-      h[1] = ini - vr * B_ - vi * G_;
+      // Current base conversion. Assumes generator and bus are same V base
+      h[0] = (inr - vr * G_ + vi * B_) * mva_base_ / 100.0;
+      h[1] = (ini - vr * B_ - vi * G_) * mva_base_ / 100.0;
```
pelesh: Flagging magic numbers ...

pelesh (changes requested):
I guess I don't understand the intent of this PR. In my view it would make more sense to have `mva_base_system_` and `mva_base_local_` parameters and then convert units at the component-system interface. 

Magic numbers are an error prone way to go about this.

abirchfield:
> I guess I don't understand the intent of this PR. In my view it would make more sense to have `mva_base_system_` and `mva_base_local_` parameters and then convert units at the component-system interface.
> 
> Magic numbers are an error prone way to go about this.

The intent of the PR is to implement a feature of generators (different MVA bases from the system base) that will be necessary for medium and large cases. 

Yes, I think that ideally we should have a system-wide parameter `mva_base_system_` instead of using the literal 100 MVA (even though for all the cases we will consider in this project we will use a 100 MVA system base).

@lukelowry  is that a change we can make easily?  I guess I'm wondering whether the component has access to system parameters.

src/Model/PhasorDynamics/SynchronousMachine/GenClassical/GenClassicalImpl.hpp
lukelowry: I defined a private constant for `100.0`, which is the system MVA base.

lukelowry:
@abirchfield @pelesh 

I have added a private variable `mva_system_base_` to Genrou and GenClassical to remove use of magic numbers. Ideally, we should pass the system MVA base to these models, but it is unclear how to do so, and this PR blocks our ability to validate the Hawaii case.

To clarify, this PR allows us to correctly implement larger cases (such as Hawaii), as the MVA bases for each generator differ in those cases. However, there is no general procedure to do base conversions, as each model/device may implement bases differently. An MVA base does not make sense for some models. This PR allows us to begin testing the Hawaii (& larger cases) against simulated transients from external tools (Power World)

pelesh:
> I have added a private variable `mva_system_base_` to Genrou and GenClassical to remove use of magic numbers. Ideally, we should pass the system MVA base to these models, but it is unclear how to do so, and this PR blocks our ability to validate the Hawaii case.

There is system setting for MVA base specified in the JSON input file format. This should be used to set the system MVA. Perhaps the best way to go about it is to have the system MVA as the member variable of the `SystemModel` and then let each component model read that value from the system model during composition.

abirchfield:
> > I have added a private variable `mva_system_base_` to Genrou and GenClassical to remove use of magic numbers. Ideally, we should pass the system MVA base to these models, but it is unclear how to do so, and this PR blocks our ability to validate the Hawaii case.
> 
> There is system setting for MVA base specified in the JSON input file format. This should be used to set the system MVA. Perhaps the best way to go about it is to have the system MVA as the member variable of the `SystemModel` and then let each component model read that value from the system model during composition.


Prior to this PR, the system MVA base and any component MVA bases in the JSON input file are ignored. Effectively we assume the MVA base is 100 MVA everywhere. 

The purpose of this PR is to implement component-level MVA bases other than 100 MVA for GenClassical and GENROU, which we need for the next round of models. 

I do think a good future task would be to update both the parser and system composer to pass system MVA base into the component model, to allow hypothetical future cases with system MVA base other than 100 MVA. If it is simple we could go ahead and do it, but I think this could be an involved task as it would require changes to the core of the parser and system composer.

For now, would it make sense to move forward with this PR with the current scope and create an issue for system MVA base to be addressed later?

pelesh:
> For now, would it make sense to move forward with this PR with the current scope and create an issue for system MVA base to be addressed later?

In that case, I suggest to move `mva_system_base_` to the component base class, hardwire it to 100.0 and put a screaming comment that this is just a temporary solution. That would require minimal changes to the code in this branch and would have `mva_system_base_` set in only one place (where we can set it up correctly from the input file at a later time).

lukelowry:
Moved to `Component` class, comment warnings added, #296 created to track the issue, and rebased branch @pelesh

pelesh (approved):
Great!

PR #290, opened 2025-10-24
Install public headers

### LINKED ISSUE #278: Specify headers to be installed using `HEADERS` argument in `gridkit_add_library`
At the moment, headers are not being installed. We need to decide which headers should be included for each library (assuming it's not all of them).

PhilipFackler:
## Description
 
Install headers needed for writing code against installed libraries
 
 _Closes #278_

 ## Proposed changes
 
Install headers via either the `HEADERS` parameter to `gridkit_add_library` or manually with `install(FILES` as necessary.
 
 ## Checklist
 
- [X] All tests pass.
 
 ## Further comments
 
This PR does NOT install the headers for the magic_enum and json libraries. I think those should be handled in a better way, and that will be for a different PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,15 +105,6 @@ if(GRIDKIT_ENABLE_ENZYME)
   include(FindEnzyme)
 endif()
 
-# The binary dir is already a global include directory
-configure_file(
-  ${CMAKE_SOURCE_DIR}/src/Definitions.hpp.in
-  ${CMAKE_BINARY_DIR}/Definitions.hpp)
-install(
-  FILES ${CMAKE_BINARY_DIR}/Definitions.hpp
-  DESTINATION include
-  )
-
 # Macro that adds libraries
 include(GridkitAddLibrary)
 
--- a/cmake/Config.cmake.in
+++ b/cmake/Config.cmake.in
@@ -6,3 +6,10 @@
 @PACKAGE_INIT@
 
 include ( "${CMAKE_CURRENT_LIST_DIR}/GridKitTargets.cmake" )
+
+include(CMakeFindDependencyMacro)
+set(GridKit_ENABLE_SUNDIALS @GRIDKIT_ENABLE_SUNDIALS@)
+if(GridKit_ENABLE_SUNDIALS)
+  list(APPEND CMAKE_PREFIX_PATH @SUNDIALS_DIR@)
+  find_dependency(SUNDIALS)
+endif()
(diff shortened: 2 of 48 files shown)


pelesh:
> This PR does NOT install the headers for the magic_enum and json libraries. I think those should be handled in a better way, and that will be for a different PR.

If everything is implemented correctly, magic_enum and json libraries should be compiled in the parser. Their headers should not be exposed to GridKit users.

PhilipFackler:
@pelesh Right now, the user has to use `json` directly as in the [examples](https://github.com/ORNL/GridKit/blob/b77daaf0b9bbe61d202e2f6664b3207f49cd566b/examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp#L73). And `magic_enum` needs to be installed because it is included in the parser [header](https://github.com/ORNL/GridKit/blob/b77daaf0b9bbe61d202e2f6664b3207f49cd566b/src/Model/PhasorDynamics/ComponentDataJSONParser.hpp#L7). And this is not just a simple wrap job because of templates (unless the set of possible combinations of template arguments is known and relatively small).

CMakeLists.txt
```diff
@@ -105,15 +105,6 @@ if(GRIDKIT_ENABLE_ENZYME)
...
 
-# The binary dir is already a global include directory
-configure_file(
-  ${CMAKE_SOURCE_DIR}/src/Definitions.hpp.in
-  ${CMAKE_BINARY_DIR}/Definitions.hpp)
-install(
-  FILES ${CMAKE_BINARY_DIR}/Definitions.hpp
-  DESTINATION include
-  )
-
```
pelesh, line 116: Why is definition file removed?

pelesh (commented):
Perhaps I am missing something but I think it would be good to include in this PR
- GridKit consumer test
- exporting `${CMAKE_PREFIX}/include/GridKit` as the include path in `GridKitConfig.cmake`.

We just need to make sure user can build out of box.

CMakeLists.txt
PhilipFackler: I moved this code to src/CMakeLists.txt

pelesh: Sorry, my bad.

PhilipFackler:
> Perhaps I am missing something but I think it would be good to include in this PR
> 
> * GridKit consumer test
> * exporting `${CMAKE_PREFIX}/include/GridKit` as the include path in `GridKitConfig.cmake`.
> 
> We just need to make sure user can build out of box.

1. See the changes to cmake/GridkitAddLibrary.cmake
2. I'm working on another PR which handles the json and magic_enum TPLs. That PR is necessary (along with this one) before I can get a functional GridKit consumer test.

pelesh (approved):

PR #297, opened 2025-10-30
CMake project name and Config install

### LINKED ISSUE #285: Project name and target namespace need to match
### Summary

Scoped library targets need to use the namespace `GridKit` rather than the current `GRIDKIT`.

### Rationale

This is to follow cmake convention. The cmake project is called `GridKit` (based on the file name convention `GridKitConfig.cmake`). If a user calls `find_package(GridKit)` it seems like it would probably make sense to link against, for example, `GridKit::solvers_steady` rather than `GRIDKIT::solvers_steady`.

PhilipFackler:
## Description
 
Use `GridKit` (rather than `GRIDKIT`) for all user-facing project names
 
 _Closes #285_
 
 ## Proposed changes
 
Name change applies to:

- Project name
- Cache variable names
- Alias target name prefixes
- Export install namespace

I also made some changes for the installation

- Removed `PACKAGE_*` specifiers to simplify top-level CMakeLists.txt
- Enable C language for `find_dependency(SUNDIALS)` call
- Add `cxx_std_20` compile feature to targets in `gridkit_add_library` (so consumers don't get surprise errors when they don't specify the standard).
 
 ## Checklist
 
- [X] All tests pass.


pelesh:
A new example was merged before this PR, so this now breaks build configuration for the new example:

```
CMake Error at examples/PhasorDynamics/Medium/Hawaii/CMakeLists.txt:4 (target_link_libraries):
   Target "hawaiiJson" links to:

     GRIDKIT::phasor_dynamics_components

     * An ALIAS target is missing.
```

pelesh (approved):

PR #299, opened 2025-10-31
test_install in CI

nkoukpaizan:
## Description
 
This adds the consumer test to CI and fixes a related bug due to public Enzyme options.
 
 ## Proposed changes
 
- Added a `Test Installation` step to CI that runs `make test_install`
- Made Enzyme-related `target_compile_options` and `target_link_libraries` `PRIVATE`.
- Added a workaround to export sanitizers without creating separate targets. Thanks @PhilipFackler. We may want to find a more elegant solution, but that likely requires more significant changes to the `gridkit_add_library` macro or a different approach altogether.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 CMake and CI changes.

### DIFF
--- a/.github/workflows/spack_default_build.yaml
+++ b/.github/workflows/spack_default_build.yaml
@@ -179,10 +179,10 @@ jobs:
         run: spack -e . install --keep-stage --no-check-signature --no-cache --fresh
 
       - name: Test Build
-        run: cd $(spack -e . location --build-dir gridkit@develop) && ctest -VV
+        run: cd $(spack -e . location --build-dir gridkit@develop) && spack -e ${OLDPWD} build-env gridkit ctest -VV
 
       - name: Test Installation
-        run: cd $(spack -e . location --build-dir gridkit@develop) && make test_install
+        run: cd $(spack -e . location --build-dir gridkit@develop) && spack -e ${OLDPWD} build-env gridkit make test_install
 
       # Push with force to override existing binaries...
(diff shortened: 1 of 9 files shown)


pelesh (approved):
I approve but would like to get @PhilipFackler review before merging, as well.

.github/workflows/spack_default_build.yaml
```diff
@@ -179,7 +180,10 @@ jobs:
         run: spack -e . install --keep-stage --no-check-signature --no-cache --fresh
 
       - name: Test Build
-        run: cd $(spack -e . location --build-dir gridkit@develop) && ctest -VV
+        run: cd $(spack -e . location --build-dir gridkit@develop) && spack -e ${OLDPWD} build-env gridkit ctest -VV
+
+      - name: Test Installation
+        run: cd $(spack -e . location --build-dir gridkit@develop) && spack -e ${OLDPWD} build-env gridkit make test_install
```
PhilipFackler, line 186: I assume this is using the installation from line 180? And what is `OLDPWD`? I don't see that defined.

nkoukpaizan: `$OLDPWD` should be available from the environment, just like `$PWD` is. It is the directory before executing `cd`. This changes the directory to where the code was built and runs the tests there.

PhilipFackler:
Sorry for being late to the game here, but this looks great.

PR #300, opened 2025-11-04
Rename src to GridKit and standardize include statements

### LINKED ISSUE #279: Reorganize source tree and use explicit include statements
There are two things to consider here:

### 1
Headers should be installed in a directory structure matching the source tree. However, this should be modified to include a root directory named `GridKit` or `gridkit`.
   - If users install GridKit at a common prefix (like `/usr`, `/usr/local`, or `~/.local`), the current include statements would be ambiguous:
   ```c++
   #include <Model/PhasorDynamics/BusBase.hpp>
   ```
   We would want it to be this:
   ```c++
   #include <GridKit/Model/PhasorDynamics/BusBase.hpp>
   ```
   - Internal include statements (headers including other headers) would also need to be made explicit following the same format

### 2
Include statements in installed headers should work the same in the installation environment as the build environment. Assuming the change made above, this would require a mild reorganization of the source tree.

To minimize directory reorganization, the "target_include_directory" for each library target could be `src/` and a top-level directory `GridKit/` could be added so that everything currently in `src/` would be moved to `src/GridKit/`.

PhilipFackler:
## Description
 
We need a standard format for include statements in both build and install environments. This PR addresses the build environment only (since currently headers are not installed at all). Install environment will be addressed in subsequent PR.
 
 _Closes #279_

 ## Proposed changes
 
- The `src` directory is renamed to `GridKit` and all library targets use `${CMAKE_SOURCE_DIR}` (that is, the top-level source directory) as the include directory in the build environment.
- All public include statements (those in header files) are modified to follow the format:
  `#include <GridKit/Model/PhasorDynamics/BusBase.hpp>`
- `.clang-format` is updated to handle `<GridKit/*` headers
- Modified `CONTRIBUTING.md` to reflect this standard
 
 ## Checklist
 
- [x] All tests pass.


pelesh (approved):
Long overdue overhaul of include statements and GridKit installation. All tests pass and GridKit installs correctly.

examples/PowerFlow/Grid3Bus/README.md
```diff
@@ -22,7 +22,7 @@ Problem variables are voltage magnitudes and phases; they are stored in bus obje
 
 **Bus 1**: Slack bus, does not store variables no residuals. Voltage and phase are set to $`V_1 \equiv 1`$p.u. and $`\theta_1 \equiv 0`$, respectively.
 
-**Bus 2**: PQ bus, stores variables $`V_2, \theta_2`$ and residuals $`P_2, Q_2`$. Load $`P_{L1} = 2.5`$p.u., $`Q_{L1} = -j0.8`$p.u. is attached to it. From the equations for [branch](../../src/Model/PowerFlow/Branch/README.md) and [load](../../src/Model/PowerFlow/Load/README.md) components, we assemble Bus 2 residuals as:
+**Bus 2**: PQ bus, stores variables $`V_2, \theta_2`$ and residuals $`P_2, Q_2`$. Load $`P_{L1} = 2.5`$p.u., $`Q_{L1} = -j0.8`$p.u. is attached to it. From the equations for [branch](../../GridKit/Model/PowerFlow/Branch/README.md) and [load](../../GridKit/Model/PowerFlow/Load/README.md) components, we assemble Bus 2 residuals as:
```
nkoukpaizan: These links need one more `../`. They likely have been broken for a while. 

There is one more `src` --> `GridKit` needed below in this file.

nkoukpaizan (approved):
One minor documentation that needs fixing. All tests pass everywhere I have tested as well.

PR #302, opened 2025-11-06
Add verify step to SystemModel::allocate and don't throw on some parse errors

### LINKED ISSUE #266: Bug: Segfault in Genrou accessing signal ports
### Description how to reproduce the bug

If a generator device is specified in the JSON file with one or both of

- a `pmech` port without a governor assigned the same `pmech` signal
- an `efd` port without an exciter assigned the same `efd` signal

a segmentation fault results instead of a meaningful error.

Ideally this should be caught by the code and an exception thrown that notifies the user of their mistake.

### GridKit™ version

develop @ 7bcded1

- ubuntu 22.04
- clang-16

### Additional information

This happens at the point when the generator calls `writeExternalVariable` on one of the signals because the `signal_` member pointer of the signal has not been set (is still a `nullptr`). For example, here:

```sh
#0  0x00007ffff7e22f9a in GridKit::PhasorDynamics::SignalNode<double, unsigned long>::init (this=0x5555556c7030, signal=0.50000000000000022)
    at /home/4pf/repos/GridKit/src/Model/PhasorDynamics/SignalNode/SignalNodeImpl.hpp:70
#1  0x00007ffff7e80ada in GridKit::PhasorDynamics::ComponentSignals<double, unsigned long, GridKit::PhasorDynamics::GenrouInternalVariables, GridKit::PhasorDynamics::GenrouExternalVariables>::writeExternalVariable<(GridKit::PhasorDynamics::GenrouExternalVariables)2> (this=0x5555556c8eb0, value=0.50000000000000022)
    at /home/4pf/repos/GridKit/src/Model/PhasorDynamics/ComponentSignals.hpp:150
#2  0x00007ffff7e80845 in GridKit::PhasorDynamics::Genrou<double, unsigned long>::initialize (this=0x5555556c8c30)
    at /home/4pf/repos/GridKit/src/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp:359
#3  0x0000555555642ee2 in GridKit::PhasorDynamics::SystemModel<double, unsigned long>::initialize (this=0x7fffffffbed8)
    at /home/4pf/repos/GridKit/src/Model/PhasorDynamics/SystemModel.hpp:380
#4  0x00007ffff7fa6bfe in AnalysisManager::Sundials::Ida<double, unsigned long>::getDefaultInitialCondition (this=0x7fffffffbdf8)
    at /home/4pf/repos/GridKit/src/Solver/Dynamic/Ida.cpp:229
#5  0x00007ffff7fa692d in AnalysisManager::Sundials::Ida<double, unsigned long>::configureSimulation (this=0x7fffffffbdf8)
    at /home/4pf/repos/GridKit/src/Solver/Dynamic/Ida.cpp:66
```

PhilipFackler:
## Description
 
We want to make sure the system model components are set up correctly. Signal nodes that will be read or written to need to be verified at initialization.
 
 _Closes #266_
 
 ## Proposed changes
 
Add `verify()` function to `Component` class. Implement to check if `SignalNode`s are linked in classes that read or write to them.
 
 ## Checklist
 
- [X] All tests pass.


tests/UnitTests/PhasorDynamics/SystemModelTests.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+      virtual ~SystemModelTests()
+      {
+      }
+
+      /**
+       * @brief Test for exception thrown when signals are incorrectly
+       * configured
+       */
+      TestOutcome signalError()
```
nkoukpaizan: Could this be added to [SystemTests.hpp](https://github.com/ORNL/GridKit/blob/PhilipFackler/verify-signal-nodes/tests/UnitTests/PhasorDynamics/SystemTests.hpp), or do they have to be separate?

PhilipFackler: Happy to do that

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -303,6 +306,38 @@ namespace GridKit
...
+      static constexpr auto EFD = GenrouExternalVariables::EFD;
+
+      int ret = 0;
+
+      if (signals_.template isAttached<PM>())
+      {
+        if (!signals_.template isLinked<PM>())
+        {
+          Log::error() << "Genrou: pmech signal attached with no linked governor\n";
+          ret += 1;
```
superwhiskers, line 325: i'm dubious of the value of incrementing this return value type. i saw this pattern in resolve too and i saw it cause more problems than it solved. i think just returning "1" on error or having meaningfully defined error codes is more useful.

GridKit/Model/PhasorDynamics/SystemModel.hpp
```diff
@@ -326,9 +328,35 @@ namespace GridKit
...
+        int ret = 0;
+
+        // Verify components
+        for (const auto& component : components_)
+        {
+          ret += component->verify();
+        }
+
+        return ret;
+      }
```
superwhiskers, line 358: likewise here. well-defined error codes are more useful than just accumulating errors.

pelesh: I think the point here is to loop through all components and verify them and let each component report an error, if applicable. The sum will give an information how many components are incorrectly connected. In this case accumulating the error makes sense to me. I agree on Re::Solve comment, though.

superwhiskers: it just doesn't seem super useful. if we still want to count the number of incorrectly connected components then we can do that but i'd rather have a specific error type made that indicates what it is doing (counting the number of incorrectly connected components) than an undifferentiated integer.

pelesh: My understanding is that each component will report a specific error. The system model just loops over all components and let each one report, if something is wrong.

You are absolutely right that we need a more meaningful error codes but this is beyond the scope of this PR.

pelesh:
I am getting linking error when building this branch:
```
[ 22%] Linking CXX shared library libgridkit_phasor_dynamics_genrou_dependency_tracking.dylib
  "GridKit::Utilities::Logger::error()", referenced from:
      GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, long>::verify() const in GenrouDependencyTracking.cpp.o
      GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, long>::verify() const in GenrouDependencyTracking.cpp.o
      GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, unsigned long>::verify() const in GenrouDependencyTracking.cpp.o
      GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, unsigned long>::verify() const in GenrouDependencyTracking.cpp.o
collect2: error: ld returned 1 exit status
make[2]: *** [GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/libgridkit_phasor_dynamics_genrou_dependency_tracking.0.1.0.dylib] Error 1
make[1]: *** [GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/CMakeFiles/phasor_dynamics_genrou_dependency_tracking.dir/all] Error 2
make: *** [all] Error 2
```

nkoukpaizan:
> I am getting linking error when building this branch:
> 
> ```
> [ 22%] Linking CXX shared library libgridkit_phasor_dynamics_genrou_dependency_tracking.dylib
> Undefined symbols for architecture arm64:
>   "GridKit::Utilities::Logger::error()", referenced from:
>       GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, long>::verify() const in GenrouDependencyTracking.cpp.o
>       GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, long>::verify() const in GenrouDependencyTracking.cpp.o
>       GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, unsigned long>::verify() const in GenrouDependencyTracking.cpp.o
>       GridKit::PhasorDynamics::Genrou<GridKit::DependencyTracking::Variable, unsigned long>::verify() const in GenrouDependencyTracking.cpp.o
> ld: symbol(s) not found for architecture arm64
> collect2: error: ld returned 1 exit status
> make[2]: *** [GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/libgridkit_phasor_dynamics_genrou_dependency_tracking.0.1.0.dylib] Error 1
> make[1]: *** [GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/CMakeFiles/phasor_dynamics_genrou_dependency_tracking.dir/all] Error 2
> make: *** [all] Error 2
> ```

I am able to reproduce this on my macOS. @PhilipFackler Some of the libraries are missing the new `PUBLIC GridKit::phasor_dynamics_core` link.

GridKit/Model/PhasorDynamics/SystemModel.hpp
PhilipFackler: I was trying to follow the convention of the existing code (for `allocate` etc.), but I'm happy to make a change however you all prefer.

PhilipFackler:
Yep, not sure why it didn't complain on my linux build

GridKit/Model/PhasorDynamics/SystemModel.hpp
pelesh: I suggest we create a separate issue on error management in the code and discuss it there. We all agree that the current convention is a make-shift solution that needs to be replaced with a more sustainable one.

PhilipFackler:
@pelesh @nkoukpaizan @superwhiskers I'd like to hear your thoughts on the error messages I introduced here:
1. Am I using the correct terms?
2. How can they be more helpful to actual users?

nkoukpaizan:
> @pelesh @nkoukpaizan @superwhiskers I'd like to hear your thoughts on the error messages I introduced here:
> 
> 1. Am I using the correct terms?
> 2. How can they be more helpful to actual users?

The messages make sense to me.

pelesh:
> @pelesh @nkoukpaizan @superwhiskers I'd like to hear your thoughts on the error messages I introduced here:
> 
> 1. Am I using the correct terms?
> 2. How can they be more helpful to actual users?

I think the terms and wording are okay.

GridKit/Model/PhasorDynamics/SystemModel.hpp
PhilipFackler: It's unclear how I should resolve this (and the above) comment.

pelesh (approved):
#266 seems to be addressed correctly here.

nkoukpaizan (approved):
Some of the changes here triggered some warnings that are out of the scope of this PR See #304. Otherwise, looks good to me.

PR #305, opened 2025-11-07
Store Jacobian values as real_type rather than ScalarT

### LINKED ISSUE #303: ScalarT in Jacobian computations needs to be changed to real_type
Currently, all Jacobians are constructed using `ScalarT`, but this should be changed to `real_type`. When models are simulated using `DependencyTracking::Variable`, we don't want the dependency tracking variables inside the actual Jacobians, and this is causing a mis-compilation in power electronics models when using `DependencyTracking::Variable` to compare sparsity patterns for #301.

### LINKED ISSUE #306: Make type definitions consistent across the code
As @pelesh noted in #305, we currently use both `RealT` and `real_type` in the code. It would be good to decide on `...T` versus `_type`, make these consistent across the code and specify the desired choice in the contributing guidelines. Perhaps we can also clarify our preference for `typedef` versus `using`.

This should be a simple search and replace exercise, but it will touch most of the files.

nkoukpaizan:
## Description
 
This add support for `DependencyTracking::Variable` in `PowerElectronics` models and Jacobian value storage as `RealT` rather than `ScalarT`.
Closes #303 
Closes #306 
 
 ## Proposed changes
- Templated matrix classes over `RealT` rather than `ScalarT`
- Added a conversion operator to the `DependencyTracking::Variable` class to support `static_cast<RealT>` (i.e., `static_cast<double>`)
- Updated the Jacobian evaluation in the `PowerElectronics` models to store `RealT` values, leveraging the conversion from `ScalarT==DependencyTracking::Variable` to `RealT` where necessary
- Added instantiations of `PowerElectronics` models over `DependencyTracking::Variable`
- Also took the opportunity to cleanup type declarations and template parameters.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 I've considered the alternative of keeping the matrices templated over `ScalarT` and using the `ScalarTraits` to store the data as `RealT`, but I think templating over `RealT` directly is clearer.


pelesh:
I suggest we decide `RealT` or `real_type`? 😄

nkoukpaizan:
> I suggest we decide `RealT` or `real_type`? 😄

In my mind, I was using `RealT` for the template parameter and `real_type` for the local variable types, typically obtained from `ScalarTraits`. But I agree this isn't consistently applied in the code. Let me take a closer look.

GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp
```diff
@@ -254,6 +254,12 @@ namespace GridKit
       inline Variable& operator/=(const double& rhs);
       inline Variable& operator/=(const Variable& rhs);
 
+      // conversion operator
+      inline operator double() const
+      {
+        return value_;
+      }
```
pelesh: Why not call the function `value()`?

```suggestion
      inline operator value() const
      {
        return value_;
      }
```
It will not necessarily return a double floating precision type.

GridKit/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
```diff
@@ -314,6 +315,8 @@ namespace GridKit
       template <typename ModelT, MemberFunctions function, class ScalarT, typename IdxT>
       struct InternalJacobian
       {
+        typedef typename GridKit::ScalarTraits<ScalarT>::real_type real_type;
+
```
pelesh: How about:
```suggestion
        using RealT = typename GridKit::ScalarTraits<ScalarT>::RealT;

```
This also assumes we replace `real_type` with `RealT` across the board.

GridKit/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
```diff
@@ -325,15 +328,15 @@ namespace GridKit
...
+        static void eval(ModelT*                                              model,
+                         size_t                                               n_res,
+                         size_t                                               n_var,
+                         const std::map<IdxT, IdxT>&                          res_indices,
+                         const std::map<IdxT, IdxT>&                          var_indices,
+                         ScalarT*                                             y,
+                         ScalarT*                                             yp,
+                         ScalarT*                                             wb,
+                         GridKit::LinearAlgebra::COO_Matrix<real_type, IdxT>& jac)
         {
```
pelesh: Is ti possible to replace 
```suggestion
                         MatrixT<real_type, IdxT>& jac)
{
```
where we define `MatrixT` witha `using` statement?

GridKit/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
pelesh: Similar at a few other places in this file.

GridKit/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
pelesh: Similar at a few other places in this file.

GridKit/LinearAlgebra/DenseMatrix/DenseMatrix.hpp
```diff
@@ -52,14 +52,14 @@ namespace GridKit
     /**
      * @brief DenseMatrix constructor
      *
-     * @tparam ScalarT
+     * @tparam RealT
      * @tparam IdxT
```
pelesh: Perhaps explain what these data types are used for:
```suggestion
     * @tparam RealT - Real type for Jacobian entries
    * @tparam IdxT - Integer data type for matrix indices
```

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -275,7 +275,7 @@ namespace GridKit
 
         ScalarT ind_low  = (this->sigmoid(Vrmin_ - x)) * (this->sigmoid(-f));
         ScalarT ind_high = (this->sigmoid(x - Vrmax_)) * (this->sigmoid(f));
-        return (1 - ind_low) * (1 - ind_high);
+        return (1.0 - ind_low) * (1.0 - ind_high);
       }
```
pelesh: We do need to define constants for GridKit. This should be something like
```suggestion
        return (ONE<RealT> - ind_low) * (ONE<RealT> - ind_high);
}
```
or

```suggestion
        auto one = ONE<RealT>;
        return (one - ind_low) * (one - ind_high);
}
```
Probably out of scope for this PR.

pelesh: Same holds for a few other places in the code.

pelesh (commented):
A few nitpicking comments:

GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp
alexander-novo: This is a casting operator - it must be called the name of the type it returns. It is necessary for the `static_cast` used elsewhere.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
nkoukpaizan: Happy to do it here, since we keep putting it off.

GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp
pelesh: Thanks for clarifying, @alexander-novo.

GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp
```diff
@@ -254,6 +254,12 @@ namespace GridKit
...
+
```
alexander-novo, line 262:
```suggestion
      // conversion operator
      explicit inline operator double() const
      {
        return value_;
      }
```
Since you're using `static_cast` elsewhere, maybe it's a good idea to mark this operator as `explicit`, that way we can avoid a bug that looks something like this:

```cpp
real_type x = y[0];
...
f[1] = x * y[1];
```

alexander-novo (commented):
Some model parameters in `PowerElectronics` e.g.

https://github.com/ORNL/GridKit/blob/7293eaedf0d59a647ec5e797530ee98dd702b2d4/GridKit/Model/PowerElectronics/MicrogridLoad/MicrogridLoad.hpp#L61-L64

still need to be changed to `real_type`.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
nkoukpaizan: Added a `Constants.hpp` file in [135e924](https://github.com/ORNL/GridKit/pull/305/commits/135e9242e7d59dc173a6545928e87117fc13ecae). Will need another PR to search and replace constants in the files that have not been touched by this PR.

GridKit/AutomaticDifferentiation/Enzyme/SparseWrapper.hpp
nkoukpaizan: Addressed in [244630c](https://github.com/ORNL/GridKit/pull/305/commits/244630c2f3899029955fc7881e2bf98784f1b758)

GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp
nkoukpaizan: Addressed in [427d82c](https://github.com/ORNL/GridKit/pull/305/commits/427d82cc5cc499e65d170c3c251cd3e1244bf08e)

nkoukpaizan:
> Some model parameters in `PowerElectronics` e.g.
> 
> https://github.com/ORNL/GridKit/blob/7293eaedf0d59a647ec5e797530ee98dd702b2d4/GridKit/Model/PowerElectronics/MicrogridLoad/MicrogridLoad.hpp#L61-L64
> 
> still need to be changed to `real_type`.

Addressed in [7b21b3f](https://github.com/ORNL/GridKit/pull/305/commits/7b21b3f4cd3a003d51a6ad4d68212bef2728b44e). I hope I got them all this time.

GridKit/LinearAlgebra/DenseMatrix/DenseMatrix.hpp
nkoukpaizan: Addressed in [d6eaba1](https://github.com/ORNL/GridKit/pull/305/commits/d6eaba1f269405adfe00b62a08ed2f3fa9ea2721)

alexander-novo (approved):
Allows unit tests for CSR sparsity pattern to compile

pelesh (approved):
Great work!

PR #307, opened 2025-11-11
Add Power Electronics sparsity pattern computation into allocate

### LINKED ISSUE #301: Implement analysis in `SystemModelPowerElectronics` class to compute system Jacobian sparsity pattern
### Summary

The system model class `SystemModelPowerElectronics` receives as the input system components and their connectivity information. Implement functions that compute number of nonzeros and matrix indices in compressed sparse row (CSR) format for the system Jacobian from the component connectivity information.

### Rationale

The key step in assembling the system is to compute number of nonzero and to obtain sparsity pattern for the system Jacobian, in our case in CSR format. Once this information is obtained, system Jacobian matrix can be instantiated and allocated.

### Description

Implementation is likely to require two loops through components and buses (starting with a loop through buses). The first loop will count number of nonzeros and the starts of matrix rows. With that information, the second loop over all components will collect all column indices in matrix rows. Column indices might need to be sorted in each row, depending on implementation details.

### Additional information

Mapping component Jacobian indices to system Jacobian indices is needed but not necessarily in the scope of this issue.

CC @nkoukpaizan

alexander-novo:
## Description
 
Adds sparsity pattern computation into `PowerElectronicsModel::allocate()`.
 
 Closes #301
 
 ## Checklist
 
 _Put an `x` in the boxes that apply. You can also fill these out after creating
 the PR. If you're unsure about any of them, don't hesitate to ask. We're here
 to help! This is simply a reminder of what we are going to look for before
 merging your code._
 
- [x] All tests pass.
 
 ## Further comments
 
I added a couple of helper structs, but these are only there for cache locality and readability purposes. They are local to the function itself. Additionally, I made use of `std::vector` as a dynamically growing container, but all final products end up as simple pointers.

### DIFF
--- a/GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
+++ b/GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
@@ -64,6 +64,8 @@ namespace GridKit
     using component_type = CircuitComponent<ScalarT, IdxT>;
 
     using CircuitComponent<ScalarT, IdxT>::size_;
+    using CircuitComponent<ScalarT, IdxT>::n_intern_;
+    using CircuitComponent<ScalarT, IdxT>::n_extern_;
     using CircuitComponent<ScalarT, IdxT>::nnz_;
     using CircuitComponent<ScalarT, IdxT>::time_;
     using CircuitComponent<ScalarT, IdxT>::alpha_;
@@ -291,10 +293,49 @@ namespace GridKit
     int allocate(IdxT s)
     {
       // Allocate all components
-      size_ = s;
+      size_     = s;
+      n_intern_ = 0;
       for (const auto& component : components_)
       {
         component->allocate();
+
+        // Count up the amount of internal variables which get mapped to system variables
+        auto extern_indices = component->getExternIndices();
+        for (IdxT comp_var_idx = 0; comp_var_idx < component->size(); comp_var_idx++)
+        {
+          IdxT sys_var_idx = component->getNodeConnection(comp_var_idx);
+          if (!extern_indices.contains(comp_var_idx) && sys_var_idx != neg1_)
+          {
+            n_intern_++;
+          }
+        }
+      }
+      n_extern_ = size_ - n_intern_;
+
+      // Ensure that all component locals are mapped to system locals
+      // and all component externals are mapped to system externals.
+      // System locals are stored first, in 0..n_intern_ and externals
+      // are stored after, in n_intern_..
+      for (auto component : components_)
+      {
+        auto extern_indices = component->getExternIndices();
+        for (IdxT comp_var_idx = 0; comp_var_idx < component->size(); comp_var_idx++)
+        {
+          IdxT sys_var_idx = component->getNodeConnection(comp_var_idx);
+
+          // Ignore local variables which aren't mapped to sytem variables
+          if (sys_var_idx == neg1_)
+            continue;
+
+          if (extern_indices.contains(comp_var_idx))
+          {
+            assert(sys_var_idx >= n_intern_);
+          }
+          else
+          {
+            assert(sys_var_idx < n_intern_);
+          }
+        }
       }
 
       // Allocate global vectors
--- a/GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
+++ b/GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
@@ -81,11 +81,13 @@ namespace GridKit
     PowerElectronicsModel()
     {
       // Set system model parameters as default
-      rel_tol_         = 1e-4;
-      abs_tol_         = 1e-4;
-      this->max_steps_ = 2000;
+      rel_tol_                 = 1e-4;
+      abs_tol_                 = 1e-4;
+      this->max_steps_         = 2000;
       // By default don't use the jacobian
-      use_jac_         = false;
+      use_jac_                 = false;
+      jac_sparsity_row_indices = nullptr;
+      jac_sparsity_col_indices = nullptr;
     }
 
     /**
@@ -104,11 +106,13 @@ namespace GridKit
                           IdxT   max_steps = 2000)
     {
       // Set system model tolerances from input
-      rel_tol_         = rel_tol;
-      abs_tol_         = abs_tol;
-      this->max_steps_ = max_steps;
+      rel_tol_                 = rel_tol;
+      abs_tol_                 = abs_tol;
+      this->max_steps_         = max_steps;
       // Can choose if to use jacobian
-      use_jac_         = use_jac;
+      use_jac_                 = use_jac;
+      jac_sparsity_row_indices = nullptr;
+      jac_sparsity_col_indices = nullptr;
     }
 
     /**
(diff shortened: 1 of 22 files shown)


alexander-novo:
I also wanted to add an additional comment on the removal of the short-circuiting condition in `COO_Matrix::axpy()`:

I found out yesterday that IDA sets the `alpha` parameter to 0 during `IDACalcIC()`, which was causing the sparsity pattern to change. This (for some unknown reason) was what was causing KLU to segmentation fault. So this change also fixes the issue that @abdourahmanbarry noticed in his fault simulation.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -81,11 +81,13 @@ namespace GridKit
     PowerElectronicsModel()
     {
       // Set system model parameters as default
-      rel_tol_         = 1e-4;
-      abs_tol_         = 1e-4;
-      this->max_steps_ = 2000;
+      rel_tol_                 = 1e-4;
+      abs_tol_                 = 1e-4;
+      this->max_steps_         = 2000;
       // By default don't use the jacobian
```
pelesh:
```suggestion
      max_steps_         = 2000;
// By default don't use the jacobian
```
No need to use `this->...` with the member variable convention we use consistently in the code.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -123,18 +127,140 @@ namespace GridKit
...
+        // The local system variable index
+        size_t local_row_idx_;
+      };
+
+      // Helper struct for saving CSR sparsities of a particular component, since
+      // right now they only report COO. Can be removed once components have CSR matrices.
+      struct CsrSparsity
+      {
+        std::vector<IdxT> row_indices_;
+        std::vector<IdxT> col_indices_;
```
pelesh: I would avoid using `std::vector` here. We created CSR matrix class to eliminate need for STL containers. We can use the CSR matrix here instead.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -123,18 +127,140 @@ namespace GridKit
...
+      };
+
+      // A reverse mapping from system variables -> component variables
+      std::vector<std::vector<ComponentContribution>> reverse_map(size_);
+      std::vector<CsrSparsity>                        component_sparsities;
```
pelesh: I would avoid using STL containers here. What we need to do is to
- populate row pointers and column indices in the CSR Jacobian matrix
- have components store mapping from their Jacobian elements to the index of column indices array

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -123,18 +127,140 @@ namespace GridKit
...
     int allocate()
     {
+      // Helper struct for identifying a particular component's local system variable
+      struct ComponentContribution
+      {
+        // The index of the component in `components_`.
+        size_t comp_idx_;
+        // The local system variable index
+        size_t local_row_idx_;
+      };
```
pelesh, line 157: I am not sure this is a good idea. Ideally, one should loop through components and components would read/write at locations determined by the connectivity information. The system composer should remain agnostic of specific details of the components.

pelesh (changes requested):
Great progress. Some design tweaks are needed but we are almost there..

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
alexander-novo: The local components need to start using CSR matrices, then, which was not part of this PR. As commented, this can be removed once components use CSR matrices

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
alexander-novo: I cannot really remove the STL container for `reverse_map`... the size of the second level arrays are unknown ahead of time and need to be dynamically grown. I can remove the vector for `component_sparsities`, though

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
alexander-novo: Well a component cannot write all of its info at once, so I must have some way of differentiating the different parts of info that a component needs to write at once.

alexander-novo:
rebase w.r.t. develop

alexander-novo:
I changed the sparsity to be calculated into `CsrMatrix` object, but because of the aforementioned issue with the assignment operators' implementations in `CsrMatrix`, there is a use-after-free bug.

@nkoukpaizan maybe it's worth fixing? I can do some wonky stuff to try and avoid it but it seems to me like maybe it's better to just fix the bug?

nkoukpaizan:
> I changed the sparsity to be calculated into `CsrMatrix` object, but because of the aforementioned issue with the assignment operators' implementations in `CsrMatrix`, there is a use-after-free bug.
> 
> @nkoukpaizan maybe it's worth fixing? I can do some wonky stuff to try and avoid it but it seems to me like maybe it's better to just fix the bug?

Wouldn't it work if we stored `csr_jac_` as a pointer, using `new` and deallocating in the system model's destructor? My understanding is that we don't want to use the assignment operator.

alexander-novo:
> > I changed the sparsity to be calculated into `CsrMatrix` object, but because of the aforementioned issue with the assignment operators' implementations in `CsrMatrix`, there is a use-after-free bug.
> > @nkoukpaizan maybe it's worth fixing? I can do some wonky stuff to try and avoid it but it seems to me like maybe it's better to just fix the bug?
> 
> Wouldn't it work if we stored `csr_jac_` as a pointer, using `new` and deallocating in the system model's destructor? My understanding is that we don't want to use the assignment operator.

That's the wonky stuff I meant. Is `CsrMatrix` only ever meant to exist on the heap, then? That should probably be mentioned somewhere.

pelesh:
> 
> Wouldn't it work if we stored `csr_jac_` as a pointer, using `new` and deallocating in the system model's destructor? My understanding is that we don't want to use the assignment operator.

I don't see are need to use assignment operator either (and it might be a good idea to delete it tbh). The implementation should go something like this:
1. In constructor `csr_jac_` is instantiated.
2. In `allocate` function, NNZ is computed. After that, the Jacobian is allocated with the computed NNZ.
3. Next (also in `allocate`), the sparsity pattern of the Jacobian is computed.

This is what is needed to address #301.

alexander-novo:
> > Wouldn't it work if we stored `csr_jac_` as a pointer, using `new` and deallocating in the system model's destructor? My understanding is that we don't want to use the assignment operator.
> 
> I don't see are need to use assignment operator either (and it might be a good idea to delete it tbh). The implementation should go something like this:
> 
>     1. In constructor `css_jac_` is instantiated.
> 
>     2. In `allocate` function, NNZ is computed. After that, the Jacobian is allocated with the computed NNZ.
> 
>     3. Next (also in `allocate`), the sparsity pattern of the Jacobian is computed.
> 
> 
> This is what is needed to address #301.

But as far as I can tell, there is no way to change the size of the matrix after construction. And the size of the system is not known at construction time for `PowerElectronicsModel`

pelesh:
> 
> That's the wonky stuff I meant. Is `CsrMatrix` only ever meant to exist on the heap, then? That should probably be mentioned somewhere.

You could also have

```c++
class SystemSolverPowerElectronics
{
public:
  // some code

private:
  CsrMatrix jacobian_;
};
```
In that case Jacobian matrix would be instantiated with a default constructor and you would allocate it and populate its data arrays. That wouldn't change much in the approach.

pelesh:
> 
> But as far as I can tell, there is no way to change the size of the matrix after construction. And the size of the system is not known at construction time for `PowerElectronicsModel`

You implemented `CsrMatrix` class, so feel free to add methods that may be needed for the system Jacobian assembly. 😄 Adding a `resize` method would make sense to me in this case.

alexander-novo:
I copied the implementation from ReSolve, I'm not really familiar with how it works. It kind of seemed to me like there may have been a good reason why there wasn't a `resize()` method. Would a `resize()` method that simply sets `m,n` be correct?

alexander-novo:
Well, as far as I can tell `m_` is used for nothing and `n_` is just used for allocating memory, so I think it should probably be harmless to just modify them. Please let me know.

tests/UnitTests/PowerElectronics/SystemTests.hpp
```diff
@@ -0,0 +1,301 @@
...
+        double rload2 = 2.0;
+        double Lload2 = 1.0 / (2.0 * M_PI * 50.0);
+
+        std::vector<double> rload_list(Nsize, rload2);
+        std::vector<double> Lload_list(Nsize, Lload2);
+        rload_list[0] = rload1;
+        Lload_list[0] = Lload1;
+
+        //							DGs	+		- refframe	   Lines +				Loads
+        size_t vec_size_internals = 13 * (2 * Nsize) - 1 + (2 + 4 * (Nsize - 1)) + 2 * Nsize;
```
nkoukpaizan: We really need a better way to set these values for `PowerElectronics`, especially if we are going to move this code from `examples` to `tests`. 

For now, perhaps define, e.g., `dg_size = 13`, etc. to reuse throughout the test. 

I don't know what to do about the user-facing `setExternalConnectionNodes` with index arithmetic, as `PowerElectronics` does not use buses.

alexander-novo: I agree with all of this. This was why I was asking for functions that setup test systems a while back, so I can can call e.g. `microgrid_system(NSize)` in both the example and the test, rather than duplicating the code. So that for this unit test I don't need to change the modeling, but when the modeling does get changed in the future the unit test will automatically update.

alexander-novo: I will say, I do think components *do* keep track of how many of their variables are internal, so I can remove some of this redundant calculation.

pelesh: > I don't know what to do about the user-facing `setExternalConnectionNodes` with index arithmetic, as `PowerElectronics` does not use buses.

I believe that the `PowerElectronics` module does use buses (nodes), it just not discriminate them. In general, buses should be treated differently because they (should) store coupling variables and residuals of the coupling equations. They are also natural partitioning points  of the system. In Jacobian assembly, derivatives of component internal residuals should be placed in top matrix rows and derivatives of coupling equations at the bottom to get favorable block-arrow structure.

nkoukpaizan: I'm not going to ague on different interpretations of "use". My point is that the system model currently has no information on what is connected to what. The local component indices are mapped to the system indices from outside by calling the component's `setExternalConnectionNodes`, which is why we're doing this index arithmetic here. The bus variables and residuals equations are eliminated.

pelesh: I agree and I think this is a flaw in the `PowerElectronics` module.

alexander-novo: > > I don't know what to do about the user-facing `setExternalConnectionNodes` with index arithmetic, as `PowerElectronics` does not use buses.
> 
> I believe that the `PowerElectronics` module does use buses (nodes), it just not discriminate them. In general, buses should be treated differently because they (should) store coupling variables and residuals of the coupling equations. They are also natural partitioning points of the system. In Jacobian assembly, derivatives of component internal residuals should be placed in top matrix rows and derivatives of coupling equations at the bottom to get favorable block-arrow structure.

I think I can actually get this correctly, let me see.

alexander-novo:
Okay, I went ahead an imposed the correct system vector ordering. The `Microgrid` examples already have this ordering, but I did have to change the `RLCircuit` example to have the correct ordering.

Now that the ordering is required to be nice, I went ahead and updated the jacobian sparsity computations to take advantage of it - first the internal rows are done component-by-component, then the heavier reverse-mapping machinery is invoked only for the external rows.

I also realized that I could use the sum of the component Jacobian `nnz`s as an upper bound for the system Jacobian `nnz` value and got rid of the vector for `global_col_indices` by just pre-allocating this upper bound. I don't think that this will be _too_ big an over-estimate to be concerned about the wasted memory.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -121,20 +126,177 @@ namespace GridKit
...
+            {
+              global_col_indices[global_row_indices[row + 1]] = global_col;
+              global_row_indices[row + 1]++;
+            }
+          }
+        }
+
+        // Sort the row by column indices. Since the mapping from local indices to global indices isn't monotonically increasing,
+        // this is necessary.
+        auto start = std::next(global_col_indices, static_cast<ptrdiff_t>(global_row_indices[row]));
```
nkoukpaizan: I added a `static_cast<ptrdiff_t` to get around sign warnings from Clang. @alexander-novo can we avoid this iterator altogether?

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
nkoukpaizan: This whole section will have to change once we remove `COO_Matrix` class altogether. I'm not opposed to keeping it until the next stage of the work.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
alexander-novo: What's your goal?

Do you just want to get rid of `std::next`? I have it formatted like an iterator, but in the end, this is just pointer arithmetic, so it can just be done as pointer addition.

alexander-novo:
@pelesh I implemented what we discussed at the SciDAC meeting, and am not depending on the components having their jacobians in CSR form anymore (just using COO).

The two outstanding pieces of feedback, though, I'm waiting for your feedback on. Maybe we can talk about them in a code review?

alexander-novo:
Addressed some feedback from the code review:

- Removed `auto` keyword usage
- Switched to using the "stealing" constructors in `CsrMatrix` by putting `csr_jac_` on the heap.

pelesh:
Addressed in #335 and #361

Issue #96, opened 2025-05-06
Bug: Components which are part of a system are not updated by the integrator after the final time step


alexander-novo:
### Description how to reproduce the bug

If the example given in [`examples/PhasorDynamics/Example1/example1.cpp`](https://github.com/ORNL/GridKit/blob/0e07a2638bf3606f6320528137269bd11922e57d/examples/PhasorDynamics/Example1/example1.cpp#L96-L104) is modified to compare values from the individual components to the reference solution rather than the system values, the test will fail with error too large. This is because
- The integrator only updates the main `Evaluator` it has been given - in this case the system.
- `System` evaluators only update their child components/buses when the integrator evaluates its residual, which won't happen after the final time step.

Therefore each component is one integrator internal time step out of date from the system.

### GridKit™ version

0e07a2638bf3606f6320528137269bd11922e57d

### System and environment details

N/A

### Additional information

It would be nice to be able to query final solution values from the individual components because this is more meaningful than querying from the system:
- If the order we store variables in the "global" system vector is changed for some reason (which I don't think we're guaranteeing against), this example will break
- Components often offer meaningful querying functions which mean we don't have to access the state vector manually

PR #308, opened 2025-11-14
Add variable monitor for solver iteration output

PhilipFackler:
## Description
 
Capability for printing monitored variables during time iterations
 
 ## Proposed changes
 
Let each bus and device specify variables to monitor based on `*Data` input. During `Ida::runSimulation` allow for monitoring output as well as user-defined `step_callback`.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
I'm starting this as a draft because I need some feedback before I can consider it completed.


.clang-format
```diff
@@ -49,6 +51,7 @@ PackConstructorInitializers: NextLine
 AllowShortFunctionsOnASingleLine: None
 BreakBeforeBinaryOperators: NonAssignment # Put binary operators after a line break, rather than before
 AllowShortIfStatementsOnASingleLine: Never
+AllowShortLambdasOnASingleLine: Inline
```
PhilipFackler: I'm allowing short lambdas to be all on one line.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -435,6 +441,25 @@ namespace GridKit
...
+      void Ieeet1<ScalarT, IdxT>::printMonitoredVariables(std::ostream& os) const
+      {
+        monitor_.print(os);
+      }
+
+      template <class ScalarT, typename IdxT>
+      void Ieeet1<ScalarT, IdxT>::initializeMonitor()
+      {
+        using Variable = model_data_type::MonitorableVariables;
+        monitor_.set(Variable::efd, [this] { return efd_signal_->read(); });
```
PhilipFackler: This seems wrong since one of the constructors takes a `data` parameter which might specify to monitor `efd` but does not supply the `efd_signal`.

GridKit/Model/VariableMonitor.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+      bool empty() const
+      {
+        return variables_.empty();
+      }
+
+      void push(VariableEnum v, ValueFunction g)
+      {
+        // variables_.emplace_back(v, g);
+      }
```
PhilipFackler: Need to remove this

GridKit/Model/VariableMonitor.hpp
```diff
@@ -0,0 +1,104 @@
...
+        if (empty())
+        {
+          return;
+        }
+        os << indent_ << label_ << ":\n";
+        for (auto v : variables_)
+        {
+          print(os, v);
+        }
+      }
```
PhilipFackler: This produces output (for ThreeBusBasicJson for example) like this:
```sh
t: 9.9875:
  Bus 1:
    Vr: 1.06
    Vi: 0
  Bus 2:
    Vr: 1.05983
    Vi: -0.0096619
  Bus 3:
    Vr: 0.961032
    Vi: -0.131193
  Genrou genrou_2_1:
    Delta: 0.17193
    Omega: -2.29047e-07
  Genrou genrou_3_1:
    Delta: -0.0197693
    Omega: -8.28322e-08
```
I'm curious how you all would recommend formatting this. I'm thinking scientific notation with a consistent number of digits. I put each value on its own line assuming we might be printing `DependencyTracking::Variable`s.

lukelowry: My immediate thoughts: Although this is readable, a more useful format (as I debug/verify the models) is something that is easily transfered to Excel (or an equivilent plotting tool). How difficult would it be to format the output as a 'column' per field and 'row' per time step so that the console output is well conditioned to be pasted into Excel for plotting? You could write the field names at the start so it is not so verbose.

PhilipFackler: We could do both...default to printing nicely on the console, and also have an option to print CSV style to a file. This would just be an extra top-level option in the JSON file.

pelesh: I agree with @lukelowry, I think a table format is more suitable for this purpose than a key-value approach. A simple csv format with time as the first column should be appropriate.

pelesh: Scientific notation with full machine precision is likely the best output.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
pelesh: I believe that `Ieeet1` is the source of the `efd` signal. Perhaps we need to revisit the constructor implementation?

.clang-format
```diff
@@ -38,7 +38,9 @@ AlignConsecutiveMacros:
 
 # Newlines
 ColumnLimit:     0
-BreakBeforeBraces: Allman         # Braces on their own lines
```
pelesh, line 41: Please keep Allman braces.

superwhiskers (changes requested):
this pull request leaves casing of enums inconsistent in data structures. for instance, the newly added monitorable variables in busdata are all caps, inconsistent with enums in other files. this should be resolved prior to merge.

additionally, we should evaluate whether we are willing to use magic_enum to sidestep the requirement of a `SIZE` variant, something i worked around with it previously.

GridKit/Model/VariableMonitor.hpp
superwhiskers: to avoid overcomplication, gridkit should emit json directly imo. i think an external tool for csv conversion would be sufficient to address that use case.

pelesh: > to avoid overcomplication, gridkit should emit json directly imo. i think an external tool for csv conversion would be sufficient to address that use case.

I second this. With a separate `json2csv` utility we make better separation of I/O formatters and the GridKit proper. We would want to abstract I/O formatting from the examples. In case we want to change I/O parsing, we don't want to have to change all examples.

PhilipFackler:
> this pull request leaves casing of enums inconsistent in data structures. for instance, the newly added monitorable variables in busdata are all caps, inconsistent with enums in other files. this should be resolved prior to merge.
> 
> additionally, we should evaluate whether we are willing to use magic_enum to sidestep the requirement of a `SIZE` variant, something i worked around with it previously.

Both good points. I'll check the guidelines for enum member styles. If we want to make magic_enum a public dependency that would help with the implementation here a lot. Or we could install it as part of gridkit so users don't have to deal with that. What do people think? I implemented this treating it purely as a build dependency.

GridKit/Model/VariableMonitor.hpp
```diff
@@ -0,0 +1,104 @@
...
+      {
+      };
+
+      template <typename TScalar, typename TIdx, template <typename, typename> typename TT>
+      struct Types<TT<TScalar, TIdx>>
+      {
+        using ScalarT = TScalar;
+        using IdxT    = TIdx;
+      };
+    } // namespace Detail
```
nkoukpaizan: I suggest keeping template typenames as `...T` rather than `T...`, granted you're operating on both here. I'm also not sure I understand the purpose of this `Detail::Types` indirection. If it is to support `DependencyTracking::Variable`, I think it's safe to assume we will only print `RealT` values through monitoring.

lukelowry (approved):
Other than my one concern regardining output format, this is an invaluable PR. Labeled output data will accelerate verification for the 37/39 bus case and larger cases. Having this in the develop branch would be incredibly valuable right now.

GridKit/Model/VariableMonitor.hpp
PhilipFackler: I'm hearing differing messages here. Can someone summarize what the approach should be?

.clang-format
PhilipFackler: I removed that because it doesn't allow for single-line lambdas. Perhaps we could customize?

GridKit/Model/VariableMonitor.hpp
PhilipFackler: This is to grab the `ScalarT` type from the Evaluator since you can't access that alias from `Evaluator`. To me it matters less what you name the parameters and more how you name the public interface members. I would change the template parameter name in `Evaluator` so that you could make a `ScalarT` alias available next to `RealT`. Anyway...if I use `RealT` instead of `ScalarT` for monitoring, wouldn't I have to have special code to handle `DependencyTracking::Variable`s? (to get the value?)

GridKit/Model/VariableMonitor.hpp
pelesh: Summary: 
- Output in scientific notation with full machine precision.
- An output format option should be CSV file with time as the first column.
- Consider emitting results in JSON and having external utility to convert JSON ---> CSV
    - If this is too distracting from the main objective of this PR, perhaps leave it for subsequent PR?
    - @superwhiskers, do you have more specific suggestions?

pelesh:
> > this pull request leaves casing of enums inconsistent in data structures. for instance, the newly added monitorable variables in busdata are all caps, inconsistent with enums in other files. this should be resolved prior to merge.
> > additionally, we should evaluate whether we are willing to use magic_enum to sidestep the requirement of a `SIZE` variant, something i worked around with it previously.
> 
> Both good points. I'll check the guidelines for enum member styles. If we want to make magic_enum a public dependency that would help with the implementation here a lot. Or we could install it as part of gridkit so users don't have to deal with that. What do people think? I implemented this treating it purely as a build dependency.

The GridKit code outside the I/O parser should not depend on magic_enum. I would try to avoid it being GridKit library dependency. It should be available only in utilities for building GridKit apps.

.clang-format
pelesh: I would advise against customizing; I don't think we want to open that can of worms again 😄. If the price we pay is not having single-line lambdas, so be it.

GridKit/Model/VariableMonitor.hpp
nkoukpaizan: The naming suggestion is more a matter of consistency. We've recently added it to the [contributing guidelines](https://github.com/ORNL/GridKit/blob/develop/CONTRIBUTING.md).

Yes, you'd have to convert `DependencyTracking::Variable` to a `RealT` (`double`) if monitoring is templated over `RealT`.

Odds are, others reading the code will not find this indirection straight forward. Perhaps add some comments in the code? I'd rather the monitoring class be explicitly templated over `ScalarT` if that parameter is really needed.

.clang-format
PhilipFackler: Very well then. 😄

GridKit/Model/VariableMonitor.hpp
PhilipFackler: Either file output is relatively easy. I'll do both. But how do you want the console output?

GridKit/Model/VariableMonitor.hpp
PhilipFackler: OK, two things here (the first one is more important since it affects this PR):

First, do we want to monitor `ScalarT` values or `RealT` values?

Second, I don't think I did a good job explaining above. I agree with wanting consistency, but there is an inconsistency already in that, given a type `PretendDevice` derived from `Evaluator`, I can use `typename PretendDevice::RealT` but not `PretendDevice::ScalarT`. This helper class was a simple way to work around that. A more thorough approach would be to decide on a consistent style for template parameter names and a  different style (since they can't be the same) for type aliases that form part of the interface, used either internally or between classes as in my example.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
PhilipFackler: I think someone who knows things should create a meaningful test case for this. But it can be a bug for another issue. :)

GridKit/Model/VariableMonitor.hpp
nkoukpaizan: To the first point, we want to monitor `RealT` values.

To the second point, we used to have different aliases for the template parameters and the local types (`real_type` vs `RealT`), but we changed it in #305. We want to avoid different names referring to the same type moving forward.

pelesh:
> > this pull request leaves casing of enums inconsistent in data structures. for instance, the newly added monitorable variables in busdata are all caps, inconsistent with enums in other files. this should be resolved prior to merge.
> > additionally, we should evaluate whether we are willing to use magic_enum to sidestep the requirement of a `SIZE` variant, something i worked around with it previously.
> 
> Both good points. I'll check the guidelines for enum member styles. If we want to make magic_enum a public dependency that would help with the implementation here a lot. Or we could install it as part of gridkit so users don't have to deal with that. What do people think? I implemented this treating it purely as a build dependency.

magic_enum is an external library over which we have little control. Making it required GridKit dependency is something I would prefer to avoid. I understand it's easier said than done, though.

PhilipFackler:
Just so y'all know, I'm planning to finish up requested changes and get this updated tomorrow.

GridKit/Model/VariableMonitor.hpp
PhilipFackler: I'm going to mark this as resolved since I changed the code significantly. Further suggestions can be done separately.

PhilipFackler:
I have added the capability of choosing the output format(s) using the JSON input file. By default, it will print CSV style to the console. I included changes to `ThreeBusBasic.json` as an example of how to use this. It prints three files (in addition to the console). If you want to change the console style, you just add a format group without the `file_name` parameter. Let me know what you all think. I'll add documentation, etc... next week. Happy Thanksgiving!

PhilipFackler:
Not sure why the consumer test is failing (it works locally for me). I'll have to dig into that later this week.

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -345,6 +352,14 @@ namespace AnalysisManager
...
       copyVec(yy_, model_->y());
       copyVec(yp_, model_->yp());
+      if (model_->monitoring())
+      {
+        model_->printMonitoredVariables();
+      }
+      if (step_callback.has_value())
+      {
+        (*step_callback)(tret);
+      }
```
PhilipFackler: @pelesh It seemed like the variables should updated one last time after this copy, but this breaks the tests in the examples. Am I right thinking this is a bug, or should I revert?

pelesh: I believe you are right, but I'd have to double-check.

@alexander-novo: I believe you came across similar issue earlier, could you please chime in?

alexander-novo: Yeah so this will only update the system's state, but there wasn't at the time a way to tell the system to update its components. So at this time in the program, the states of components and the system will be out of sync.

If `printMonitoredVariables()` is recursively observing state in components, then their state will be one step out of date, which is why tests may have been broken (tests previously observed the state of the system, not of components)

See #96

PhilipFackler: So, should I update the tests, or is `printMonitoredVariables()` doing the wrong thing?

pelesh: The tests were working fine before, so my guess would be that `printMonitoredVariables()` is doing something wrong. The test are failing because they are segfaulting not because they are getting incorrect values, as far as I can tell. It does not look like it is the same issue that @alexander-novo came across.

PhilipFackler: 1. I added an `updateVariables()` function which is called from `updateTime()`. I moved the `updateTime()` call in `Ida::runSimulation` to after the solution copy. (I assume this is fine since copying is a temporary solution.)
2. I left the call to `printMonitoredVariables()` after the final copy because, while the last printed variables were indeed very close to the final time for the run, the final variables are different.
3. I removed the call to `step_callback` after the final copy because the tests are not written to account for this (and therefore segfault) and because we intend to remove the callback altogether at some point.

PhilipFackler: I should point out that when we transition to the `Ida`, the `SystemModel`, and the components and buses pointing to the same solution data, `updateVariables` will be removed because copying will not be necessary.

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -31,18 +32,38 @@ namespace GridKit
...
+      II1,
+      IM1,
+      P1,
+      Q1,
+      IR2,
+      II2,
+      IM2,
+      P2,
+      Q2,
+      SIZE
```
pelesh: Echoing comment by @superwhiskers:

_this pull request leaves casing of enums inconsistent in data structures. for instance, the newly added monitorable variables in busdata are all caps, inconsistent with enums in other files. this should be resolved prior to merge._

pelesh: I understand we need to update [contributing guidelines](https://github.com/ORNL/GridKit/blob/develop/CONTRIBUTING.md#enums-enumerated-types) for this, as well.

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
```diff
@@ -6,15 +6,36 @@
...
   namespace PhasorDynamics
   {
+    /// Indices of the variables able to be monitored on this component
+    enum class BusMonitorableVariables : size_t
+    {
+      VR,
+      VI,
+      VM,
+      VA,
+      SIZE
```
pelesh: Capitalization?

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
```diff
@@ -6,15 +6,36 @@
...
+    };
+
+    /**
+     * @brief Convert enum value to string label
+     */
+    inline const std::string& enumLabel(BusMonitorableVariables v)
+    {
+      static const std::string labels[] = {"Vr", "Vi", "Vm", "Va"};
+      return labels[Utilities::enumId(v)];
```
pelesh: We need a discussion whether to allow Magic Enum inside GridKit proper.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -435,6 +441,14 @@ namespace GridKit
...
       }
 
+      template <class ScalarT, typename IdxT>
+      void Ieeet1<ScalarT, IdxT>::initializeMonitor()
+      {
+        using Variable = model_data_type::MonitorableVariables;
+        monitor_.set(Variable::EFD, [this]
+                     { return efd_signal_->read(); });
+        // monitor_.set(Variable::KSAT, [this] { return ?; });
+      }
```
pelesh, line 418: Could this be implemented on the parser side? A component would provide a method `setMonitored(VariableEnum )` with its variables enumeration as the input argument, and the parser would go through all monitored variables and set them one by one. Would such implementation be possible?

CC @superwhiskers

pelesh (changes requested):
More offline discussion is needed before this PR can be merged. Some issues to consider: 
- Conversion of strings to enums should ideally be in parser functions only. Moving them to component data structures is not an ideal design.
- Setting monitored variables should be done in parser. I am not sure the current solution is sustainable.
- Capitalization of enums should not be changed in this PR.
- It is unclear what is the purpose of the `VariableMonitorBase` class.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
PhilipFackler: 1. I would say that this seems out-of-scope for the parser, that the parser should just do parsing, storing the parsed data in a conveniently accessible structure (which I think is done pretty well already).
2. At the time of parsing, the components haven't been constructed (since the components to be constructed depends on the input file). You would have to completely couple `SystemModel` construction with the parser...which (again)...please don't... :pleading_face:

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
PhilipFackler: This follows the guidelines. Do they need to change?

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
PhilipFackler: I corrected this to follow the guidelines (after @superwhiskers pointed it out). Do they need to change?

PhilipFackler:
> Conversion of strings to enums should ideally be in parser functions only. Moving them to component data structures is not an ideal design.

@pelesh That IS done only in the parser. If you're referring to the `enumLabel` function, that is for printing enum values as labels in the monitor output.

GridKit/Model/PhasorDynamics/Branch/Branch.hpp
```diff
@@ -165,6 +173,9 @@ namespace GridKit
       /* Derivied parameters */
```
superwhiskers:
```suggestion
      /* Derived parameters */
```
not the result of this pr, but since it's right here

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
```diff
@@ -31,18 +32,38 @@ namespace GridKit
...
     };
 
+    /**
```
superwhiskers: just calling this out: if we go along with permitting magic enum to be used in this context, we'd have to undo the guideline changes and then we would be able to get rid of this convenience function

GridKit/Model/PhasorDynamics/Branch/BranchImpl.hpp
```diff
@@ -277,6 +251,60 @@ namespace GridKit
...
+    void Branch<ScalarT, IdxT>::initializeMonitor()
+    {
+      using Variable = typename model_data_type::MonitorableVariables;
+      monitor_.set(Variable::IR1, [this]
+                   { return Ir1(); });
+      monitor_.set(Variable::II1, [this]
+                   { return Ii1(); });
+      // monitor_.set(Variable::IM1, [this] { return ?(); });
+      // monitor_.set(Variable::P1, [this] { return ?(); });
+      // monitor_.set(Variable::Q1, [this] { return ?(); });
```
superwhiskers: are these not implemented?

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
```diff
@@ -49,20 +70,12 @@ namespace GridKit
...
-      enum class MonitorableVariables : size_t
-      {
-        VR,
-        VI,
-        VM,
-        VA,
-        MAXIMUM,
-      };
+      /// Alias
+      using MonitorableVariables = BusMonitorableVariables;
```
superwhiskers, line 62: we have this alias here but it's not being done everywhere else?

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -259,6 +265,22 @@ namespace GridKit
...
+    template <class ScalarT, typename IdxT>
+    void Genrou<ScalarT, IdxT>::initializeMonitor()
+    {
+      using Variable = typename model_data_type::MonitorableVariables;
+      monitor_.set(Variable::IR, [this]
+                   { return y_[15]; });
+      monitor_.set(Variable::II, [this]
+                   { return y_[16]; });
+      // monitor_.set(Variable::P, [this] { return ?(); });
+      // monitor_.set(Variable::Q, [this] { return ?(); });
```
superwhiskers: same here: are these not implemented on the side of the genrou?

GridKit/Model/PhasorDynamics/INPUT_FORMAT.md
```diff
@@ -55,6 +55,25 @@ Contained in the `header` key is an object with the following items:
...
+### Monitors
+
+Contained in the `monitors` key is an array of objects, each of which describes
+an output for monitored variables (those listed in the `mon` field of a
+[bus](#buses) or [device](#devices). The following fields are supported:
+
+  Name               | Description
+  -------------------|------------------------------------------------------
+  `file_name`        | Optional string indicating output file name. If omitted, `stdout` is used.
+  `format`           | One of { "CSV", "JSON", "YAML" } (case-insensitive)
```
superwhiskers, line 67: i recall bringing up at some point that it would be better to exclusively output json and use a simpler conversion tool to convert to some other format for one's convenience. are we not doing that---it looks like we're not based on the csv output currently done in this pr.

superwhiskers (changes requested):
overall the code quality looks good. i only have architectural questions that i would prefer be answered before we merge this.

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
pelesh: @superwhiskers, could you explain what you mean by this?

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
pelesh: Thanks for clarifying.

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
superwhiskers: the recent guideline changes nullify this

GridKit/Model/PhasorDynamics/Branch/BranchImpl.hpp
pelesh: I am not sure what is `IM1`. Outputs `P1` and `Q1` need to be computed -- they are not system variables -- and we need a separate function for that. Perhaps @lukelowry or @abirchfield could add those functions in a subsequent PR?

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
PhilipFackler: This is defined in `BusData` and in each component data class (see `GenrouData`, etc...). It is used generically by `VariableMonitor`.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
PhilipFackler: I provided value functions to the monitors where it was obvious how to access the variable associated with given enum values. I left them commented where it was not obvious, intending that someone else who knows the actual meaning of the variables should provide these in a subsequent PR.

GridKit/Model/PhasorDynamics/INPUT_FORMAT.md
PhilipFackler: Comments in that discussion seemed contradictory, so I tried to satisfy all of them since I wasn't sure exactly what to do. :) So I did not write a separate conversion tool. You can use any of these three formats, printed to the console or a file.

GridKit/Model/PhasorDynamics/Branch/Branch.hpp
PhilipFackler: done

GridKit/Model/PhasorDynamics/INPUT_FORMAT.md
pelesh: I think since we have these outputs already implemented, I don't see harm in keeping them.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
pelesh: We need to implement computing active and reactive power injection by the generator in the model. Then we can add them to the output.

GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
pelesh: Updated guidelines provided in #319

GridKit/Model/PhasorDynamics/Bus/BusData.hpp
pelesh: Updated guidelines provided in https://github.com/ORNL/GridKit/pull/319

pelesh (approved):

superwhiskers (approved):

lukelowry (approved):
Invaluable. Thank you @PhilipFackler ! This opens so many doors for case validation.

Any feedback of value from me will come after I use this feature to validate/debug the larger cases.

PR #312, opened 2025-11-24
PhasorDynamics simulations with sparse Jacobians

### LINKED ISSUE #314: Bug: IDA::initializeSimulation errors if t0 is too close to 0.1 and findConsistent is set to true
### Description how to reproduce the bug

If you create a simulation with an initial time of t0, `initializeSimulation` will fail with an error if `findConsistent` is set to true:

```
[ERROR][rank 0][/tmp/sundials/src/idas/idas_ic.c:179][IDACalcIC] tout1 too close to t0 to attempt initial condition calculation.

ERROR: Function IDACalcIC failed with flag -22!

terminate called after throwing an instance of 'AnalysisManager::Sundials::SundialsException'
  what():  Method in Ida class failed!

Aborted (core dumped)
```

### GridKit™ version

latest

N/A

### Additional information

When `findConsistent` is set to true, there is a call to `IDACalcIC`:

https://github.com/ORNL/GridKit/blob/b58914966e0c7d61da74c7afa3eafb99d4eeee5c/GridKit/Solver/Dynamic/Ida.cpp#L277

The third parameter [as documented by SUNDIALS](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDACalcIC):
> tout1 – is the first value of at which a solution will be requested (from [IDASolve()](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDASolve)). This value is needed here only to determine the direction of integration and rough scale in the independent variable .

The passed argument isn't based on t0 at all, and in fact if t0 is too close to 0.1, the above error will be thrown

nkoukpaizan:
## Description

This adds the ability to run `PhasorDynamics` examples with sparse Jacobian. The component Jacobians are obtained with Enzyme and assembled into the system Jacobian.

cc @lukelowry @abirchfield 
 
 ## Proposed changes
- Added missing Jacobian block terms such as $\partial f/\partial y'$, $\partial h/\partial y$ and $\partial f/\partial w_b$
- Added verification in the unit tests for the $\partial f/\partial y'$ terms
- Multiple bug fixes
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 This is still using `COO_Matrix`. I suggest we merge #307 and re-use Jacobian assembly.

 ## Comparison plots between current (sparse Jacobians) and reference (dense Jacobians computed by IDA with finite differences)

ThreeBus_Basic
Example_ThreeBus_Basic_results

ThreeBus_Classical
Example_ThreeBus_Classical_results

TenGenGenrou
TenGenGenrou_Results

TenGenGenClassical
TenGenClassical_results


examples/PhasorDynamics/Tiny/ThreeBus/Classical/ThreeBusClassical.cpp
```diff
@@ -177,6 +177,9 @@ int main()
 
   SystemModel<scalar_type, index_type> sys(data);
   sys.allocate();
+  sys.initialize();
+  sys.evaluateResidual();
+  sys.evaluateJacobian();
```
pelesh: These calls should not be made here. Evaluating residual and Jacobian should be done by the numerical integrator only, not a user code.

```suggestion
```

examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasic.cpp
```diff
@@ -96,6 +96,9 @@ int main()
 
   SystemModel<scalar_type, index_type> sys(data);
   sys.allocate();
+  sys.initialize();
+  sys.evaluateResidual();
+  sys.evaluateJacobian();
```
pelesh: Since this is a workaround we should either hide these lines in SystemModel or comment them as temporary.

examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasicJson.cpp
```diff
@@ -76,6 +76,9 @@ int main(int argc, const char* argv[])
 
   SystemModel<scalar_type, index_type> sys(data);
   sys.allocate();
+  sys.initialize();
+  sys.evaluateResidual();
+  sys.evaluateJacobian();
```
pelesh: See earlier comments. Allocating and initializing system here is ok but evaluating residual and Jacobian may be confusing  for the user.

examples/PhasorDynamics/Tiny/TwoBus/Basic/TwoBusBasic.cpp
nkoukpaizan: Addressed in [f470a32](https://github.com/ORNL/GridKit/pull/312/commits/f470a32788b83bcec1c6362ae658307243298c2c)

GridKit/AutomaticDifferentiation/Enzyme/DfDwb.hpp
```diff
@@ -0,0 +1,171 @@
...
+    {
+      /**
+       * @brief Enzyme automatic differentiation Jacobian evaluator : df/dwb
+       *
+       * @tparam ModelT - model type
+       * @tparam MemberFunctions - member function parameter key
+       * @tparam ScalarT - scalar data type
+       * @tparam IdxT - matrix index data type
+       */
+      template <typename ModelT, MemberFunctions function, class ScalarT, typename IdxT>
```
pelesh, line 29: This is really nitpicking but perhaps we should use `FunctionT` instead of `function` for the template parameter name?
```suggestion
      template <typename ModelT, MemberFunctions FunctionT, class ScalarT, typename IdxT>
```

GridKit/AutomaticDifferentiation/Enzyme/DfDwb.hpp
```diff
@@ -0,0 +1,171 @@
...
+                         ScalarT*                 y,
+                         ScalarT*                 yp,
+                         ScalarT*                 wb,
+                         MatrixT&                 jac)
+        {
+          if (n_res > 0 && n_var > 0)
+          {
+            std::vector<Triple<ScalarT>> triplets;
+            std::vector<ScalarT>         elementary_v(n_var);
+            for (size_t var_i = 0; var_i < n_var; ++var_i)
```
pelesh: I would add a comment here that vector of triplets is to be replaced soon when we start using CSR matrix object.

```suggestion
            // TODO: The vector of triplets will be replaced with `CsrMatrix` object
            std::vector<Triple<ScalarT>> triplets;
            std::vector<ScalarT>         elementary_v(n_var);
            for (size_t var_i = 0; var_i < n_var; ++var_i)
```

GridKit/AutomaticDifferentiation/Enzyme/DfDwb.hpp
```diff
@@ -0,0 +1,171 @@
...
+              ScalarT* output   = __enzyme_todense<ScalarT*>((void*) ident_load<ScalarT>, (void*) ident_store<ScalarT>, var_i);
+              ScalarT* d_output = __enzyme_todense<ScalarT*>((void*) sparse_load<ScalarT>, (void*) sparse_store<ScalarT>, var_i, &triplets);
+
+              // Elementary vector for Jacobian-vector product
+              std::ranges::fill(elementary_v, 0.0);
+              elementary_v[var_i] = 1.0;
+
+              // Autodiff
+              __enzyme_fwddiff<void>((void*) ModelWrapper<ModelT, function, ScalarT>::eval,
+                                     enzyme_const,
```
pelesh, line 72: I would be really good to document this function. It is not self-explanatory.

GridKit/AutomaticDifferentiation/Enzyme/DfDwb.hpp
```diff
@@ -0,0 +1,171 @@
...
+          if (n_res > 0 && n_var > 0)
+          {
+            std::vector<Triple<ScalarT>> triplets;
+            std::vector<ScalarT>         elementary_v(n_var);
+            for (size_t var_i = 0; var_i < n_var; ++var_i)
+            {
+              // Sparse storage
+              ScalarT* output   = __enzyme_todense<ScalarT*>((void*) ident_load<ScalarT>, (void*) ident_store<ScalarT>, var_i);
+              ScalarT* d_output = __enzyme_todense<ScalarT*>((void*) sparse_load<ScalarT>, (void*) sparse_store<ScalarT>, var_i, &triplets);
+
```
pelesh, line 132: It would be good to comment what these functions do.

GridKit/AutomaticDifferentiation/Enzyme/DfDws.hpp
```diff
@@ -0,0 +1,112 @@
...
+namespace GridKit
+{
+  namespace Enzyme
+  {
+    namespace Sparse
+    {
+      /**
+       * @brief Enzyme automatic differentiation Jacobian evaluator: df/dws
+       *
+       * @tparam ModelT - model type
```
pelesh, line 25: Should this file be in this folder? This seems specific to phasor domain dynamics.

GridKit/AutomaticDifferentiation/Enzyme/LowerSparseStorage.hpp
```diff
@@ -0,0 +1,116 @@
+/**
+ * @file LowerSparseStorage.hpp
+ * @author Nicholson Koukpaizan (koukpaizannk@ornl.gov)
```
pelesh, line 8: What is "lower sparse storage"?

GridKit/AutomaticDifferentiation/Enzyme/ModelWrappers.hpp
```diff
@@ -0,0 +1,188 @@
...
+         */
+        static void eval(ModelT* model, ScalarT* y, ScalarT* yp, ScalarT* wb, ScalarT* h)
+        {
+          model->evaluateBusResidual(y, yp, wb, h);
+        }
+      };
+
+      /**
+       * @brief Residual wrapper partial template specialization for BusResidual11
+       *
```
pelesh: Did you mean _branch_ residual?

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -275,9 +273,12 @@ namespace AnalysisManager
         if (tag_)
           initType = IDA_YA_YDP_INIT;
 
-        retval = IDACalcIC(solver_, initType, 0.1);
+        retval = IDACalcIC(solver_, initType, t0 + 0.1);
```
pelesh, line 276: This magic number has to go. The `SystemModel` should estimate `dt` (now hardwired to 0.1) and `Ida` class should obtain it from there. This is probably out of scope of this PR.

pelesh (approved):
Great job! All tests pass and the PR fixes quite a few minor issues along the way. I left a few nitpicking comments mainly requesting better documentation within the code.

GridKit/AutomaticDifferentiation/Enzyme/DfDwb.hpp
PhilipFackler: I don't think so, since `function` is a "value" parameter, not a "type" parameter. Do we have a naming convention for template value parameters?

pelesh: @PhilipFackler, thanks for clarification. We don't have a naming convention for template value parameters, so I suggest we let this stand as the precedent.

GridKit/AutomaticDifferentiation/Enzyme/LowerSparseStorage.hpp
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      /**
+       * @brief Enzyme sparse accumulation for double
+       *
+       */
+      [[maybe_unused]] __attribute__((enzyme_sparse_accumulate)) static void inner_storedbl(size_t row, size_t col, double val, std::vector<Triple<double>>& triplets)
+      {
+        triplets.emplace_back(row, col, val);
+      }
```
PhilipFackler, line 74: I don't know enzyme, so there may be a reason for this. Otherwise, these two functions should be combined as a template function.

GridKit/AutomaticDifferentiation/Enzyme/DfDy.hpp
```diff
@@ -0,0 +1,259 @@
...
+                         size_t                   n_res,
+                         size_t                   n_var,
+                         const std::vector<IdxT>& res_indices,
+                         const std::vector<IdxT>& var_indices,
+                         ScalarT*                 y,
+                         ScalarT*                 yp,
+                         ScalarT*                 wb,
+                         RealT                    alpha,
+                         MatrixT&                 jac)
+        {
```
PhilipFackler, line 57: This is more of a style issue, but I think I would make these free template functions rather than static member functions to make the function call site(s) cleaner. For example, if you change this call to:

```c++
template <MemberFunctions function, typename ModelT, typename ScalarT, typename RealT, typename IdxT>
void dfdy(
    ModelT*                  model,
    size_t                   n_res,
    size_t                   n_var,
    const std::vector<IdxT>& res_indices,
    const std::vector<IdxT>& var_indices,
    ScalarT*                 y,
    ScalarT*                 yp,
    ScalarT*                 wb,
    RealT                    alpha,
    GridKit::LinearAlgebra::COO_Matrix<RealT, IdxT>&                 jac)
{
    static_assert(std::is_same_v<RealT, GridKit::ScalarTraits<ScalarT>::RealT>);
    /*
    ...
    */
}
```

Then the call in GenrouEnzyme.cpp would look like this:

```c++
constexpr auto method = GridKit::Enzyme::Sparse::MemberFunctions::InternalResidualWithSignal;

GridKit::Enzyme::Sparse::dfdy<method>(
                                                this,
                                                f_.size(),
                                                y_.size(),
                                                this->getResidualIndices(),
                                                this->getVariableIndices(),
                                                y_.data(),
                                                yp_.data(),
                                                wb_.data(),
                                                ws_.data(),
                                                alpha_,
                                                J_);
```

I would go a step further and refactor the `MemberFunctions` enum as a namespace with each value as a struct name. That way they could be used as arguments to the function instead of having to have the one template argument.

But again, of course, all of this is just style.

GridKit/Model/PhasorDynamics/SignalNode/SignalNode.hpp
```diff
@@ -131,7 +131,7 @@ namespace GridKit
       IdxT     signal_id_{0};
 
     protected:
-      const IdxT bus_id_{static_cast<IdxT>(-1)};
+      const IdxT bus_id_{INVALID_INDEX<IdxT>};
```
PhilipFackler, line 134: This kind of update is great. high-five!

PhilipFackler (approved):
Looks good!

GridKit/AutomaticDifferentiation/Enzyme/DfDy.hpp
pelesh: @PhilipFackler, could you create an issue along these lines? What you suggest is a good idea but somewhat beyond the scope of this PR.

GridKit/AutomaticDifferentiation/Enzyme/LowerSparseStorage.hpp
pelesh: Clarified in code comments.

PR #319, opened 2025-12-17
Update CONTRIBUTING.md

pelesh:
## Description
 
 Remove style guidelines for `enum`s because we are already making exceptions for enums entering the JSON parser. 
 
@superwhiskers 
 
 ## Proposed changes
 
For now, just remove style guidelines for enums. 
 
## Checklist
 
N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -249,17 +249,6 @@ For consistency, use the same name everywhere the same type is used.
   template <typename real_type>; // No, `_type` used in a template parameter name
 ```
 
-### Enums (enumerated types)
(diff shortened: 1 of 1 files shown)


PhilipFackler:
I'm surprised you're removing the guideline altogether. I would think we would want enum value formatting to be consistent across the project, even if the style should be different from what this section says.

pelesh:
> I'm surprised you're removing the guideline altogether. I would think we would want enum value formatting to be consistent across the project, even if the style should be different from what this section says.

Updated guidelines for `enums`.

PR #329, opened 2026-01-06
Add IDA Stat struct

### LINKED ISSUE #322: Get IDA stats as data
### Summary

Currently, the only way to get statistics about the underlying IDA workspace (such as number of function evaluations, number of linear solves, number of nonlinear solves, etc.) is by calling `Ida::printFinalStats()`, which prints this information to the console.

It would be nice to have a function which collects this data into a structure to be used further, rather than just printing it straight to output.

### Rationale

When a discrete event, such as a fault, is encountered, the IDA simulation needs to be re-initialized using `Ida::initializeSimulation()`, but doing so will clear all current stats from the workspace and the next time `Ida::printFinalStats()` is called, the stats will all be zero.

It's difficult to use this to collect statistics about simulations that include faults - instead statistics must be printed separately for all of the fault regions and cannot be aggregated.

As well, sometimes stats need to be put into a file with a specific formatting to be plotted or put into a table and compared to stats from other integrators, but the current usage makes that difficult.

### Description

It would be nice if there was an `Ida::getStats()` function that returned a simple struct that kept track of all of the stats printed by IDA and allowed for accumulation.

### Additional information

_No response_

alexander-novo:
## Description
 
Adds a new structure, `IdaStats`, which keeps track of several statistics returned by IDA. A user can query these statistics from IDA by calling `Ida::getStats()` and can accumulate several runs of stats in the structure. There is also functionality to print a report, similar to the `Ida::printFinalStats()` function after the stats have been accumulated together.
 
 Closes #322
 
@Steven-Roberts let me know if there's anything else you think I should put in.
 
 ## Proposed changes

It makes sense for `Ida::getStats()` to be `const`, but then `Ida::checkAllocation()` and `Ida::checkOutput()` are inaccessible. These functions never touch any IDA state, so it makes sense for these functions to be `static` anyway.
 
 ## Checklist
 
- [x] All tests pass.

There are not unit tests, but I have added a printout of the report to the `rlcircuit` example:

```
                         Steps :           70
                Residual evals :           86
         Linear decompositions :           19
           Error test failures :            0
                 Linear solves :           86
Nonlinear convergence failures :            0
```
 
 ## Further comments

### DIFF
--- a/GridKit/Solver/Dynamic/Ida.cpp
+++ b/GridKit/Solver/Dynamic/Ida.cpp
@@ -979,7 +979,7 @@ namespace AnalysisManager
       num_residual_evals_              += other.num_residual_evals_;
       num_linear_decompositions_       += other.num_linear_decompositions_;
       num_error_test_fails_            += other.num_error_test_fails_;
-      num_linear_solves_               += other.num_linear_solves_;
+      num_nonlinear_iters_             += other.num_nonlinear_iters_;
       num_nonlinear_convergence_fails_ += other.num_nonlinear_convergence_fails_;
 
       return *this;
@@ -1000,7 +1000,7 @@ namespace AnalysisManager
           << std::setw(label_width) << "Residual evals" << " : " << std::setw(stat_width) << num_linear_decompositions_ << '\n'
           << std::setw(label_width) << "Linear decompositions" << " : " << std::setw(stat_width) << num_linear_decompositions_ << '\n'
           << std::setw(label_width) << "Error test failures" << " : " << std::setw(stat_width) << num_error_test_fails_ << '\n'
-          << std::setw(label_width) << "Linear solves" << " : " << std::setw(stat_width) << num_linear_solves_ << '\n'
+          << std::setw(label_width) << "Nonlinear iterations" << " : " << std::setw(stat_width) << num_nonlinear_iters_ << '\n'
           << std::setw(label_width) << "Nonlinear convergence failures" << " : " << std::setw(stat_width) << num_nonlinear_convergence_fails_;
 
       return out.str();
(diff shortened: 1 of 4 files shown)


GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -968,6 +969,87 @@ namespace AnalysisManager
...
+                                         &stats.num_error_test_fails_,
+                                         &dummy,
+                                         &dummy,
+                                         &dummy2,
+                                         &dummy2,
+                                         &dummy2,
+                                         &dummy2);
+      checkOutput(retval, "IDAGetIntegratorStats");
+
+      retval = IDAGetNonlinSolvStats(solver_, &stats.num_linear_solves_, &stats.num_nonlinear_convergence_fails_);
```
Steven-Roberts: `IDAGetNonlinSolvStats` gives the number of **non**linear iterations, so `&stats.num_linear_solves_` shouldn't be the argument. I suppose since the linear solve is direct, these number coincide, but this could cause issues in the future.

Steven-Roberts (changes requested):
> @Steven-Roberts let me know if there's anything else you think I should put in.

This looks like a reasonable set of stats to start with.

GridKit/Solver/Dynamic/Ida.cpp
alexander-novo: I think the stat we're interested in is number of linear solves. That's what @pelesh has asked for in reports previously. But it should be that number of linear solves = number of nonlinear iterations, right?

Steven-Roberts: On second though, I think this is ok then.

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -968,6 +969,87 @@ namespace AnalysisManager
...
+          "Residual evals",
+          num_residual_evals_,
+          "Linear decompositions",
+          num_linear_decompositions_,
+          "Error test failures",
+          num_error_test_fails_,
+          "Linear solves",
+          num_linear_solves_,
+          "Nonlinear convergence failures",
+          num_nonlinear_convergence_fails_);
```
pelesh: For consistency, use I/O stream.

GridKit/Solver/Dynamic/Ida.cpp
pelesh: Probably better to call it what it is in the IDA output. Later these numbers can be interpreted as needed.

pelesh (approved):
A couple of minor things to fix before this could be merged, otherwise very helpful. Thanks!

PR #330, opened 2026-01-08
Minor performance improvements to PowerElectronics residual evaluation

### LINKED ISSUE #324: Small performance improvements to power electronics module
When profiling the 10,000 IBR simulation in the partitioned integrators repo, we found some not-insignificant performance impacts from some very small changes to the power electronics module:

- `CircuitComponent::connection_nodes_` is accessed in a random-access manner extremely often, which incurs the cost of red-black tree traversal a lot. It fits very nicely into an array instead, and making this replacement removed some percents from the simulation time. [(see)](https://github.com/ORNL/GridKit/commit/dee2fda0304671ec62b19992be10ab56e795579f)
- A few functions in `PowerElectronicsModel` (such as `evaluateResidual()` and `distributeVectors()` call several small functions many many times in loops. These functions are often just accessors, but they are virtual and aren't marked as `final`/`const` as necessary in sub classes. That means they can't be inlined and the compiler can't infer that there are no side effects, so LICM can't be performed. I was able to remove some percents from the simulation time by performing LICM manually [(see)](https://github.com/ORNL/GridKit/commit/66b4bdc1a132371f52edcf771c650c404a44c630). It may also be a good idea to add `final`/`const` to functions which need it to help the compiler perform optimizations elsewhere as well.

Let me know if these seem like good changes to make, and I can add a PR

alexander-novo:
## Description

Changed `connection_nodes_` to an array to prevent tree searches in hot loops.
 
There are some virtual functions implemented by `CircuitComponent` which aren't marked `final` or `const` and are called in a couple of hot loops inside `evaluateResidual()`. The optimizer can't inline these functions or perform LICM, so these functions end up being called billions of times during the simulation, which contributes a few percents to the overall runtime of the simulation.

I've gone through and marked many of these functions as `final` and added `const` variants. These functions could probably be marked as `const` in `Evaluator`, but I didn't want to change something that would affect the phasor dynamics module without a thumbs up.

I've also manually performed LICM in the hot loops in `evaluateResidual()` because I don't want to depend on the optimizer getting this right - this could very easily regress in the future.
 
Closes #324
 
 ## Proposed changes

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -31,7 +31,10 @@ namespace GridKit
       }
     };
 
-    void updateTime(RealT t, RealT a)
+    /**
+     * @note Cannot be marked final, since it is overriden to recurse in the system model.
+     */
+    void updateTime(RealT t, RealT a) override
     {
       this->time_  = t;
       this->alpha_ = a;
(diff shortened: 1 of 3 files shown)


pelesh (approved):
Great work! Please fix changeling conflict before merging.

GridKit/Model/PowerElectronics/CircuitComponent.hpp
```diff
@@ -71,185 +86,205 @@ namespace GridKit
...
   protected:
-    size_t               n_extern_;
-    size_t               n_intern_;
-    std::set<IdxT>       extern_indices_;
+    size_t         n_extern_;
+    size_t         n_intern_;
+    std::set<IdxT> extern_indices_;
     ///@todo may want to replace the mapping of connection_nodes to Node objects instead of IdxT. Allows for container free setup
-    std::map<IdxT, IdxT> connection_nodes_;
+    IdxT*          connection_nodes_ = nullptr;
```
Steven-Roberts, line 287: Could you replace this with `std::vector` to avoid the manually memory management?

pelesh: @Steven-Roberts I would avoid STL containers for now. .

Steven-Roberts: I don't think it would have any overhead, but the main reason I suggested this was the raw pointer will cause a double delete if a copy constructor or assignment is used.

PR #332, opened 2026-01-09
Fixed Step IDA

### LINKED ISSUE #326: IDA Fixed Time Step Support
### Summary

If this is something that we want as a proper GridKit option, I think this would require adding a `setTimeStep` member function for `GridKit::Model::Evaluator` and its subclasses. If the out parameter is 0, this will indicate to use adaptive steps, otherwise the given time step will be used. In `IDA.c` there would need to be some minor changes to the integrator and nonlinear solver configuration when using fixed steps as done in https://code.ornl.gov/scidac-oe/fixed-step-ida-demo. I can start a PR if this sounds reasonable.

### Rationale

This will be useful for convergence studies and comparisons against adaptive time stepping

### Description

_No response_

### Additional information

_No response_

Steven-Roberts:
## Description
 
Adds option to run IDA with fixed time steps
 
Closes #326 
 
 ## Proposed changes
 
Adds `dt` argument to IDA constructor. If nonzero, that fixed time step will be used, otherwise the time step is adaptive.
 
 ## Checklist
 
- [ ] All tests pass.
 
 ## Further comments
 
Any preference on passing `dt` via constructor versus getter/setter @pelesh? Whatever we decide for `dt` I can do for tolerances and max steps too.


Steven-Roberts:
Relative tolerances are now specified by `IDA::setTolerance` (and `IDA::setBackwardTolerance`), and absolute tolerances are specified by each `Evaluator` instance similar using an approach similar to tagging of differential vs algebraic variables. Currently, I have all the abs tols defaulted to 0. If this approach looks ok, I will need some help with determining reasonable defaults for each `Evaluator` subclass @pelesh.

examples/PowerElectronics/ScaleMicrogrid/ScaleMicrogrid.cpp
```diff
@@ -333,6 +329,8 @@ int test(index_type Nsize, real_type error_tol, bool debug_output)
 
   // setup simulation
   idas->configureSimulation();
+  idas->setTolerance(rel_tol);
+  idas->setMaxSteps(SCALE_MICROGRID_MAX_STEPS);
```
pelesh, line 333: Why not provide option to set global absolute tolerance?

pelesh: Similar to other instances in the code.

GridKit/Model/PhasorDynamics/Branch/BranchImpl.hpp
```diff
@@ -140,6 +140,15 @@ namespace GridKit
...
     }
 
+    /**
+     * @brief Set absolute tolerance
+     */
+    template <class ScalarT, typename IdxT>
+    int Branch<ScalarT, IdxT>::setAbsoluteTolerance()
+    {
+      return 0;
+    }
```
pelesh, line 150: It would be good to provide an instance of setting absolute tolerances. My suggestion would be to have nominal values for variables defined in each component model and then absolute tolerance computed as `abs_tol[i] = rel_tol*y_nominal[i]`.

There should be an option to set global absolute tolerance directly in the solver circumventing component local settings imho.

GridKit/Model/PhasorDynamics/Component.hpp
```diff
@@ -53,15 +53,14 @@ namespace GridKit
...
-        atol = abs_tol_;
+        return abs_tol_;
       }
 
-      virtual void setMaxSteps(IdxT& msa) const override
+      const std::vector<ScalarT>& absoluteTolerance() const override
       {
-        msa = max_steps_;
+        return abs_tol_;
       }
```
pelesh, line 64: What happens if component does not set its absolute values? 

My suggestion would be to return default `abs_tol_[i] == rel_tol`, i.e. assuming that the nominal value of the variable is ~1.

pelesh:
> Any preference on passing `dt` via constructor versus getter/setter @pelesh? Whatever we decide for `dt` I can do for tolerances and max steps too.

I am in favor of getter/setters.

pelesh:
> Currently, I have all the abs tols defaulted to 0. If this approach looks ok, I will need some help with determining reasonable defaults for each `Evaluator` subclass @pelesh.

I would suggest `abs_tol` to default to `rel_tol` or `0.1*rel_tol`.

PR #345, opened 2026-03-02
Performance improvements in PhasorDynamics component Jacobian evaluation.

nkoukpaizan:
## Description
 
This improves the performance of component Jacobian evaluation in PhasorDynamics. On ma Mac, this brings the runtime down from ~4.5sec to ~1.5sec for Hawaii.
 
 ## Proposed changes
 - Modified sparse storage approach in the Enzyme wrappers to directly update `rows`, `columns` and `values` and keep track on `nnz` (as opposed to using `emplace_back` on `std::vectors`).
 - Pre-allocate buffers for `rows`, `columns` and `values` to save on allocation/deallocations.
 - Store unsorted entries in `J_` to save on sorting overhead. Instead, rely on the one-time sort and deduplication in `CooMatrix`.
 - For bus Jacobian, perform the reduction into the 2x2 matrices right away at the component level. This remains low overhead for such a small matrix.

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 The main obstacle to removing `COO_Matrix` from the component level is the current need to deduplicate bus Jacobian entries right away. First, because the bus is not yet able to allocate a large enough buffer. Second, because faults introduce new entries that would break the System-level mappings. Perhaps something for a follow-up PR.

### DIFF
--- a/GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
+++ b/GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
@@ -48,7 +48,7 @@ namespace GridKit
 
       // --- Functions which call sort ---
       std::tuple<std::vector<IdxT>, std::vector<RealT>>                       getRowCopy(IdxT r);
-      std::tuple<std::vector<IdxT>&, std::vector<IdxT>&, std::vector<RealT>&> getEntries();
+      std::tuple<std::vector<IdxT>&, std::vector<IdxT>&, std::vector<RealT>&> getEntries(const bool sort = true);
       std::tuple<std::vector<IdxT>, std::vector<IdxT>, std::vector<RealT>>    getEntryCopies();
       std::tuple<std::vector<IdxT>, std::vector<IdxT>, std::vector<RealT>>    getEntryCopiesSubMatrix(std::vector<IdxT> submap);
 
@@ -57,10 +57,12 @@ namespace GridKit
 
       // Set values from vector storage. Will sort before storing
       void setValues(std::vector<IdxT> r, std::vector<IdxT> c, std::vector<RealT> v);
+      void setValues(RealT alpha, IdxT* r, IdxT* c, RealT* v, IdxT nnz);
 
       // BLAS. Will sort before running
       void  axpy(RealT alpha, COO_Matrix<RealT, IdxT>& a, const bool sort = true);
       void  axpy(RealT alpha, std::vector<IdxT> r, std::vector<IdxT> c, std::vector<RealT> v, const bool sort = true);
+      void  axpy(RealT alpha, IdxT* r, IdxT* c, RealT* v, IdxT nnz);
       void  scal(RealT alpha);
       RealT frobNorm();
 
@@ -78,6 +80,7 @@ namespace GridKit
       void sortSparse();
       bool isSorted();
       IdxT nnz();
+      void deduplicate();
 
       std::tuple<IdxT, IdxT> getDimensions();
 
@@ -85,7 +88,7 @@ namespace GridKit
 
       void printMatrixMarket(const std::string& filename, const std::string& comment);
 
-      static void sortSparseCOO(std::vector<IdxT>& rows, std::vector<IdxT>& columns, std::vector<RealT>& values);
+      static void sortSparseCOO(std::vector<IdxT>& rows, std::vector<IdxT>& columns, std::vector<RealT>& values, size_t nnz);
 
     private:
       IdxT indexStartRow(const std::vector<IdxT>& rows, IdxT r);
@@ -135,9 +138,9 @@ namespace GridKit
      * @return std::tuple<std::vector<IdxT>, std::vector<IdxT>, std::vector<RealT>>
      */
     template <typename RealT, typename IdxT>
-    inline std::tuple<std::vector<IdxT>&, std::vector<IdxT>&, std::vector<RealT>&> COO_Matrix<RealT, IdxT>::getEntries()
+    inline std::tuple<std::vector<IdxT>&, std::vector<IdxT>&, std::vector<RealT>&> COO_Matrix<RealT, IdxT>::getEntries(const bool sort)
     {
-      if (!this->sorted_)
+      if (!this->sorted_ && sort)
       {
         this->sortSparse();
       }
@@ -239,7 +242,7 @@ namespace GridKit
     inline void COO_Matrix<RealT, IdxT>::setValues(std::vector<IdxT> r, std::vector<IdxT> c, std::vector<RealT> v)
     {
       // sort input
-      this->sortSparseCOO(r, c, v);
+      this->sortSparseCOO(r, c, v, r.size());
 
       // Duplicated with axpy. Could replace with function depdent on lambda expression
       size_t a_iter = 0;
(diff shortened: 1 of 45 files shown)


examples/Enzyme/Standalone/EnzymeSparse.cpp
```diff
@@ -100,35 +36,59 @@ __attribute__((always_inline)) static void f(size_t N, ScalarT* input, ScalarT*
...
+        rows[current_nnz] = idx;
+        cols[current_nnz] = idy;
+        vals[current_nnz] = 2.0 * x[idy];
+        current_nnz++;
       }
     }
   }
-  jac.setValues(rtemp, ctemp, valtemp);
+  return new SparseMatrix(N, N, current_nnz, &rows, &cols, &vals);
 }
```
pelesh, line 63:
```suggestion
  // Hijacking constructor sets rows, cols and vals to nullptr
  return new SparseMatrix(N, N, current_nnz, &rows, &cols, &vals);
}
```

examples/Enzyme/Standalone/EnzymeSparse.cpp
```diff
@@ -100,35 +36,59 @@ __attribute__((always_inline)) static void f(size_t N, ScalarT* input, ScalarT*
...
 
 /// Function that computes the Jacobian via automatic differentiation
 template <typename ScalarT>
-__attribute__((noinline)) void jac_f(size_t N, ScalarT* input, SparseMatrix& jac)
+__attribute__((noinline)) SparseMatrix* jac_f(size_t N, ScalarT* input)
```
pelesh, line 67: Why not use reference to `SparseMatrix`?

examples/Enzyme/Standalone/EnzymeSparse.cpp
```diff
@@ -100,35 +36,59 @@ __attribute__((always_inline)) static void f(size_t N, ScalarT* input, ScalarT*
...
-  std::vector<Triple<ScalarT>> triplets;
+  size_t* index_maps = new size_t[N];
+  for (size_t i = 0; i < N; i++)
+  {
+    index_maps[i] = i;
+  }
+
+  size_t*  rows_buffer = new size_t[N * N];
+  size_t*  cols_buffer = new size_t[N * N];
+  ScalarT* vals_buffer = new ScalarT[N * N];
```
pelesh, line 77: Why use dense buffer here?

examples/Enzyme/Standalone/EnzymeSparse.cpp
```diff
@@ -141,30 +101,45 @@ __attribute__((noinline)) void jac_f(size_t N, ScalarT* input, SparseMatrix& jac
...
-    valtemp.push_back(tup.val);
+    rows[ind] = rows_buffer[ind];
+    cols[ind] = cols_buffer[ind];
+    vals[ind] = vals_buffer[ind];
   }
-  jac.setValues(rtemp, ctemp, valtemp);
+  delete[] index_maps;
+  delete[] rows_buffer;
+  delete[] cols_buffer;
+  delete[] vals_buffer;
```
pelesh, line 117: Creating and deleting dense matrix buffer will likely create a memory complexity bottleneck and performance issues. Probably a good idea to mark this with a comment, indicating this is a temporary solution.

GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
```diff
@@ -279,6 +282,32 @@ namespace GridKit
...
+     *
+     * @param[in] r row indices to be stored
+     * @param[in] c column indices to be stored
+     * @param[in] v values to be stored
+     * @param[in] nnz to be stored
+     *
+     */
+    template <typename RealT, typename IdxT>
+    inline void COO_Matrix<RealT, IdxT>::setValues(RealT alpha, IdxT* r, IdxT* c, RealT* v, IdxT nnz)
+    {
```
pelesh, line 299: Why are we implementing this in deprecated `COO_Matrix` class instead of in `CooMatrix`?

GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
```diff
@@ -422,6 +451,94 @@ namespace GridKit
...
+     * @tparam IdxT - Integer data type for matrix indices
+     *
+     * @param[in] r row indices to be stored
+     * @param[in] c column indices to be stored
+     * @param[in] v values to be stored
+     * @param[in] nnz to be stored
+     *
+     */
+    template <typename RealT, typename IdxT>
+    inline void COO_Matrix<RealT, IdxT>::axpy(RealT alpha, IdxT* r, IdxT* c, RealT* v, IdxT nnz)
```
pelesh, line 471: Consider making this a standalone method rather than member of `COO_Matrix` class.

GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
```diff
@@ -422,6 +451,94 @@ namespace GridKit
...
+     * @brief Deduplicate matrix entries
+     *
+     * @node This is currently only used in component-level Jacobian tests,
+     * which don't use CooMatrix or CsrMatrix yet, to compare with Dependency::Tracking maps.
+     *
+     * @tparam RealT - Real type for Jacobian entries
+     * @tparam IdxT - Integer data type for matrix indices
+     */
+    template <typename RealT, typename IdxT>
+    inline void COO_Matrix<RealT, IdxT>::deduplicate()
```
pelesh, line 507: This should probably be a standalone method rather than `COO_Matrix` member.

pelesh:
Hawaii case runs in ~1.5s on my laptop now.

pelesh (commented):
This is significant performance improvement and I suggest reviewing and merging it as soon as it is ready (not in the draft state).

I suggest not adding new member methods to `COO_Matrix` class but creating standalone methods instead.

I am not sure I understand the argument for keeping `COO_Matrix` quite well. Probably requires more offline discussion.

examples/Enzyme/Standalone/EnzymeSparse.cpp
nkoukpaizan: The buffers should only be allocated once per simulation. Note that this is an example; we happen to know `nnz` a priori here (lower triangular matrix), but that's not the case in general. I can't think of any other way to allocate buffers a priori.

examples/Enzyme/Standalone/EnzymeSparse.cpp
nkoukpaizan: I'm creating the matrix in the function here and using the hijacking constructor. There are other ways to do it, but I'm not sure what the advantages would be in this example.

GridKit/LinearAlgebra/SparseMatrix/COO_Matrix.hpp
nkoukpaizan: Sloppy on my part, I know. That was the shortest path to getting things to work. There are things that require more thought before I can remove the use of COO_Matrix at the component level.

lukelowry:
I can run the Hawaii case in 2.7s on my machine with this branch! Getting very close to existing solvers. A beta version of the Texas case in GK solves in ~30 seconds, which is just 3x the runtime for PowerWorld. Great work

pelesh (approved):
I suggest we merge it as an intermediate step. Performance improvement is significant and we have a clear to-do list of further improvements, which could be reviewed in the next PR.

PR #346, opened 2026-03-04
Texas Case Simulation

lukelowry:
## Description
 
Added the skeleton code needed for the new case (Texas, 2k bus), along with some other minor tweaking of the other models that was only possible to identify due to insane new solve time. I can solve the Texas case *almost* in real time (1:1) which brings GK in onto the same order of magnitude as PowerWorld (whoop @nkoukpaizan !)
 
 
 ## Proposed changes
 
- Texas Case added
- Tweaks to the limits and models of Hawaii & ISO New England Case
- Simplified one of the genrou initialization equations
 
 ## Checklist
 
 
- [X] All tests pass.
 
 ## Further comments
 
Want to get thoughts before PR. The saturation preventing a steady state means I had to solve for 10s to get a steady solution, then apply the fault, then let it run for 10s more in Texas. Not really a problem right now, but we should fix the nonlinear initialization.
 
texas_results


GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/README.md
```diff
@@ -159,9 +159,8 @@ from the steady-state initial conditions.
   k_{sat}     &= S_B(\psi^{''}-S_A)^2 \\
   T_{elec}    &= (\psi''_{d} - I_dX_d^{''})I_q-(\psi''_{q} - I_qX_d^{''})I_d \\
   P_{m}    &= T_{elec} \\
-  \psi^{'}_d  &=
-  \dfrac{\psi^{''}_d-X_{d5}X_{d2}I_d}{X_{d5}+1}\\
-  \psi^{'}_q  &=\dfrac{X_{q5}X_{q2}I_q-\psi^{''}_q}{X_{q5}+1}\\
+  \psi^{'}_d  &= \psi^{''}_d - (X^{''}_d - X_\ell)I_d \\
+  \psi^{'}_q  &= (X^{''}_q - X_\ell)I_q - \psi^{''}_q \\
```
pelesh: I think this will give you nicer LaTeX rendering:
```suggestion
  \psi_d'  &= \psi_d'' - (X_d'' - X_\ell)I_d \\
  \psi_q'  &= (X_q'' - X_\ell)I_q - \psi_q'' \\
```

pelesh (commented):
Since we have a reusable app for dynamic simulations, we might want to consider having it in a single place and then launching it with different JSON input files. That way, we need to change code at only one place.

lukelowry:
> Since we have a reusable app for dynamic simulations, we might want to consider having it in a single place and then launching it with different JSON input files. That way, we need to change code at only one place.

I think a case library would be a useful thing to have at the root of the repo - `examples/` could be strictly GridKit implementations, `cases/` would contain the .json input formats. Is this the direction you were thinking? 

We might reconsider having 'BusFault' be a .json input format attribute. It is not intuitive to batch this in with the case IMO. If we can create a clearer boundary between the case (i.e., Texas, Hawaii) and the *event* (i.e. Fault, Outage) the examples would become more approachable and readable, I think.

pelesh:
> We might reconsider having 'BusFault' be a .json input format attribute. It is not intuitive to batch this in with the case IMO. If we can create a clearer boundary between the case (i.e., Texas, Hawaii) and the _event_ (i.e. Fault, Outage) the examples would become more approachable and readable, I think.

We might want to replace each bus with what we now call `BusFault`. That way, we could inject a fault in any bus and the event would be separate from the model.

lukelowry:
This is ready for others to use. I have been using this to simulate and test. It works well - the concerns raised above are not addressed, but are deferred to #359 and maybe #363

abirchfield (approved):

pelesh (approved):
Texas case ran under 20s on my laptop. Great job!

PR #347, opened 2026-03-04
Add support for fixed time steps with IDA

### LINKED ISSUE #326: IDA Fixed Time Step Support

Steven-Roberts:
## Description
 
Adds option to run IDA with fixed time steps
Improves setting of absolute tolerances
Fixes incorrect resizing of the tagging vectors

Closes #326
 
 ## Proposed changes
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
This replaces #332 which was not rebased correctly and the branch name didn't follow naming conventions.

Prior to this PR, many tests and examples included algebraic errors in IDA's error norm because the tagging vectors were improperly sized and [this branch](https://github.com/ORNL/GridKit/blob/d8128ce4c6c3a925ac092f5207e1a3d9d568ff19/GridKit/Solver/Dynamic/Ida.cpp#L103) was never entered. With this fixed, the Microgrid and ScaleMicrogrid tests now fail because error is measured differently. Digging a bit deeper, I suspect the reference solution used in those examples is not very accurate or corresponded to a slightly different residual function. Even on the develop branch, reducing the tolerances does not improve the error reported [here](https://github.com/ORNL/GridKit/blob/d8128ce4c6c3a925ac092f5207e1a3d9d568ff19/examples/PowerElectronics/Microgrid/Microgrid.cpp#L462).


Steven-Roberts:
All the tests are finally passing, and it should be ready for review. The examples using Ipopt require tighter tolerances with these changes. Otherwise the optimizations were not converging, seemingly because the objective function was "noisy" at the level of the IDA tolerances. I don't fully understand why this wasn't the case before, but it could be related to this branch introducing solves for consistent initial conditions. If you have any other explaination @pelesh, I'd be curious to know.

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -886,6 +852,14 @@ namespace AnalysisManager
     template <class ScalarT, typename IdxT>
     void Ida<ScalarT, IdxT>::copyVec(const N_Vector x, std::vector<ScalarT>& y)
     {
+      const auto xsize = static_cast<size_t>(N_VGetLength(x));
+      if (xsize != y.size())
+      {
+        std::cerr << "\nN_Vector size (" << xsize << ") does not match std::vector size ("
+                  << y.size() << ").\n\n";
+        throw SundialsException();
+      }
```
pelesh, line 861: Probably we should use GridKit output function, which is xSDK compliant and avoid throwing exceptions.

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -899,6 +873,14 @@ namespace AnalysisManager
     template <class ScalarT, typename IdxT>
     void Ida<ScalarT, IdxT>::copyVec(const std::vector<ScalarT>& x, N_Vector y)
     {
+      const auto ysize = static_cast<size_t>(N_VGetLength(y));
+      if (x.size() != ysize)
+      {
+        std::cerr << "\nstd::vector size (" << x.size() << ") does not match N_Vector size ("
+                  << ysize << ").\n\n";
+        throw SundialsException();
+      }
```
pelesh, line 882: On the other hand, these copy functions will be removed soon, so it doesn't matter that much :).

pelesh (approved):
Great work! Thank you @Steven-Roberts.

PR #364, opened 2026-04-07
LoadZIP Jacobians for new formulations and tests

nkoukpaizan:
## Description
 
This updates the Jacobian terms computed for LoadZIP and adds Jacobian tests.

 ## Proposed changes
- Replaced the computation of $\frac{\partial h}{\partial w_b}$ with $\frac{\partial h}{\partial y}$, $\frac{\partial f}{\partial y}$ and $\frac{\partial f}{\partial w_b}$  for the modified model implementation
- Added a Jacobian test comparing Enzyme Jacobians and DependencyTracking Jacobians
- Added an initialization for internal variables
- Bug fix for the GovernorTest, where the Jacobian entries were not deduplicated

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
- <s>The residual test fails after activating the `LoadZIPTests`. Either the model implementation or the answer keys are incorrect.</s> @abirchfield Please check.
- <s> The $\frac{\partial f}{\partial y}$ is failing to build. This is because the internal variables are all algebraic, so $\frac{\partial f}{\partial y_p}$ is null. I need to implement a workaround for this. </s> Implemented a workaround that I don't particularly like. We can discuss more offline.

### DIFF
--- a/GridKit/AutomaticDifferentiation/Enzyme/DfDy.hpp
+++ b/GridKit/AutomaticDifferentiation/Enzyme/DfDy.hpp
@@ -41,6 +41,79 @@ namespace GridKit
          * @param[in] y - Internal variables
          * @param[in] yp - Internal variable derivatives
          * @param[in] wb - Bus variables
+         * @param[in,out] jac - Jacobian
+         */
+        static void eval(ModelT*     model,
+                         size_t      n_res,
+                         size_t      n_var,
+                         const IdxT* res_indices,
+                         const IdxT* var_indices,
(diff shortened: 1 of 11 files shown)


pelesh:
CC @PhilipFackler

pelesh (approved):
I believe residual tests were failing because of incorrect answer key. I updated test answers with correct numbers and pushed to this branch.

I also tweaked model parameters so that unit test answers are integer numbers for easier comparisons.

@abirchfield, please check and merge into your branch.

Issue #353, opened 2026-03-31
Separate adding components from adding buses in PowerElectronics system assembly.


pelesh:
#### Suggested Change

In class `PowerElectronicsModel` when assembling the system and evaluating residual vectors and Jacobian matrices use separate loops to add contributions from components and connection nodes (buses).
 
#### Definitions
 
- Components are objects that:
    - own and evaluate their internal equations
    - contribute to the coupling equations
    - may own their internal but not coupling variables
- Coupling nodes (buses) are objects that:
    - own coupling equations and coupling variables
    - may contribute to coupling equations

#### Current State

- Internal and external variables are updated from the system solver class using `distribute Vectors` method called by the system model.
- Components have coupling variables tagged with values in `extern_indices_`, which is a `std::set`. It would give us more flexibility if components own arrays with mapping indices for internal and external (coupling) variables instead.
-  There are no mapping vectors for component residual values.

### Rationale

- Internal component equations can be updated by embarrassingly parallel computations. Separating components from buses helps expose this parallelism.
- Coupling variables are owned by coupling nodes (buses). Exposing these, provides better access to the variables that need to be communicated.
- Separating component's residual and Jacobian into internal part and contributions to the coupling (external) part will likely simplify implementation of co-simulation and multimethod integration methods.
- Mapping indices to variable and residual vectors, as well as Jacobians should be set in allocate function. That way we can preconfigure sparsity of the Jacobian to have favorable pattern for matrix factorization.
- Components owning mapping indices will allow us to remove local data copies in components and let components  operate on system vectors and Jacobian.

### Description

- Consider making function `distributeVectors` an internal method to component classes (and select better descriptive name for it). Upon allocation, each component should have a mapping vector with indices to the variable locations in global variable vector. Probably a good idea to have method `updateVariables` that would be called inside component's `evaluateResidual` and `evaluateJacobian` methods.
- Consider implementing separate methods to update internal component variables and coupling variables.
- Create separate component methods to update internal equations residuals and to add contributions to the coupling equations. Have each component own a mapping vector to the global residual vector.
- Change class name `PowerElectronicsModel` --> `SystemModelPowerElectronics`.

### Additional information

Also, see #339.

CC @nkoukpaizan @PhilipFackler

alexander-novo:
> - Components have coupling variables tagged with values in extern_indices_, which is a std::set. It would give us more flexibility if components own arrays with mapping indices for internal and external (coupling) variables instead.
[...]

PR #367, opened 2026-04-09
Add node objects to PowerElectronics module

alexander-novo:
## Description
 
Partially addresses #353
 
Adds node objects to system so the system can calculate the optimal variable mappings itself. Right now, these node objects consist of:

- `Bus` (single voltage bus)
- `MicrogridBus` (two voltage bus)
- `InfiniteBus` (single voltage reference bus)
- `SignalNode` (single-variable signal node)

All of the `PowerElectronics` examples have been updated to use these nodes, and there original logic for creating variable mappings has been removed. Since the `RLCircuit` example didn't have the correct mapping, some of its variables have been re-arranged.

As well, all component models included in these examples have been updated. Other component models can also be updated, since I feel fairly confident that I should be able to update them correctly... I just don't have tests to verify.
 
 ## Proposed changes
 
Node objects work fairly similarly to components:
- they have an internal size, which contributes to the overall system size
- they have an external size (for e.g. reference nodes)
- they have a mapping from these variables to system variables

The system adds up the internal sizes for all of the nodes and components to get its own internal size, and then is able to form mappings for all of the node and component internal variables - putting the node variables at the end of the system vector for correct Jacobian sparsity.

Components with external variables take node pointers in their constructor, and in the `allocate()` function will query the mapping of the node variables to properly construct the mapping for their own external variables. Reference variable mappings are also set here.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments


examples/PowerElectronics/Microgrid/Microgrid.cpp
```diff
@@ -84,213 +85,88 @@ int main(int /* argc */, char const** /* argv */)
...
-  size_t vec_size_externals = 1 + 2 * (2 * Nsize);
-  size_t dqbus1             = vec_size_internals + 1;
-  size_t dqbus2             = vec_size_internals + 3;
-  size_t dqbus3             = vec_size_internals + 5;
-  size_t dqbus4             = vec_size_internals + 7;
+  using SignalNode                      = GridKit::PowerElectronics::SignalNode<double, size_t>;
+  std::unique_ptr<SignalNode> dg_signal = std::make_unique<SignalNode>();
 
-  size_t vec_size_total = vec_size_internals + vec_size_externals;
+  sysmodel->addNode(&*dg_signal);
```
pelesh: I suggest removing this syntax. It is cumbersome and confusing.

examples/PowerElectronics/Microgrid/Microgrid.cpp
```diff
@@ -84,213 +85,88 @@ int main(int /* argc */, char const** /* argv */)
...
-  dg1->setExternalConnectionNodes(2, dqbus1 + 1);
-  //"grounding" of the difference
-  dg1->setExternalConnectionNodes(3, static_cast<size_t>(-1));
-  // internal connections
-  for (size_t i = 0; i < 12; i++)
-  {
+  sysmodel->addNode(&*bus1);
+  sysmodel->addNode(&*bus2);
+  sysmodel->addNode(&*bus3);
+  sysmodel->addNode(&*bus4);
```
pelesh: Similar here, there is no need for `&*`.

examples/PowerElectronics/RLCircuit/RLCircuit.cpp
```diff
@@ -31,58 +34,46 @@ int main(int /* argc */, char const** /* argv */)
   double linit = 1.0;
   double vinit = 1.0;
 
+  using Bus                           = GridKit::PowerElectronics::Bus<double, size_t>;
+  using InfiniteBus                   = GridKit::PowerElectronics::InfiniteBus<double, size_t>;
+  std::unique_ptr<InfiniteBus> bus_iv = std::make_unique<InfiniteBus>(0.0);
+  std::unique_ptr<Bus>         bus_vr = std::make_unique<Bus>();
+  std::unique_ptr<Bus>         bus_ir = std::make_unique<Bus>();
+
+  sysmodel.addNode(&*bus_iv);
```
pelesh: How about this:
```suggestion
  Bus bus_ir;

  sysmodel.addNode(&bus_ir);
```

pelesh (changes requested):
Greta work!

Before we merge, I suggest following minor changes:
- Remove infinite bus. We should not be using that approach in EMT simulations.
- Signal note should not inherit from the same base class as electrical bus.
- Address some nitpicking comments below.
- Rebase to `develop`.

alexander-novo:
Renaming infinite bus to grounded bus.

alexander-novo:
@pelesh Right now we don't have any examples of signal nodes being used differently than buses. I can make it so that `SignalNode` doesn't inherit from `NodeBase` but then I'd just have to duplicate the code in `NodeBase` for `SignalNode`. Maybe it'd be better to punt this down the road for when we have an actual example that uses the functionality?

examples/PowerElectronics/ScaleMicrogrid/ScaleMicrogrid.cpp
```diff
@@ -303,7 +257,7 @@ int test(index_type Nsize, real_type error_tol, bool debug_output)
   }
 
   // since the intial P_com = 0, the set the intial vector to the reference frame
-  sys_model->y()[vec_size_internals] = DG_parms1.wb_;
+  sys_model->y()[dg_signal.getNodeConnection(0)] = DG_parms1.wb_;
```
pelesh, line 260: Why is signal connection here? I am not sure if we have any signal connectors in this model.

examples/PowerElectronics/ScaleMicrogrid/ScaleMicrogridArbitrary.cpp
```diff
@@ -288,7 +244,7 @@ int printMicrogridSystems(index_type N_size)
   }
 
   // since the initial P_com = 0, set the initial vector to the reference frame
-  sys_model.y()[vec_size_internals] = DG_parms1.wb_;
+  sys_model.y()[dg_signal.getNodeConnection(0)] = DG_parms1.wb_;
```
pelesh, line 248: I suspect this is a misuse of the signal connector.

pelesh (approved):
I am not sure I understand the use of signal node in your examples but I think it is a minor issue we could discuss separately.

examples/PowerElectronics/ScaleMicrogrid/ScaleMicrogrid.cpp
alexander-novo: Yes, there is one signal node in this model - the reference rotor speed omega. All lines and DGs are connected to this signal node, and the reference DG writes to it.

pelesh: Thanks for explanation. Let's see how it works for us.

PR #365, opened 2026-04-08
Stabilizer Docs + IEEEST Implementation

lukelowry:
## Description
 
Cleaned the Stabilizer folder in the PhasorDynamics to align it with the other models' folder structures. The documentation and implementation, or IEEEST, was also added.

This is one of the last essential models we need to faithfully model some of the larger cases. It's best to implement this before creating rigorous tests for big cases.
 
 ## Proposed changes
 
- Make a general Stabilizer README
- Subfolder `Stabilizer/PSS1A` for the existing doc has been partially cleaned but not validated
- Subfolder `Stabilizer/IEEEST` with docs which should be the first stabilizer implemented.
- IEEEST Implementation 
- The exciter and stabilizer can now be wired through the new external variable in `IEEET1`, `vs`.
- Lots of potentially unsafe parameters that are still valid model definitions, so I used a masking method that works with Enzyme. Not sure if it's optimal, but it works with a 240 bus case.
 
 ## Checklist
 
- [x] All tests pass. ( 44 - EnzymePowerElectronicsCheck is failing but that happens for me on development as well)
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [x] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 
 ## Further comments
 
I have an uncommitted `GENROU` saturation fix and WECC 240 case that are pending this model!
 
The input signal delay is not implemented and is discussed in #369 . not a blocker

PR #373, opened 2026-04-12
SEXS-PTI Exciter Model + GENROU Saturation

lukelowry:
## Description
 
SEXS-PTI Model added; needed for some of the team's requested cases. Also, to validate the cases that depend on this PR, I need to add a temporary initialization routine for saturation. It is not how we should do this, but I simply need a steady-state solution to verify that the cases in question provide meaningful data or can be simulated at all.

Relies on the changes in the exciter and stabilizer input in #365, so PR is to that branch.
 
 ## Proposed changes
 
- SEXS-PTI Exciter Model
- Temporary iteration in the initialization of GENROU to initialize nonlinear saturation for the steady state.
- Consistent documentation for windup limits, which are now used identically in three models
- Dedicated tests for windup/indicator function
 
 ## Checklist
 
- [X] All tests pass. **(There is One preexisting error that is happpening in the PowerElectronics module!)**
 
 ## Further comments
 
Some of the equations regardining the smooth approximation might need to be updated.

This model is simple and widely used. Let's include this and its data format changes, so we don't need to change the cases as much.


lukelowry:
I thought I set this up so that once #365 was merged, this would point towards the develop branch, but that appears not to have happened. Otherwise ready

nkoukpaizan:
@lukelowry Could you take a look at the failing test? Looks like something flagged by the address sanitizer.

lukelowry:
@nkoukpaizan Thanks for pointing that out. It was an issue with the signals on the new model. All good now, I believe.

abirchfield (approved):

GridKit/CommonMath.hpp
```diff
@@ -119,10 +119,12 @@ namespace GridKit
...
-      return indicator_zero(limit_min, limit_max, f) * dsigmoid(f) + //
-             (indicator_low(limit_min, x) * sigmoid(-f) +            //
-              indicator_high(limit_max, x) * sigmoid(f))
-                 * (ONE<RealT> - dsigmoid(f));
+      ScalarT above_min = indicator_low(limit_min, x);
+      ScalarT below_max = indicator_high(limit_max, x);
+
+      return above_min * below_max +                  //
+             (ONE<RealT> - below_max) * sigmoid(-f) + //
+             (ONE<RealT> - above_min) * sigmoid(f);
```
nkoukpaizan, line 128: Was there a particular reason for switching this back? To my recollection, there was a discontinuity when `f=0`. In any case, you might want to update the related README files if you change the code here.

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
```diff
@@ -0,0 +1,308 @@
...
+        exciter.y()[1] = 5.05;
+        exciter.y()[2] = 0.0;
+        exciter.evaluateResidual();
+        auto small_push_above_limit  = std::abs(exciter.getResidual()[1]);
+        success                     *= (small_push_above_limit < static_cast<ScalarT>(1.0e-5));
+
+        return success.report(__func__);
+      }
+
+      TestOutcome smoothIndicator()
```
nkoukpaizan: I would move a test for the indicator to its own test suite. Something like `tests/UnitTests/Math/SmoothnessIndicatorTests.hpp`.

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
```diff
@@ -0,0 +1,308 @@
...
+        auto blocked = std::abs(exciter.getResidual()[1]);
+
+        exciter.y()[0] = 1.0;
+        exciter.y()[1] = 10.0;
+        exciter.y()[2] = 0.0;
+        exciter.evaluateResidual();
+        auto returning = std::abs(exciter.getResidual()[1]);
+
+        success *= (blocked < static_cast<ScalarT>(1.0e-8));
+        success *= (returning > static_cast<ScalarT>(1.0));
```
nkoukpaizan: I suggest adding a few comments here. It's ot obvious to me why these `<` and `>` values are expected.

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
```diff
@@ -0,0 +1,308 @@
...
+        exciter.getSignals().template assignSignalNode<PhasorDynamics::Exciter::SexsPtiInternalVariables::EFD>(&efd_node);
+        exciter.getSignals().template attachSignalNode<PhasorDynamics::Exciter::SexsPtiExternalVariables::VS>(&vs_node);
+
+        exciter.allocate();
+        efd_node.init(1.2);
+        success *= (exciter.verify() == 0);
+        exciter.initialize();
+        exciter.evaluateResidual();
+
+        success *= isEqual(exciter.getResidual()[2], vs_value, static_cast<ScalarT>(1.0e-12));
```
nkoukpaizan: Do we not expect these to match to machine precision? Perhaps add a few comments if machine precision is not expected. There are a few instances of looser tolerances in this file that it would be good to clarify.

nkoukpaizan (commented):
This mostly looks good to me. There are a couple of things I'd like to see addressed before merging.

GridKit/CommonMath.hpp
lukelowry: It is not the same as the original, though it looks similar; this one is smooth. The proposed method is faster for AD and is smoother at the bounds. I attached the images below to show how the current implementation compares. Let me know what you think. 

indicator_compare
indicator_structure

nkoukpaizan: Oh, I see there's an additional oscillation when trying to clip the value to 1 when `f=0`. Good catch!

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
lukelowry: I added a kTol parameter and added some clarifying notes. I think I resolved your concern, lmk!

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
lukelowry: See new test file

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
lukelowry: Added! I also made a central doc for CommonMath (which probably should be its own folder in the future) so we don't have to maintain three docs when we adjust the smooth windup approximation. This will be used in countless other models. good suggestion

lukelowry:
I did some further doc cleanup and consistency chores. It's ready for your review again! @nkoukpaizan

nkoukpaizan (approved):
Looks good to me. Nice work!

PR #377, opened 2026-04-20
BranchBase Class for Other Branch Types

lukelowry:
## Description
 
Makes a `BranchBase` component family class so we can share boilerplate and equations for network construction across different transformer and line models.

There is a related open issue that I cannot find, but in an ideal world, the branch should own the line's current states and the residual injection equations. The Bus should own the voltage states and have one continuity equation (two in phasor dynamics, for real and imag):

Bus (owns voltage states)
```math
0=\sum_{k=0}^K  i_{mk}
```
Branch (From and To Currents) for pi model:
```math
\begin{aligned}
0 &=i_{mk} - v_{k} * y_{mk}/2 \\ 
0 &=i_{km}- v_{m} * y_{mk}/2
\end{aligned}
```
That is the next step after this PR, changing the ownership of the current is a big change.

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
None


PhilipFackler (changes requested):
I think we should reduce the scope of this PR. The new base class with the factory and JSON parser should be included here. That can be done without changing the syntax for using the branch. That is, the changes relating to "bus" vs "port" and how V and I are accessed should be in a separate PR. They deserve their own discussion.

EDIT: I should add that I like the architectural changes :)

pelesh (changes requested):
I agree that we need to manage the scope of this PR a little better. I suggest creating an issue to discuss the proposed changes there first, followed by documentation-only PR.

I would also suggest creating a transformer model from branch using copy-paste-tweak approach. That would give us needed component for system modeling and would also provide better clues how a base class branch should look like.

Architectural changes and new models are different topics and should be handled in separate issues/PRs imho.

lukelowry:
@pelesh @PhilipFackler thank you both for the feedback. I will close this PR I agree with the scope issue. I will start with a more practical implementation with the Transformer instead of working backward. I hope you see the vision, however!

lukelowry:
Close

PR #378, opened 2026-04-22
Reorder DistributedGenerator variables

alexander-novo:
## Description
 
Re-ordered the internal variables of `DistributedGenerator` to place the angle delta at the end of the internal variables. This is a prerequisite for removing the copy to components vectors, since the indexing of internal variables currently differs between the references and non-reference generators.

This PR changes it so all internal variables (besides the angle delta) have the same index between reference and non-reference generators.
 
 ## Proposed changes

 
 ## Checklist
 
- [x] All tests pass.

I will include this in the changelog update for removing the data copying (coming next)
 
 ## Further comments


GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
```diff
@@ -43,17 +43,17 @@ namespace GridKit
...
     n_extern_ = refframe_ ? 4 : 3;
     idc_      = id;
-    nnz_      = refframe_ ? 80 : 78;
+    nnz_      = refframe_ ? 77 : 78;
 
     if (refframe_)
     {
-      extern_indices_ = {0, 1, 2, 3};
+      extern_indices_ = {0, 1, 2, 15};
     }
```
pelesh, line 57: Looking at all magic number here, it would be good replace them with a well-commented enum(s). probably ouyt of scope for this PR.

pelesh (approved):
Looks good.  Go ahead and merge to the target branch.

GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
alexander-novo: @pelesh This `extern_indices_` variable is likely to be removed once external variables and internal variables are indexed separately.

alexander-novo:
@pelesh Rebased and pointing at `develop` now

PR #379, opened 2026-04-23
Remove data copying for internal component variables, derivatives, and residuals in PowerElectronics

alexander-novo:
## Description
 
Partially addresses #353.

Removes data copying between the system and components for internal variables, derivatives, and residuals when evaluating residuals and Jacobians.
 
 ## Proposed changes
 
`CircuitComponent` now has three pointers:
- `y_int_`: a pointer to the component's internal variables in the system vector
- `yp_int_`: a pointer to the component's internal derivatives in the system vector
- `f_int_`: a pointer to the component's internal residuals in the system vector

`PowerElectronicsModel::distributeVectors()` and `PowerElectronicsModel::evaluateResidual()` have been modified to only copy variables between `y_` and `f_` if they are external. Therefore, components must use the three new pointers above to correctly calculate residuals and Jacobians.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
This solution won't work quite right for system objects which are components in another system. Maybe we want some sort of flag that checks for this situation and doesn't allocate a `y_` or `f_` vector?

### DIFF
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -168,17 +168,17 @@ namespace GridKit
 
     void setInternalPointer(const ScalarT* internals)
     {
-      int_ = internals;
+      y_int_ = internals;
     }
 
     void setInternalDerivativePointer(const ScalarT* internals_p)
     {
-      intp_ = internals_p;
+      yp_int_ = internals_p;
     }
 
     void setInternalResidualPointer(ScalarT* internal_res)
     {
-      int_f_ = internal_res;
+      f_int_ = internal_res;
     }
 
   protected:
@@ -428,11 +428,11 @@ namespace GridKit
     size_t current_jac_size_{0};
 
     /// @brief A pointer to the internal variables of this component.
-    const ScalarT* int_;
+    const ScalarT* y_int_;
     /// @brief A pointer to the internal derivatives of this component.
-    const ScalarT* intp_;
+    const ScalarT* yp_int_;
     /// @brief A pointer to the internal residuals of this component
-    ScalarT*       int_f_;
+    ScalarT*       f_int_;
 
     std::vector<ScalarT> y_;
     std::vector<ScalarT> yp_;
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -427,8 +427,11 @@ namespace GridKit
     /// The number of non-zero elements currently inserted into the Jacobian. See \ref setJacValues()
     size_t current_jac_size_{0};
 
+    /// @brief A pointer to the internal variables of this component.
     const ScalarT* int_;
+    /// @brief A pointer to the internal derivatives of this component.
     const ScalarT* intp_;
+    /// @brief A pointer to the internal residuals of this component
     ScalarT*       int_f_;
 
     std::vector<ScalarT> y_;
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -166,6 +166,21 @@ namespace GridKit
      */
     virtual int evaluateExternalResidual() = 0;
 
+    void setInternalPointer(const ScalarT* internals)
+    {
+      int_ = internals;
+    }
+
+    void setInternalDerivativePointer(const ScalarT* internals_p)
+    {
+      intp_ = internals_p;
+    }
+
+    void setInternalResidualPointer(ScalarT* internal_res)
+    {
+      int_f_ = internal_res;
+    }
+
   protected:
     /**
      * @brief Reset the Jacobian so it can be constructed. Helper method for \ref setJacValues().
@@ -411,6 +426,10 @@ namespace GridKit
 
     /// The number of non-zero elements currently inserted into the Jacobian. See \ref setJacValues()
     size_t current_jac_size_{0};
+
+    const ScalarT* int_;
+    const ScalarT* intp_;
+    ScalarT*       int_f_;
 
     std::vector<ScalarT> y_;
     std::vector<ScalarT> yp_;
--- a/GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
+++ b/GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
@@ -84,8 +84,8 @@ namespace GridKit
   template <class ScalarT, typename IdxT>
   int DistributedGenerator<ScalarT, IdxT>::evaluateInternalResidual()
   {
-    ScalarT omega = wb_ - mp_ * int_[0];
-    ScalarT delta = refframe_ ? ScalarT(0.0) : int_[12];
+    ScalarT omega = wb_ - mp_ * y_int_[0];
+    ScalarT delta = refframe_ ? ScalarT(0.0) : y_int_[12];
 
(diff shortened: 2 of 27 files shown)


alexander-novo:
@pelesh I'm not really sure what to do about the enzyme example, since the DG model's residual has diverged from what's given in that example. Let me know what you think is best.

pelesh:
> @pelesh I'm not really sure what to do about the enzyme example, since the DG model's residual has diverged from what's given in that example. Let me know what you think is best.

Let's merge the other power electronics PRs, first. 

My understanding is that only variables were reordered and equations remained the same. In that case, it should not be difficult to fix the answer key for the Enzyme test.

pelesh:
I set this to draft for now until we merge other related PRs.

alexander-novo:
@pelesh I'm not sure the `DistributedGeneratorTest` makes sense. The test expects to evaluate the residual of a reference DG where the delta variable is non-zero, which is now no longer possible and I don't think should be a supported feature of the model?

alexander-novo:
Well I just changed it to be a test of a non-reference generator, and edited the answer key to match.

pelesh:
This is very nice work for what I could see thus far. I would only avoid calling internal variables `int_` becayuse in C-speak "int" is widely recognized as "integer" and `int_` is too close for comfort to the keyword `int`.

How about `y_int_`?

alexander-novo:
@pelesh Rebased and ready for review.

GridKit/Model/PowerElectronics/InductionMotor/InductionMotor.cpp
```diff
@@ -70,21 +70,21 @@ namespace GridKit
...
-    f_[6] = (1.0 / std::sqrt(3.0)) * (-y_[1] + y_[2]) - Rs_ * y_[6] - (Lls_ + Lms_) * yp_[6] - Lms_ * yp_[5];
-    f_[7] = (y_[0] + y_[1] + y_[2]) / 3.0 - Rs_ * y_[7] - Lls_ * yp_[7];
-    f_[8] = Rr_ * y_[8] + (Llr_ + Lms_) * yp_[8] + Lms_ * yp_[5] - (P_ / 2.0) * y_[3] * ((Llr_ + Lms_) * y_[9] + Lms_ * y_[6]);
-    f_[9] = Rr_ * y_[9] + (Llr_ + Lms_) * yp_[9] + Lms_ * yp_[6] + (P_ / 2.0) * y_[3] * ((Llr_ + Lms_) * y_[8] + Lms_ * y_[5]);
+    f_int_[0] = (1.0 / 3.0) * (2.0 * y_[0] - y_[1] - y_[2]) - Rs_ * y_int_[0] - (Lls_ + Lms_) * yp_int_[0] - Lms_ * yp_int_[1];
+    f_int_[1] = (1.0 / std::sqrt(3.0)) * (-y_[1] + y_[2]) - Rs_ * y_int_[1] - (Lls_ + Lms_) * yp_int_[1] - Lms_ * yp_int_[0];
+    f_int_[2] = (y_[0] + y_[1] + y_[2]) / 3.0 - Rs_ * y_int_[2] - Lls_ * yp_int_[7];
+    f_int_[3] = Rr_ * y_int_[3] + (Llr_ + Lms_) * yp_int_[8] + Lms_ * yp_int_[0] - (P_ / 2.0) * y_[3] * ((Llr_ + Lms_) * y_int_[4] + Lms_ * y_int_[1]);
+    f_int_[4] = Rr_ * y_int_[4] + (Llr_ + Lms_) * yp_int_[9] + Lms_ * yp_int_[1] + (P_ / 2.0) * y_[3] * ((Llr_ + Lms_) * y_int_[3] + Lms_ * y_int_[0]);
     return 0;
```
pelesh, line 78: Question: Should we use enums for residual and solution vector indexing? This is probably out of scope of this PR but worth considering for the future development.

CC @PhilipFackler @nkoukpaizan

pelesh (approved):
Tests pass and the redundant data copying was removed.

Some nitpicking comments that we might want to address:
- Using "int" to describe internal variables may be confusing for some developers as "int" usually is understood as short for integer (and I realize `y_int_` was my suggestion :)).
- There are too many "magic numbers" in power electronics component models. This may be out of scope of this PR but we do need to address this at some point.

GridKit/Model/PowerElectronics/CircuitComponent.hpp
nkoukpaizan, line 155: Would be good to update the documentation for `evaluateInternalResidual` to reference `f_int_` (or whatever name we settle on).

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -206,11 +212,18 @@ namespace GridKit
       }
 
       {
+        // The offset for each component's internal variables in the system vector.
+        // They start at 0, and are stacked of each other.
```
nkoukpaizan:
```suggestion
        // They start at 0, and are stacked on top of each other.
```

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -206,11 +212,18 @@ namespace GridKit
...
         size_t component_internal_idx = 0;
         for (component_type* comp : components_)
         {
           comp->allocate();
 
+          // Update component internal pointers to their correct offsets
+          comp->setInternalPointer(&y_[component_internal_idx]);
```
nkoukpaizan, line 223: I would use the more verbose `component` (versus `comp`).

GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
```diff
@@ -153,8 +148,8 @@ namespace GridKit
 
     // output
     // current transformed to common frame
-    f_[1] = std::cos(y_[15]) * y_[13] - std::sin(y_[15]) * y_[14];
-    f_[2] = std::sin(y_[15]) * y_[13] + std::cos(y_[15]) * y_[14];
+    f_[1] = std::cos(delta) * y_int_[10] - std::sin(delta) * y_int_[11];
+    f_[2] = std::sin(delta) * y_int_[10] + std::cos(delta) * y_int_[11];
```
nkoukpaizan, line 152: Would be good to further document the current structure of `f_`. If I understand correctly, it is still allocated to `size_ = n_intern_ + n_extern_`, but only the external part (first `n_extern_` values) gets updated at the component level.

nkoukpaizan (approved):
A few comments, mostly about clarifying documentation. Otherwise, looks good.

nkoukpaizan:
> * Using "int" to describe internal variables may be confusing for some developers as "int" usually is understood as short for integer (and I realize `y_int_` was my suggestion :)).

One suggestion I have is to use the verbose `y_internal_` for the member variable. The downside is that this is too long for the residual evaluation terms, where we may want to use a local alias.

GridKit/Model/PowerElectronics/CircuitComponent.hpp
pelesh: To be addressed in #416.

GridKit/Model/PowerElectronics/DistributedGenerator/DistributedGenerator.cpp
pelesh: To be addressed in #416.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
pelesh: To be addressed in #416.

PR #372, opened 2026-04-11
State Sync + Simulation Performance

lukelowry:
## Description
 
In my feature branch, where I have larger models with more signals, I was running into some strange state synchronization issues between the machine, exciter, and bus. I noticed that `Ida.cpp`and `SystemModel.hpp` seem to pass state in a different order, fixing the inter-model issues I had. Ida stopped failing during fault events in a 240-Bus case I am constructing, and it reduced the solve time by ~3x across all examples.

I'm not sure why this has such a big impact on performance, or whether it's related to my build configuration. Worth looking at because GK now simulates faster than PowerWorld for my machine on this branch, but not on the `develop` branch.

Example of change

*Before*
```
  model->updateTime(tres, 0.0);
  copyVec(yy, model->y());
  copyVec(yp, model->yp());
```

*After* (i.e., share state before invoking so it's not stale)
```
  copyVec(yy, model->y());
  copyVec(yp, model->yp());
  model->updateTime(tres, 0.0);
```
 
 _Closes #(issue)_
 
 _Mentions @(user)_
 
 ## Proposed changes
 
 _Describe how your changes here address the issue and why the proposed changes
 should be accepted._
 
 ## Checklist
 
 _Put an `x` in the boxes that apply. You can also fill these out after creating
 the PR. If you're unsure about any of them, don't hesitate to ask. We're here
 to help! This is simply a reminder of what we are going to look for before
 merging your code._
 
- [ ] All tests pass.
- [ ] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [ ] The new code follows GridKit™ style guidelines.
- [ ] There are unit tests for the new code.
- [ ] The new code is documented.
- [ ] The feature branch is rebased with respect to the target branch.
- [ ] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
 _If this is a relatively large or complex change, kick off the discussion by explaining
 why you chose the solution you did and what alternatives you considered, etc..._

PR #380, opened 2026-04-23
Make `updateTime` calls consistently *after* data updates

PhilipFackler:
## Description
 
Make updating `SystemModel` data from `Ida` consistent.
 
 ## Proposed changes
 
Calling `updateTime` after calls to `copyVec` ensures that constituent models always have the most up-to-date data before doing more with their copies.

This will also make the transition consistent when we move to constituents having (non-owning) references to system arrays.
 
 ## Checklist
 
- [X] All tests pass.

### DIFF
--- a/GridKit/Solver/Dynamic/Ida.cpp
+++ b/GridKit/Solver/Dynamic/Ida.cpp
@@ -473,9 +473,9 @@ namespace AnalysisManager
       }
 
       // Final copy out. No gaurentee last residual evaluation is final step.
-      model_->updateTime(tf, 0.0);
       copyVec(yy_, model_->y());
       copyVec(yp_, model_->yp());
+      model_->updateTime(tf, 0.0);
 
       return retval;
     }
@@ -669,9 +669,9 @@ namespace AnalysisManager
       }
 
       // Final copy out. No gaurentee last residual evaluation is final step.
-      model_->updateTime(tf, 0.0);
       copyVec(yy_, model_->y());
       copyVec(yp_, model_->yp());
+      model_->updateTime(tf, 0.0);
 
       return retval;
     }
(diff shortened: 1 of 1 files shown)


lukelowry:
Yes this dramtically improves performance which I attempted to describe in #372

Steven-Roberts:
Out of curiosity, do you know why this reordering makes a difference? From a scan of the `copyVec` and `updateTime` functions, it looks like they read from and write to disjoint sets of variables, so I'd expect the order not to matter.

lukelowry:
> Out of curiosity, do you know why this reordering makes a difference? From a scan of the `copyVec` and `updateTime` functions, it looks like they read from and write to disjoint sets of variables, so I'd expect the order not to matter.

I have a suspicion that it has to do with signals between models and how that impacts residual calculations. @PhilipFackler might have better insight, but I have no better explanation

nkoukpaizan:
> > Out of curiosity, do you know why this reordering makes a difference? From a scan of the `copyVec` and `updateTime` functions, it looks like they read from and write to disjoint sets of variables, so I'd expect the order not to matter.
> 
> I have a suspicion that it has to do with signals between models and how that impacts residual calculations. @PhilipFackler might have better insight, but I have no better explanation

There is a call to `updateVariables` within `updateTime` to distribute to components. [See](https://github.com/ORNL/GridKit/blob/develop/GridKit/Model/PhasorDynamics/SystemModel.hpp#L862). We might want to consider moving that `updateVariables` to a more intuitive location. The proposed change will do for now!

nkoukpaizan (approved):

PhilipFackler:
> > > Out of curiosity, do you know why this reordering makes a difference? From a scan of the `copyVec` and `updateTime` functions, it looks like they read from and write to disjoint sets of variables, so I'd expect the order not to matter.
> > 
> > 
> > I have a suspicion that it has to do with signals between models and how that impacts residual calculations. @PhilipFackler might have better insight, but I have no better explanation
> 
> There is a call to `updateVariables` within `updateTime` to distribute to components. [See](https://github.com/ORNL/GridKit/blob/develop/GridKit/Model/PhasorDynamics/SystemModel.hpp#L862). We might want to consider moving that `updateVariables` to a more intuitive location. The proposed change will do for now!

Once we replace copies of solution vectors with references in constituent models, we won't need `updateVariables` at all. :)

PR #374, opened 2026-04-12
Illinois 200 Bus + WECC 240 Bus + IEEE ISONE

lukelowry:
## Description
 
Had to make one small numerical conditioning change to `SEXS-PTI` Exciter after rebase.

Adds the Illinois 200 bus case and WECC 240 bus case, and finishes the implementation and documentation of the IEEE New England ISO 39 bus case. Depends on #373 deeply for verification and case data parsing.

I should note that at this point, GridKit is dramatically outperforming PowerWorld. In some instances, it solved nearly `4x` faster.

Closes #357 
 
 
 ## Proposed changes
 
- WECC 240 Case and solver file
- Illinois 200 Case and solver file
- IEEE 39 Bus finalize
- Polish Texas case docs and examples
- Polish Hawaii case docs and examples
- Small tweak to `IEEEST` to handle some zero-valued parameters in stabilizer I did not account for originally.
 
 ## Checklist
 
 
- [X] All tests pass. (except on preexisting power electronics issue)
- [X] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [X] The new code follows GridKit™ style guidelines.
- [X] There are unit tests for the new code.
- [X] The new code is documented.
- [X] The feature branch is rebased with respect to the target branch.
- [ ] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
All 5 models start in a steady state now.

I'm waiting on the case comparison CLI args or generally better results comparison be- but visual inspection against the power world simulation confirms the two new models and the cases are being simulated correctly:

## Illinois Generator Speed

Illinois_GridKitGenSpeed
Illinois_PowerWorldGenSpeed

## New England Generator Speed

PowerWorldNE
GridKitNE

## WECC 

WECC_GRIDKIT
WECC_POWERWORLD_V
WECC_

## Texas

texas_results

## Hawaii

hawaii_results

PR #382, opened 2026-04-25
Event Cues + Branch Trips for more Diverse Simulations

lukelowry:
## Description

 After some reflection, I think it is wise to model events as a typed `Action` performed on a `Component`, much like `BusFault` is implemented now.  

- A model of the `Component` class can perform an  `Action`, which is expressed by the implementation of the interface `apply(Action)` .
- The `SystemModel` cues a target `Component` to perform said  `Action`  by invoking  `sys->cue(target, action)`
- Alternatively, a component can be directly invoked via `target->apply(Action)`.
- A `Study` is a timed sequence of `Cue`s and `Component` targets.

For every event I can consider, this is very intuitive (i.e., `Fault` a `Bus`, `Fault` a `Branch`, `Close`/`Open` a `Branch`, `Oscillate` a machine (forced oscillation), `Ramp` a `Signal`, etc.). 
 
 ## Proposed changes
 
- Using this representation, I have implemented a Branch trip action that opens a broad class of studies in Power Systems. 
- `SystemModel` no longer treats `BusFaults` as a special set of components. In the future, we should probably make `Fault` or `Faultable` and interface so that `Bus` and `Branch` can implement, so there is no need for a dedicated `BusFault` component.
- A `README` for `PDSim` which details the proposed `*.study.json` schema. 
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
Nice to have before we merge #374 so that we can do more diverse studies and non-fault events!

A stepping stone toward a more general solution to #377.

### DIFF
--- a/GridKit/Model/PhasorDynamics/Branch/Branch.hpp
+++ b/GridKit/Model/PhasorDynamics/Branch/Branch.hpp
@@ -106,6 +106,16 @@ namespace GridKit
         setDerivedParams();
       }
 
+      /// True when the branch is in service (closed). The underlying
+      /// `status_` is a smooth multiplier (1.0 closed / 0.0 open) so that
+      /// `evaluateBusResidualNN` kernels see a constant expression structure
+      bool status() const
+      {
+        return status_ > RealT{0.5};
+      }
+
+      void apply(Action action) override final;
+
       const Model::VariableMonitorBase* getMonitor() const override;
 
     private:
@@ -166,6 +176,7 @@ namespace GridKit
       RealT     X_{0.0};
       RealT     G_{0.0};
       RealT     B_{0.0};
+      RealT     status_{1.0}; ///< Multiplier on residual contributions: 1.0 = closed, 0.0 = open
       IdxT      bus1_id_{0};
       IdxT      bus2_id_{0};
 
--- a/GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
+++ b/GridKit/Model/PhasorDynamics/Branch/BranchData.hpp
@@ -15,10 +15,11 @@ namespace GridKit
     /// Initial parameters for a branch
     enum class BranchParameters
     {
-      R, ///< Line series resistance
-      X, ///< Line series reactance
-      G, ///< Line shunt conductance
-      B, ///< Line shunt charging
+      R,      ///< Line series resistance
+      X,      ///< Line series reactance
+      G,      ///< Line shunt conductance
+      B,      ///< Line shunt charging
+      state0, ///< Initial branch status (true = closed/in-service, default true)
     };
 
     /// Ports for a branch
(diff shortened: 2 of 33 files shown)


PhilipFackler (changes requested):
This took me a while to think about. I kept thinking there has to be a better way to handle this. But honestly (like you mention yourself) until we have more complicated use-cases, we don't need a more complicated solution. This fits the current need quite well.

I like using the concept of "schedule" and moving toward being able to define actions generically on different components.

I have one true request and one subjective complaint. Respectively:

- The `SystemModel` member `by_id_` needs a more descriptive name.
- I don't like the word "cue" here. 😸  This is just preference, and I can easily live with it if it makes sense to everyone else. I would suggest "trigger" or "execute". 🤷 @pelesh thoughts?

Finally, since you are allowing for a case with multiple events at the same time point, you should document the fact that those must be listed consecutively. Maybe add some logic to check for this (or even sort the events with respect to time...). It would be good to have a helpful error when that mistake is made.

lukelowry:
> * I don't like the word "cue" here. 😸  This is just preference, and I can easily live with it if it makes sense to everyone else. I would suggest "trigger" or "execute". 🤷 @pelesh thoughts?

Not married to `cue` either, tried to be creative, but now I think it's confusing. Some other options:
- Event
- Task
- Directive

We can let @pelesh choose!

> Finally, since you are allowing for a case with multiple events at the same time point, you should document the fact that those must be listed consecutively. Maybe add some logic to check for this (or even sort the events by time). It would be good to have a helpful error when that mistake is made.

Agree, I will do so once @pelesh gives feedback

pelesh:
Having branch faults implemented would be a valuable feature. Thank you, @lukelowry!

As for the earlier questions, my recommendation would be to have two configuration files: model and simulation. Model would describe a grid (and that is already documented and implemented in GridKit), while the simulation file would describe solver settings, simulation time, events, etc.

I think "event" is the most commonly used name for what is described in this PR as "cue".

Given the scope of changes here, my recommendation would be to have documentation-only PR first, and then make changes to the code in a follow on PR. I think we first need to decide _how_ we implement faults and then modify bus faults (if necessary) and add branch faults.

pelesh (commented):
For a more agile review I suggest we split this PR in documentation/specs and implementation PR, respectively.

lukelowry:
> For a more agile review I suggest we split this PR in documentation/specs and implementation PR, respectively.

I can get behind this. I will break this into two ASAP

Also, just to clarify, the event type I added is not a Branch fault (tho that would be easy to add, now). It's a line opening and closing, which is different from a branch fault.

Doing documentation first will help with clarification!

pelesh:
I suggest we set this to draft for now. This idea might need more offline discussion.

lukelowry:
Closing outdated

PR #383, opened 2026-04-28
Add INSTALL.md file.

pelesh:
## Description
 
 Add more detailed installation instructions for GridKit.
 
 
 ## Proposed changes
 
 Add INSTALL.md file with detailed installation instructions and relevant links.
 
 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -231,8 +231,10 @@ cmake --install build
 
 ### Enzyme
 
-Enzyme support is experimental. Only `Release` and `RelWithDebInfo` build
-types are supported due to a known Enzyme limitation.
+Enzyme support is experimental. There is a known bug in Enzyme
+that causes errors when compiler optimization is turned off with
+`-O0` flag. A workaround is to use `Release` or `RelWithDebInfo`
+build types in CMake.
 
 Enzyme requires a matching LLVM installation (currently tested with LLVM 16)
 and must be compiled with the same Clang used to build GridKit.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,358 @@
+# GridKit Installation Guide
+
+GridKit has been built and tested on Linux and macOS. Windows users are
+encouraged to use [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install),
+which provides a Linux environment on Windows and is the easiest path to
+a working GridKit build on that platform. Native Windows builds (Cygwin or
+MSVC toolchain) are possible but not regularly tested.
+
+## Contents
+
+- [Requirements](#requirements)
+- [Getting the source](#getting-the-source)
+- [Building GridKit](#building-gridkit)
+- [CMake options](#cmake-options)
+- [Installing dependencies](#installing-dependencies)
+  - [SUNDIALS](#sundials)
+  - [Ipopt and HSL](#ipopt-and-hsl)
+  - [Enzyme](#enzyme)
+- [Using GridKit](#using-gridkit)
+- [Building with Spack](#building-with-spack)
+
+---
+
+## Requirements
+
+| Requirement | Version | Notes |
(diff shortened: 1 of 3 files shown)


lukelowry (approved):
This is very helpful, thank you. Perhaps out of scope, but please consider updating the dev container as well to make setting up the Enzyme dev environment on new machines easier. Otherwise the new instructions are clear from my perspective

INSTALL.md
```diff
@@ -0,0 +1,358 @@
...
+> `LINKER_LANGUAGE Fortran` for executables that use Ipopt+HSL. This
+> causes `gfortran` to drive the final link step, which automatically adds
+> the Fortran runtime libraries (`libgfortran`, `libquadmath`). No manual
+> `-lgfortran` flags should be needed.
+
+---
+
+### Enzyme
+
+Enzyme support is experimental. Only `Release` and `RelWithDebInfo` build
```
alexander-novo: Do they have to be specifically these build types, or is it a flag associated with these types by default? For instance can I set it to `Release` but then remove the optimization and `NDEBUG` flags?

alexander-novo (approved):

INSTALL.md
pelesh: I believe `-O 0` is what trips the Enzyme. I should clarify that in the instructions. Thanks!

CC @nkoukpaizan

INSTALL.md
```diff
@@ -0,0 +1,358 @@
...
+# macOS with Homebrew
+brew install suite-sparse
+```
+
+**Build SUNDIALS with KLU:**
+
+```sh
+cmake -B sundials-build -S sundials-src \
+  -DCMAKE_INSTALL_PREFIX=/path/to/sundials/install \
+  -DENABLE_KLU=ON \
```
Steven-Roberts, line 156: In version 7.7.0, this cmake option was deprecated and replaced with `SUNDIALS_ENABLE_KLU`. It's ok for now, but will need to be updated by SUNDIALS 8.0.0.

Steven-Roberts (approved):

INSTALL.md
pelesh: Clarified in [e761b96](https://github.com/ORNL/GridKit/pull/383/commits/e761b9675cba8d4d0274fc98899310ead82449fa).

ianmtaylor1:
Everything except the ipopt/HSL instructions looks good to me and mirrors what I did to install on Ubuntu 24.04. I just have a few suggestions for things that may be worth noting:

1. If suitesparse is installed in the default system location via the package manager, then its directories don't need to be specified when building SUNDIALS. In my experience they're found automatically.
2. The built-in Ubuntu package `coinor-libipopt-dev` will *not* work for ipopt. (I had installed this and thought it was working, but `GridKit_ENABLE_IPOPT` wasn't set. Setting it caused errors due to the Ubuntu package having an old version of an ipopt header: https://list.coin-or.org/pipermail/ipopt/2018-February/004545.html)

pelesh:
> Everything except the ipopt/HSL instructions looks good to me and mirrors what I did to install on Ubuntu 24.04. I just have a few suggestions for things that may be worth noting:
> 
> 1. If suitesparse is installed in the default system location via the package manager, then its directories don't need to be specified when building SUNDIALS. In my experience they're found automatically.
> 2. The built-in Ubuntu package `coinor-libipopt-dev` will _not_ work for ipopt. (I had installed this and thought it was working, but `GridKit_ENABLE_IPOPT` wasn't set. Setting it caused errors due to the Ubuntu package having an old version of an ipopt header: https://list.coin-or.org/pipermail/ipopt/2018-February/004545.html)

The instructions assume you are installing everything from scratch. When mixing pre-built binaries with custom built packages one has to ensure versions are compatible as you pointed out. I should add that clarification to the installation instructions.

PR #384, opened 2026-04-30
BusFault at each Bus as temporary request

lukelowry:
## Description
 
Temporary approach for the paper, so we can have a fault simulation suite, per @abirchfield
 

 ## Further comments
 
Recommend adding BusFault as a direct feature of `Bus`, similar to Branch opening and closing in #382


abirchfield (approved):

abirchfield:
This is just so we have the dataset to start developing a multi-scenario run system.

pelesh (approved):

PR #389, opened 2026-05-01
Event System Design Specification

lukelowry:
## Description
 
A draft of the event system draft with three supported events (fault on bus and branch, and branch opening).
 
 ## Proposed changes
 
- `EVENTS.md` detailing how to invoke `apply` and typed Actions
- `README.md` for `PDSim` detailing solver file format
- A new `Actions` section for `Bus` and `Branch`, which details the supported actions for each respective component.

@pelesh Thinking through this is helping inform some downstream concerns with other models and behavior. If you have a different design direction you want to consider, I am all ears; this is just what is intuitive to me personally.

## Further comments

This will allow us to remove `BusFault`, simplifying case creation and reducing bloat in case files. My external tool that translates cases from PowerWorld can then support what PW calls `contingencies`, which is on a par with the `schedule` I proposed here (allowing me to further streamline verification)

### DIFF
--- a/GridKit/Model/PhasorDynamics/Branch/README.md
+++ b/GridKit/Model/PhasorDynamics/Branch/README.md
@@ -147,6 +147,23 @@ The branch current relation is $0 = -\mathbf{I} + \mathbf{Y}\mathbf{V}$.
 These current contributions are added to the connected bus residuals with
 positive sign because branch current is oriented entering the bus.
 
+## Actions
+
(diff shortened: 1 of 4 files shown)


lukelowry:
I think this would be very valuable to have right now so we can express different kinds of disturbances, but I don't think there has been enough collective thought or discussion yet on how perturbations/events will work. Given the pace of GK, closing

PR #391, opened 2026-05-02
Branch Status for Out-Of-Service Cases

### LINKED ISSUE #370: Support Component Status + Enrich Branch Model
### Summary

Many power system cases come preconfigured with lines that are out of service. We need to support this and add a field to indicate whether, for example, a branch is open or closed, or whether a generator is in service.

Additionally, we might consider treating `Branch` as a special category, much like `Bus` is treated in the input format. 

### Rationale

This is a precursor to support a new event type `Open`/`Close` or something similarly named to indicate component status. This will also allow us to fully import case data without omitting inactive elements.

Regarding the second suggestion, there are quite a few branch models that are needed and should be completed before we push for more models, IMHO.

### Description

The system model should compose and support a status field that takes a component out of service, in addition to a field in the file format that is standard and well-named to communicate intent

The branch model class in the input format could be like:

```
   "buses": [
       { "class": "bus", ... },
       { "class": "infinite_bus",... }
   ],
   "branches": [
       { 
         "class": "xfmr", 
          "ports": [1, 2],   
         "name": "...", 
          "status": "active", 
          "init": {...},
          "params": {...}
       },
       { 
         "class": "xfmrtertiary", // three-winding transformer
          "ports": [4, 5, 6]   ,
         "name": "...", 
          "status:" "active", 
          "init": {...},
          "params": {...}
       },
       { 
         "class": "hvdc", 
          "ports": [4, 5]  ,
         "name": "...", 
          "status": "active", 
          "init": {...},
          "params": {...}
       },
       { 
         "class": "line", 
          "ports": [4, 5]   ,
         "name": "...", 
          "status": "active", 
          "init": {...},
          "params": {...}
       },
   ],
```

### Additional information

(regarding status as a field) If this requires changing the system model's dimension at runtime, it should be noted.

lukelowry:
## Description

Adds a `closed` parameter to the `Branch` model, allowing case files to faithfully import out-of-service transmission lines. Implemented as a `RealT` mask multiplier on residual contributions, following the `d65bd4f` IEEEST mask precedent (branch-free residuals for AD).

Closes #370.

## Proposed changes

- `BranchParameters::closed` (bool, default `true`) new optional param in the `Branch` device's `params` block.
- `RealT closed_{1.0}` member on `Branch` multiplied into all four `evaluateBusResidualNN` outputs. When `closed = false`, residual contributions and Jacobian entries are zero, with no change to AD behavior. This approach, therefore, includes out-of-service branches in the sparsity structure. Long term, this should be optimized, but is out of scope here.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
A few of the larger cases have branches that I must omit when translating to `GridKit` because we don't have a way to declare branch status. This is a part of a greater effort of mine to have as little friction in case translation as possible, so we can have full parity with PowerWorld cases.

### DIFF
--- a/tests/UnitTests/PhasorDynamics/BranchTests.hpp
+++ b/tests/UnitTests/PhasorDynamics/BranchTests.hpp
@@ -3,6 +3,7 @@
 #include <iomanip>
 #include <iostream>
 #include <limits>
+#include <variant>
 
 #include <GridKit/AutomaticDifferentiation/DependencyTracking/Variable.hpp>
 #include <GridKit/Model/PhasorDynamics/Branch/Branch.hpp>
@@ -127,10 +128,11 @@ namespace GridKit
         branch.allocate();
         branch.evaluateResidual();
 
-        success *= isEqual(bus1.Ir(), Ir1);
-        success *= isEqual(bus1.Ii(), Ii1);
-        success *= isEqual(bus2.Ir(), Ir2);
-        success *= isEqual(bus2.Ii(), Ii2);
+        const RealT tol{1.0e-12};
+        success *= isEqual(bus1.Ir(), Ir1, tol);
+        success *= isEqual(bus1.Ii(), Ii1, tol);
+        success *= isEqual(bus2.Ir(), Ir2, tol);
+        success *= isEqual(bus2.Ii(), Ii2, tol);
 
         return success.report(__func__);
       }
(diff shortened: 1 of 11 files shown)


tests/UnitTests/PhasorDynamics/BranchTests.hpp
```diff
@@ -120,6 +120,56 @@ namespace GridKit
...
+        data.parameters[PhasorDynamics::BranchParameters::B]      = B;
+        data.parameters[PhasorDynamics::BranchParameters::closed] = false;
+
+        PhasorDynamics::Branch<DependencyTracking::Variable, IdxT> branch(&bus1, &bus2, data);
+        branch.allocate();
+        branch.evaluateResidual();
+
+        /// Open branch contributes nothing: residual values AND every tracked
+        /// derivative coefficient must be zero. Verifies the `closed` parse
+        /// path, the residual mask, and mask propagation through AD in one shot.
```
nkoukpaizan: I suggest adding a test for the Enzyme Jacobian as well, since that is the one that will be used in simulations.

pelesh (changes requested):
There are a few things that need to be addressed before this PR could be merged:
- Documentation: Mathematical model of the governing equations and derivatives need to be defined for the case when the line is out of service. Must ensure that the system is not singular when the line is out of service.
- Complete set of unit tests needs to be implemented including Jacobian tests per @nkoukpaizan comment.
- Parameter `closed` should be boolean. There should be "derived" parameter private to the branch model to set the factor multiplying injection currents. This should prevent bugs when the input is incorrect (e.g. 1.1 instead of 1.0 as the current solution requires).

lukelowry:
moving to draft and will bring back once #417 is merged to prevent confusion

lukelowry:
I have thought about this PR for a while now.  I think there are too many details left up in the air about how GridKit should respond when part of the system is islanded. We will need to address this soon once we do more advanced things other than `BusFault`. For the time being I will change this so it is an initialization/enabled parameter only so that we can stored disabled devices in the JSON. Will open for review when done

lukelowry:
I highly highly prefer we take apporach #514 . closing in favor of that design.

PR #394, opened 2026-05-06
Initial implementation of multi-contingency app

### LINKED ISSUE #392: Add contingency analysis app for phasor domain dynamics
### Summary

Create an $N-1$ contingency analysis app that goes through all faults in the system and runs a simulation for each of them. 

### Rationale

$N-1$ contingency analysis is a standard procedure for stability studies and GridKit should provide an app that does that rather than expecting user to write scripts to run a single fault app with different contingency in succession.

### Description

The app should load the system and then loop through the list of all fault components, run a stability analysis simulation for each fault, and store complete simulation results (each in separate csv file).

Ideally, the loop will be parallelized. For now, shared memory parallelism (e.g. OpenMP ) should be okay. In that case, we probably need a separate instance of a model and a solver on each thread, together with a list of contingencies assigned to that thread.

For the first pull request, a partially hard-wired solution with limited configuration options should suffice.

### Additional information

_No response_

PhilipFackler:
## Description
 
Refactored `pdsim` code as `ContingencyStudy` class and reused in initial implementation of concurrent contingency app.
 
 _Closes #392_
 
 ## Proposed changes
 
For the work-in-progress I implemented three different versions of the simulation for easy timing comparison.
- serial
- async (using std::async)
- openmp

The easiest interesting case to try locally is "new england" (I have 22 cores in my laptop).
 
 ## Checklist
 
- [x] All tests pass.


lukelowry:
'ContingencyStudy' -> 'CTGStudy'

CTG is common short hand and way easier to type! One option.

PhilipFackler:
yes, I need feedback on naming things

pelesh:
A few comments:
- Great to have multiple execution options to compare performance!
- I suggest restoring PDSim as it was. It is a different app and should not be modified in this PR.
- I would contain implementation of `ContingencyStudy` to a single file or a source and a header file.
- We have development of application utilities in our roadmap, but this is beyond the scope of this PR.

application/PhasorDynamics/CMakeLists.txt
```diff
@@ -9,4 +9,18 @@ target_include_directories(PDSim PRIVATE
     ${GRIDKIT_THIRD_PARTY_DIR}/nlohmann-json/include
     ${GRIDKIT_THIRD_PARTY_DIR}/magic-enum/include)
 
+find_package(OpenMP REQUIRED)
```
nkoukpaizan: At first glance, I would expect OpenMP to be an optional dependency. We might want a CMake option to enable/disable this.

application/PhasorDynamics/CMakeLists.txt
```diff
@@ -9,4 +9,18 @@ target_include_directories(PDSim PRIVATE
...
+add_executable(ConCon ConCon.cpp CTGStudy.cpp)
```
nkoukpaizan: `ConCon` is not very descriptive to me. Some documentation in the current `ConCon.cpp` would also help. It's currently hard for me to follow, so I don't know what other name to suggest.

nkoukpaizan (commented):
A couple of comments on this below. We'll also need to update the applications `README` when this is ready, to minimize users' confusion.

lukelowry:
Very excited for this! I think once we have this merged we should address the events-as-components modeling so we don't need to store a BusFault for each bus in the case. 

Also, if possible, I'd like the *.study.json to have a "ida_stats" field for a file output string that would if passed calls the IDA api and writes out the IDA statistics (verbose extended version). I would use that ASAP with this + PDSIM

application/PhasorDynamics/CMakeLists.txt
PhilipFackler: It definitely needs a new name. `ConCon` is short for "concurrent contingencies". I made the class `ContingencyStudy` to represent one "study". `PDSim` performs one study where as `ConCon` performs a lot of studies concurrently. And I'm happy to rename both of these apps to whatever would make more sense to users.

application/PhasorDynamics/CMakeLists.txt
PhilipFackler: Yes, this makes sense. My intention here was to provide two different parallel implementations (omp and stdc++) and discuss which route we want to take.

application/PhasorDynamics/CMakeLists.txt
pelesh: I would go with:
- `ConCon` --> `ContingencyAnalysis`, and
- `CTGStudy` --> `ContingencyUtilities`

I would also replace `PDSim` --> `DynamicSimulation`. The "PD" stands for phasor dynamics but they reside in `PhasorDynamics` subdirectory, so "PD" is redundant imho.

PhilipFackler: @pelesh I'm confused by the suggestion to name a class with "Utilities". At the moment the `CTGStudy` class represents one study (as specified by the input file). If we write other utilities for these apps, wouldn't they go in separate classes?

PhilipFackler: I'll make the suggested change for `PDSim` for #406

application/PhasorDynamics/CMakeLists.txt
PhilipFackler: I made openmp default but optional

pelesh (changes requested):
The new code needs to be better commented. The purpose of classes, structures and standalone functions implemented here needs to be better explained.

It is unclear what is the purpose of `CTGStudy::run()` function. This could be implemented within standalone function `runStudy`. I would suggest removing `CTGStudy` class altogether and moving utility functions for I/O processing to `ContingencyUtilities.*pp` file(s).

application/PhasorDynamics/CMakeLists.txt
```diff
@@ -1,12 +1,30 @@
...
+  PUBLIC
+  GridKit::phasor_dynamics_components
+  GridKit::solvers_dyn
+  GridKit::Utilities
+  GridKit::testing)
+target_include_directories(ContingencyAnalysis PRIVATE
+  ${GRIDKIT_THIRD_PARTY_DIR}/nlohmann-json/include
+  ${GRIDKIT_THIRD_PARTY_DIR}/magic-enum/include)
+
+option(GridKit_ENABLE_OPENMP "Use OpenMP in contingency analysis application" ON)
```
nkoukpaizan: I would put all options in the top level CMakeLists.txt, so we can easily find them later. I understand this is for testing the contingency analysis, but we are likely to keep the OpenMP option anyway.

Also, I'm not opposed to `GRIDKIT_ENABLE_` --> `GridKit_ENABLE_`; I would rather change all of them at the same time rather than having both co-exist.

nkoukpaizan (commented):
Almost there I well. I'm able to follow what's in `ContigencyAnalysis.cpp`, though more documentation (`runStudy`, `runStudyAsync`, etc.) is always welcome.

I would add an example that uses the `ContingencyAnalysis` executable that shows how this works and runs along with the other tests/examples.

application/PhasorDynamics/CMakeLists.txt
PhilipFackler: I moved it and followed the pattern with the the other "ENABLE"s.

application/PhasorDynamics/CMakeLists.txt
```diff
@@ -1,12 +1,29 @@
...
+  GridKit::phasor_dynamics_components
+  GridKit::solvers_dyn
+  GridKit::Utilities
+  GridKit::testing)
+target_include_directories(ContingencyAnalysis PRIVATE
+  ${GRIDKIT_THIRD_PARTY_DIR}/nlohmann-json/include
+  ${GRIDKIT_THIRD_PARTY_DIR}/magic-enum/include)
+
+if(GRIDKIT_ENABLE_OPENMP)
+  find_package(OpenMP REQUIRED)
```
nkoukpaizan: I would move the `find_package` up as well.

application/PhasorDynamics/CMakeLists.txt
```diff
@@ -1,12 +1,29 @@
...
+  target_link_libraries(ContingencyAnalysis PUBLIC OpenMP::OpenMP_CXX)
+endif()
+
+install(TARGETS DynamicSimulation EXPORT gridkit-targets RUNTIME)
+install(TARGETS ContingencyAnalysis EXPORT gridkit-targets RUNTIME)
```
nkoukpaizan, line 28: Please add an automated example of how to use this executable.

CMakeLists.txt
```diff
@@ -38,8 +38,11 @@ option(GridKit_ENABLE_ASAN "Enable the address sanitizer" OFF)
 # Enable the undefined behavior sanitizer
 option(GridKit_ENABLE_UBSAN "Enable the undefined behavior sanitizer" OFF)
 
+# Enable OpenMP
+option(GridKit_ENABLE_OPENMP "Use OpenMP in contingency analysis application" ON)
```
nkoukpaizan:
```suggestion
option(GridKit_ENABLE_OPENMP "Enable OpenMP" ON)
```

application/PhasorDynamics/CMakeLists.txt
PhilipFackler: What do you mean by "automated" in this context?

nkoukpaizan: An example with `add_test` that illustrates the expected inputs, etc. "Automated" in the sense that I want to be able to test this with the usual `make test`/`ctest`.

PhilipFackler: Right. Sorry, that should have been obvious. :facepalm: I'll do that now.

PhilipFackler: OK, I added the newengland case. I had to set the `WILL_FAIL` property, though, since sundials aborts on three of the faults.

CMakeLists.txt
```diff
@@ -38,8 +38,11 @@ option(GridKit_ENABLE_ASAN "Enable the address sanitizer" OFF)
 # Enable the undefined behavior sanitizer
 option(GridKit_ENABLE_UBSAN "Enable the undefined behavior sanitizer" OFF)
 
+# Enable OpenMP
+option(GridKit_ENABLE_OPENMP "Enable OpenMP" ON)
```
nkoukpaizan:
```suggestion
option(GridKit_ENABLE_OPENMP "Enable OpenMP" OFF)
```

nkoukpaizan (approved):
Let's keep the options off by default. Otherwise, the changes make sense to me. I'm seeing `std::async` outperforming `OpenMP`. 

Will wait for @pelesh's review.

CMakeLists.txt
PhilipFackler: Sounds good to me

pelesh (approved):

pelesh:
There is quite bit of redundant output coming from the app but it looks as something inherited from the `Ida` class. I suggest we merge this as is and have another PR to make this more user friendly. The PR says this is "Initial implementation" :).

pelesh:
Rebase to fix changelog conflict and address comment by @nkoukpaizan.

PhilipFackler:
@pelesh @nkoukpaizan should be good to go

PR #395, opened 2026-05-06
Improved Schema Input Format for PD

lukelowry:
## Description
 
More consistent and navigable schema, which will make case translation much easier for both ingress and egress
 
 
 ## Proposed changes
 
- ~~`branches` are promoted to the same level as `buses` to support many branch types~~
- `buses` are the only exception to the component system. While `branches` are also treated differently by the system model, adding a third type offers few benefits beyond manual readability.
 
 ## Further comments
 
I strongly advocate for a shared schema between EMT and PhasorDynamics. This helps practically for case translation, creating synthetic EMT cases, and conceptually for co- and hybrid-simulation.

### DIFF
--- a/GridKit/Model/PhasorDynamics/INPUT.md
+++ b/GridKit/Model/PhasorDynamics/INPUT.md
@@ -1,0 +1,193 @@
+# PhasorDynamics Input Format
+
+A PhasorDynamics input file is a JSON object that describes one dynamic phasor
+case: metadata, buses, network branches, devices, non-network signals, and
+monitor output sinks.
+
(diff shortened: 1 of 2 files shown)


PhilipFackler (commented):
I don't have a judgement about this, so I'll let someone else approve. It would be nice to see an example json file to get a better idea of what this changes and what will need to change in the parser functions.

...I guess this PR should ideally end up including changes to the parser functions and the all the existing input files.

lukelowry:
Close for now Id rather focus on just doing EMT correctly and using what we learned in PhasorDynamics to inform decisions there

PR #397, opened 2026-05-13
EMT Lumped Model Documentation

lukelowry:
## Description
 
Initial documentation for the lumped EMT bus and branch model. This is one possible variant of EMT implementation.

Additional design for the initial basic model for the voltage source and RL load.
 
Per offline meeting with @abirchfield
 
 ## Proposed changes
 
Documentation for the following

- `Bus` folder
  - `Bus` component, which owns the three-phase vector of bus voltages
- `Component` folder, which contains everything else
  - `BranchLumpedConstant` as initial branch model for EMT without time delays
  - `LoadRL` with RL series components.
  - `Source` with resistive terminal and temporally defined source (would need to pass `t` directly to this component and manually construct the Jacobean for this as a temporary solution - but since this is only a surrogate for a more advanced model, this is sufficient for now)


pelesh (changes requested):
There should be separate directory `GridKit/Model/Emt` where all EMT models and documentation are placed. The `PhasorDomain` models should remain separate.

lukelowry:
Moved and organized into preliminary sub-folders. Added Circuit diagram for `BranchLumpedConstant` aswell

pelesh (approved):
Looks like a good start. Will merge as soon as @abirchfield approves.

abirchfield (approved):

PR #401, opened 2026-05-15
Introduce cosimulation interface component

### LINKED ISSUE #393: Create a cosimulation interface component in phasor dynamics library
### Summary

As an alternative to the solution proposed in #354, create a separate component that would connect to an electrical bus on one side and and have voltage signal outputs and injection current signal inputs on the other.

CC @alexander-novo @lukelowry @abirchfield @nkoukpaizan 

### Rationale

Such component would abstract away cosimulation environment from the rest of the grid model. It would also allow for storing prior history locally for implementing advanced cosimulation methods.

This approach is less invasive than the one proposed in #354.

### Description

The component would have following connection ports:
- electrical bus
- signal output port transmitting bus $V_r$
- signal output port transmitting bus $V_i$
- signal input port receiving injection current $I_r$
- signal input port receiving injection current $I_i$

The initial implementation should simply make voltages from the electrical bus accessible through output signals. Residual evaluation from this component should add the injection current to the residual evaluated at the electrical bus.

The implementation should include an integration test (e.g. a two-bus system connected through cosimulation interface components.

### Additional information

We need to decide whether to use this approach or #354. One excludes the other.

PhilipFackler:
## Description
 
Basic implementation of cosimulation component
 
 _Closes #393_
 
 ## Proposed changes
 
Uses two output signals to transmit $V_r$ and $V_i$ from connected bus and two input signals to transfer $I_r$ and $I_i$ to connected bus.
 
 ## Checklist
 
- [X] All tests pass.


GridKit/Model/PhasorDynamics/Connector/CoSimImpl.hpp
```diff
@@ -0,0 +1,172 @@
...
+       */
+      template <class ScalarT, typename IdxT>
+      int CoSim<ScalarT, IdxT>::evaluateResidual()
+      {
+        static constexpr auto IREAL = CoSimExternalVariables::IREAL;
+        static constexpr auto IIMAG = CoSimExternalVariables::IIMAG;
+
+        if (signals_.template isAttached<IREAL>())
+        {
+          bus_->Ir() = signals_.template readExternalVariable<IREAL>();
```
lukelowry: Should this be += instead of =? Unless we call the evalResidual before all other components, but check my logic here

lukelowry (changes requested):

GridKit/Model/PhasorDynamics/Connector/CoSimImpl.hpp
PhilipFackler: That wouldn't fit MY intuition, but I don't know. I'll defer. @pelesh can you answer this?

lukelowry:
Only had one implementation concern regarding correctness. 

The facade of the Vr/Vi being treated as an internal variable is confusing to me. What are your thoughts on an auxiliary internal state? That would be more consistent, and since we would only have a small number of CoSim components for any given system model, the state duplication isn't too costly.

PhilipFackler:
> Only had one implementation concern regarding correctness.
> 
> The facade of the Vr/Vi being treated as an internal variable is confusing to me. What are your thoughts on an auxiliary internal state? That would be more consistent, and since we would only have a small number of CoSim components for any given system model, the state duplication isn't too costly.

My opinion is that the arrays in `ComponentSignals` should be based on input/output ports rather than internal/external variables. I would split the `Ports` enum for a `ComponentData` implementation into `InputPorts` and `OutputPorts`. If we made that change, things would make more sense, because you would only allow signal nodes at actual ports. Internal and external variables seem orthogonal (or at least linearly independent :smile_cat: ) to signal nodes and should not be used here.

All that to say, given the current implementation of `ComponentSignals`, the `CoSim` component has to pretend.

lukelowry:
> My opinion is that the arrays in `ComponentSignals` should be based on input/output ports rather than internal/external variables. I would split the `Ports` enum for a `ComponentData` implementation into `InputPorts` and `OutputPorts`. 

I am a big fan of this solution, this would resolve some modeling vocab concerns I have in our exciter and stabilizers.

GridKit/Model/PhasorDynamics/Connector/CoSim.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "CoSimImpl.hpp"
+
+namespace GridKit
+{
+  namespace PhasorDynamics
+  {
+    namespace Connector
+    {
+      // Available template instantiations
+      template class CoSim<double, long int>;
```
nkoukpaizan: `Connector::CoSim` as a component is confusing to me. Domain experts can chime in on the naming, but my understanding is that `Co-Simulation` is the approach. Perhaps we can call the class `CoSimConnector` (unless we plan to add other things to the namespace)? I would understand `CoSim::Connector` as well, though I expect the co-simulation environment to live outside of `PhasorDynamics`.

GridKit/Model/PhasorDynamics/Connector/CoSim.hpp
```diff
@@ -0,0 +1,109 @@
...
+      /// External variables of a `CoSim`
+      enum class CoSimExternalVariables : size_t
+      {
+        IREAL,
+        IIMAG,
+        MAXIMUM,
+      };
+
+      template <class ScalarT, typename IdxT>
+      class CoSim : public Component<ScalarT, IdxT>
```
nkoukpaizan: I'm not sure this should inherit from `Component`. If it does, we will need a README in the `Connector` folder describing the model. But to me, this should behave closer to the signal nodes or buses than other components.

nkoukpaizan (commented):
Setting to draft for now. I have some thoughts, but I'll let @pelesh discuss this offline.

GridKit/Model/PhasorDynamics/Connector/CoSim.cpp
lukelowry: CoSim::Connector is okay. Most other power system frameworks actually model the interfacing component as a circuit element, so either a lossless branch with a time delay, or a Bus.

From an electrical perspective a Bus or Branch implementation (BusCoSim, or BranchCoSim) would be correct and has an equivalent circuit model which is nice.

I'm okay with either

PhilipFackler: Good points. I was not at all sure about the naming/organizing. Happy to change as you all see fit.

GridKit/Model/PhasorDynamics/Connector/CoSimImpl.hpp
pelesh: Yes, your intuition is correct, injection currents add to the sum of currents on the bus.

pelesh:
> > My opinion is that the arrays in `ComponentSignals` should be based on input/output ports rather than internal/external variables. I would split the `Ports` enum for a `ComponentData` implementation into `InputPorts` and `OutputPorts`.
> 
> I am a big fan of this solution, this would resolve some modeling vocab concerns I have in our exciter and stabilizers.

This is correct understanding. Signals represent directed communication and are defined by inputs/outputs. Physics-based couplings are based on conservation laws and have associated variables that are external to coupled components.

GridKit/Model/PhasorDynamics/Connector/CoSim.cpp
pelesh: I would call it `BusToSignalAdapter` or something along those lines. While we are developing it for use in co-simulations, it is not a co-simulation object itself.  I also think "connector" is a misleading name because it is not a model of a physical connector. It is just a modeling "adapter" without physical meaning, used to convert data for use by models of different type.

pelesh: I suggest for the purpose of this PR to use temporary names (something that we can search/replace easily).

GridKit/Model/PhasorDynamics/Connector/CoSim.hpp
pelesh: I would let it be a component for now (and, yes, we need a README). It connects to a bus and has signal ports, so it can use the infrastructure of a component.

The other end of the connection (outside the scope of this PR) should probably inherit from the bus and allow components to connect to it.

GridKit/Model/PhasorDynamics/Connector/CoSimImpl.hpp
PhilipFackler: Sounds good. I switched to `+=` in both `initialize` and `evaluateResidual`

GridKit/Model/PhasorDynamics/Connector/CoSim.cpp
PhilipFackler: I have renamed to `BusToSignalAdapter` and removed the `Connector` namespace.

PhilipFackler:
@nkoukpaizan after name changes, I believe this PR should be ready for review again. Let me know what you think.

GridKit/Model/PhasorDynamics/INPUT_FORMAT.md
```diff
@@ -141,14 +143,14 @@ are specified:
...
   `Tgov1 `      | the TGOV1 governor model                             | `pmech`, `speed`                 | `R`, `T1`, `T2`, `T3`, `Pvmax`, `Pvmin`, `Dt` | `none`
   `Ieeet1`      | the IEEET1 exciter model                             | `bus`, `speed`, `efd`, `vs`\*    | `Tr`, `Ka`, `Ta`, `Ke`, `Te`, `Kf`, `Tf`, `Vrmin`, `Vrmax`, `E1`, `E2`, `Se1`, `Se2`, `Ispdlim` | `efd`, `ksat`
   `SexsPti`     | the SEXS-PTI simplified exciter model                | `bus`, `efd`, `vs`\*             | `Ta`, `Tb`, `Te`, `K`, `Efdmax`, `Efdmin` | `efd`
   `Ieeest`      | the IEEEST stabilizer model                          | `input`, `output`, `cutout`\*    | `A1`, `A2`, `A3`, `A4`, `A5`, `A6`, `T1`, `T2`, `T3`, `T4`, `T5`, `T6`, `Ks`, `Lsmin`, `Lsmax`, `Vcl`, `Vcu`, `Tdelay` | `vs`
   `BusFault`    | simple impedance-based fault at a bus                | `bus`, `status`\*                | `state0`, `R`, `X` | `state`, `ir`, `ii`
+  `CoSim`       | co-simulation interface component                    | `bus`, `vr`, `vi`, `ir`, `ii`    |                             |
```
nkoukpaizan: Please update to `BusToSignalAdapter`.

CHANGELOG.md
```diff
@@ -47,15 +47,15 @@
...
-- Added `IEEEST` Stabilizer Model
-- Added `SEXS-PTI` Exciter Model
-- Added `GENSAL` Machine Model
+- Added IEEEST Stabilizer Model
+- Added SEXS-PTI Exciter Model
 - Added 200 Bus Synthetic Illinois Case
 - Added node objects to `PowerElectronics` module & updated all examples to make use of them.
 - Separated internal and external residuals of `PowerElectronics` models.
 - Added `CliArgs` class for better management of command-line options.
+- Added `CoSim` co-simulation interface component.
```
nkoukpaizan: Update to `BusToSignalAdapter`

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
```diff
@@ -0,0 +1,169 @@
...
+      if (signals_.template isAttached<IIMAG>())
+      {
+        bus_->Ii() += signals_.template readExternalVariable<IIMAG>();
+      }
+
+      return 0;
+    }
+
+    /**
+     * @brief Residual evaluation
```
nkoukpaizan:
```suggestion
     * @brief Jacobian evaluation
```

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
```diff
@@ -0,0 +1,169 @@
...
+    {
+      static constexpr auto VREAL = BusToSignalAdapterInternalVariables::VREAL;
+      static constexpr auto VIMAG = BusToSignalAdapterInternalVariables::VIMAG;
+      // FIXME: using a placeholder for variable index here is wrong. But it
+      //        also seems like nothing ever uses the index from the signal.
+      //        So we should either...
+      //        - add methods to BusBase to get variable indices (satisfy the
+      //          interface correctly)
+      //        - remove variable index from SignalNode (because maybe it is
+      //          not actually needed)
```
nkoukpaizan: The index is used in the Jacobian computations.

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
```diff
@@ -0,0 +1,169 @@
...
+      return 0;
+    }
+
+    /**
+     * @brief Residual evaluation
+     */
+    template <class ScalarT, typename IdxT>
+    int BusToSignalAdapter<ScalarT, IdxT>::evaluateJacobian()
+    {
+      return 0;
```
nkoukpaizan, line 150: Would be good to compute the Jacobian and test it. If I understand correctly, this would have a $\partial h/\partial w_s$ term, which is new. I can help with the implementation if needed.

nkoukpaizan (commented):
A couple of minor comments. 

The Jacobian may be out of scope for this PR. Just let me know if you want me to take a crack at the implemetation.

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
pelesh: I think Jacobian should be out of scope for this PR. For now we are trying to set up just a simple co-simulation where we exchange variables but not derivatives between cosimulated parts.

pelesh:
> > My opinion is that the arrays in `ComponentSignals` should be based on input/output ports rather than internal/external variables. I would split the `Ports` enum for a `ComponentData` implementation into `InputPorts` and `OutputPorts`.
> 
> I am a big fan of this solution, this would resolve some modeling vocab concerns I have in our exciter and stabilizers.

I believe that was the original intent: You would have signal-in, signal-out and bus ports. Note that electrical buses cannot be considered in or out because they enforce algebraic constraints, which are not directional.

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
PhilipFackler: I'll create an issue for this.

PhilipFackler: Technically this functions as-is because you don't need the index for this component. Am I right in assuming we don't want to fix the BusBase interface for this PR?

nkoukpaizan: It is correct that it is not needed here. `BusBase` does not set an index, but `Bus` does. If a component does not own any variable (as is the case here), there is no index to set.

nkoukpaizan: To satisfy the signal interface, we can set `vr_index_` and `vi_index_` to `INVALID_INDEX` here.

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapterImpl.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+      if (signals_.template isAttached<IREAL>())
+      {
+        bus_->Ir() += signals_.template readExternalVariable<IREAL>();
+      }
+      if (signals_.template isAttached<IIMAG>())
+      {
+        bus_->Ii() += signals_.template readExternalVariable<IIMAG>();
+      }
+
```
nkoukpaizan:
```suggestion
```

I don't think this is needed in `initialize`. `evaluateResidual` is sufficient.

GridKit/Model/PhasorDynamics/BusToSignalAdapter/BusToSignalAdapter.hpp
```diff
@@ -0,0 +1,104 @@
...
+          -> ComponentSignals<ScalarT,
+                              IdxT,
+                              BusToSignalAdapterInternalVariables,
+                              BusToSignalAdapterExternalVariables>&
+      {
+        return signals_;
+      }
+
+    private:
+      // Placeholders for variable indices (see note in allocate() method)
```
nkoukpaizan, line 90: Please update the comments. I would include somewhere in those comments that this component only passes bus voltages and updates residual information, which justifies ignoring the indices.

nkoukpaizan (commented):
This is almost there in my opinion. I would just add a REAME file to the folder describing the purpose of the component and the ports (reads voltages from the bus to make them available via signals and adds external current injections).

tests/UnitTests/PhasorDynamics/BusToSignalAdapterTests.hpp
```diff
@@ -0,0 +1,98 @@
...
+        auto bus = BusT(Vr, Vi);
+        bus.allocate();
+        bus.initialize();
+
+        auto vr_sig = SignalT({.name = "vr", .signal_id = 0});
+        auto vi_sig = SignalT({.name = "vi", .signal_id = 1});
+        auto ir_sig = SignalT({.name = "ir", .signal_id = 2});
+        auto ii_sig = SignalT({.name = "ii", .signal_id = 3});
+
+        IdxT ir_index;
```
nkoukpaizan: I would set to `INVALID_INDEX` here as well.

nkoukpaizan (approved):
Thanks for cleaning this up. One minor comment, and I will wait for @pelesh to approve before merging.

lukelowry (approved):
Good for now, great job!

PR #403, opened 2026-05-15
Update Phasor Dynamics Load model to match documentation.

nkoukpaizan:
## Description
 
This refactors the Phasor Dynamics Load model to store the current injection as internal variables
 
 ## Proposed changes
 
- Added current injection as internal variables for the `Load` model. 
- Updated Jacobians, tests and examples accordingly
 
 ## Checklist

- [x] All tests pass.
- [N/A This is not a new feature. It is making the implementation match the documentation] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so. 
 
 ## Further comments


lukelowry (approved):
Looks good in terms of correctness! 

BusFault is the only other model that doesn't have internal current state but it's a bit of an exception and I don't think it should be a component personally.

Issue #145, opened 2025-06-18
Add method(s) to update variables of a specific component


pelesh:
### Summary

Based on issues #142 and #143 consider adding following method to the `GridKit::PhasorDynamics::SystemModel` class:
```c++
int updateComponentVariables(index_type component_id);
``` 
The method would prompt component (or subsystem) `component_id` to copy variable values from the system solution vector to component member variables. These are denoted as $y^r$ in #143.

Similarly, method
```c++
int updateBusVariables(index_type bus_id);
```
would prompt bus `bus_id` to update its own variables from the system solution vector. These variables are (should have been) denoted as $w^b$ in #143.

CC @Steven-Roberts @reid-g 

### Rationale

Multimethod numerical schemes may need to update only a subset of system variables, so there need to be methods that allow for that.

### Description

In current GridKit implementation, the proposed function would look like this:
```c++
template<class scalar_type, class index_type>
int SystemSolver<scalar_type, index_type>::updateComponentVariables(index_type component_id)
{
  for (IdxT j = 0; j < component->size(); ++j)
  {
    component->y()[j]  = y_[varOffset[component_id] + j];
    component->yp()[j] = yp_[varOffset[component_id] + j];
  }
  return 0;
}
```

### Additional information

Eventually, with tighter integration with numerical integrator, these methods may become obsolete. Component internal variables could be simply references to the elements of system variables vector. In that case, the numerical integrator would write solution to the appropriate elements in the system solution vector and each component would have the up-to-date variables.

PR #411, opened 2026-05-22
Global state views in PhasorDynamics SystemModel 

lukelowry:
## Description

This PR updates PhasorDynamics so buses and components use their slices of the `SystemModel` state vectors directly. This removes the local state copy path and keeps component accessors consistent with the integrator-updated system state.

Partially addresses #96. Related to #145.

## Proposed changes

- Added `StateVector` support so `GridElement` can own standalone state or alias `SystemModel` state.
- Updated `SystemModel` allocation to bind each bus/component to global state and verify element sizes remain fixed.
- Updated Enzyme call sites to use the explicit `yData()` accessor for live state pointers.
- Added a PhasorDynamics system test for element/system state aliasing.

| Case | `develop` (s) | This branch (s) | Improvement |
| --- | ---: | ---: | ---: |
| Hawaii | 0.201302 | 0.165187 | 17.9% |
| NewEngland | 0.272596 | 0.220406 | 19.1% |
| Illinois | 0.664262 | 0.592922 | 10.7% |
| WECC | 1.203380 | 1.045540 | 13.1% |
| Texas | 9.063860 | 7.515690 | 17.1% |

## Checklist

- [x] All tests pass. (except the known PE failure)
- [x] N/A: no [CHANGELOG.md](/CHANGELOG.md) update is needed for this internal PhasorDynamics state-storage change.

## Further comments

This work was motivated by the addition of IEEEST stabilizers: adding 333 stabilizers to the Texas case dramatically reduced performance, and this change removes avoidable state-copy overhead in PhasorDynamics system evaluation.

This is not intended to be the final state-vector API design. The cleaner long-term solution would be to update `Evaluator` so state and residual accessors expose non-owning vector views instead of requiring owned `std::vector` storage, e.g.

```cpp
virtual VectorView<ScalarT> y() = 0;
virtual VectorView<ScalarT> yp() = 0;
virtual VectorView<ScalarT> residual() = 0;
```

This PR keeps the change local to PhasorDynamics while removing the current copy path.


pelesh (changes requested):
Blocking this PR until Jacobian implementation in `PhasorDynamics` is refactored to eliminate deprecated `Matrix_COO` object.

pelesh:
@lukelowry, please take a look at [vector class](https://github.com/ORNL/ReSolve/tree/develop/resolve/vector) in Re::Solve. This is something we need to add to the `GridKit/LinearAlgebra` directory and use it in place of  `StateVector` class you added to `GridElement.hpp`. I'm afraid `StateVector` class may not be portable as implemented in this branch.

lukelowry:
@pelesh Are you suggesting `Re::Solve` become a dependency? or use as inspiration for our own `Vector` class? I moved and refined the implementation as you suggested, but a full implementation would be a larger change. I added a comment in the code for the future work we would need to do.

pelesh:
> @pelesh Are you suggesting `Re::Solve` become a dependency? or use as inspiration for our own `Vector` class? I moved and refined the implementation as you suggested, but a full implementation would be a larger change. I added a comment in the code for the future work we would need to do.

See #418

PhilipFackler (approved):
I think this is great as-is. Replacing this `Vector` class with the Re::Solve version in a later PR would be pretty straightforward.

lukelowry:
@pelesh I rebased this and semi-implemented the same zero-copy behavior to speed up. I am struggling to find an easier way to achieve what I am doing here without making impactful API changes to `Evaluator`.

lukelowry:
Close

PR #414, opened 2026-05-25
Transformer Documentation

lukelowry:
## Description

Transformer README for the component that supports off-nominal tap ratios and phase shifting.
 
 ## Proposed changes
 
- Transformer Component Documentation
 
 ## Further comments
 
Last item for the large cases to have the same steady state solutions as the non-simplified version of the case.


lukelowry:
After careful thought, I think it's best to simply add `tap` and `phase` as parameters to `Branch` and never introduce `Transformer` unless there is a specific need in the future. Will open a new PR soon for that instead.

PR #418, opened 2026-05-29
Add a vector class to GridKit

pelesh:
## Description
 
 Add vector class to GridKit to replace `std::vector` and provide portable data management. 

 
## Proposed changes
 
The `LinearAlgebra::Vector` class was ported from Re::Solve. It has basic initialization and data management functionality. I can be use as a vector, multivector or vector view class. 
 
 ## Checklist
 
 
- [x] All tests pass.
 
 ## Further comments
 
The multivector capability in this class should replace `DenseMatrix` object in GridKit.

### DIFF
--- a/GridKit/LinearAlgebra/Vector/Vector.cpp
+++ b/GridKit/LinearAlgebra/Vector/Vector.cpp
@@ -122,8 +122,7 @@ namespace GridKit
       using namespace memory;
       if (d_data_ || h_data_)
       {
-        out::error() << "Trying to set vector data, but the data already exists!\n";
-        out::error() << "Ignoring setData function call ...\n";
+        out::error() << "Vector::setData - data already exists, ignoring call\n";
         return 1;
       }
 
@@ -246,16 +245,14 @@ namespace GridKit
       case memory::HOST:
         if (h_data_ == nullptr)
         {
-          out::error() << "Vector::copyFromExternal - destination memory space"
-                       << " on host not allocated\n";
+          out::error() << "Vector::copyFromExternal - host destination not allocated\n";
           return 1;
         }
         break;
       case memory::DEVICE:
         if (d_data_ == nullptr)
         {
-          out::error() << "Vector::copyFromExternal - destination memory space"
-                       << " on device not allocated\n";
+          out::error() << "Vector::copyFromExternal - device destination not allocated\n";
           return 1;
         }
         break;
@@ -327,16 +324,14 @@ namespace GridKit
       case HOST:
         if (cpu_updated_[0] == false)
         {
-          out::error() << "Vector::getData - requesting data from host"
-                       << " but host data is stale. Returning nullptr ...\n";
+          out::error() << "Vector::getData - host data is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return h_data_;
       case DEVICE:
         if (gpu_updated_[0] == false)
         {
-          out::error() << "Vector::getData - requesting data from device"
-                       << " but device data is stale. Returning nullptr ...\n";
+          out::error() << "Vector::getData - device data is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return d_data_;
@@ -364,16 +359,14 @@ namespace GridKit
       case HOST:
         if (cpu_updated_[0] == false)
         {
-          out::error() << "Trying to get data on the host, but host data is out of date!\n"
-                       << "Use syncData function to sync host data with the device data!\n";
+          out::error() << "Vector::getData - host data is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return h_data_;
       case DEVICE:
         if (gpu_updated_[0] == false)
         {
-          out::error() << "Trying to get data on the device, but device data is out of date!\n"
-                       << "Use syncData function to sync device data with the host data!\n";
+          out::error() << "Vector::getData - device data is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return d_data_;
@@ -405,8 +398,7 @@ namespace GridKit
 
       if (k_ <= j)
       {
-        out::error() << "Trying to get data for vector " << j << " in multivector"
-                     << " but there are only " << k_ << " vectors!\n";
+        out::error() << "Vector::getData - vector index " << j << " out of range, multivector has only " << k_ << " vectors\n";
         return nullptr;
       }
 
@@ -415,18 +407,14 @@ namespace GridKit
       case HOST:
         if (cpu_updated_[j] == false)
         {
-          out::error() << "Trying to get data for vector " << j << " on the host, "
-                       << "but host data is out of date!\n"
-                       << "Use syncData function to sync host data with the device data!\n";
+          out::error() << "Vector::getData - host data for vector " << j << " is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return &h_data_[j * n_size_];
       case DEVICE:
         if (gpu_updated_[j] == false)
         {
-          out::error() << "Trying to get data for vector " << j << " on the device, "
-                       << "but device data is out of date!\n"
-                       << "Use syncData function to sync device data with the host data!\n";
+          out::error() << "Vector::getData - device data for vector " << j << " is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return &d_data_[j * n_size_];
@@ -455,8 +443,7 @@ namespace GridKit
 
       if (k_ <= j)
       {
-        out::error() << "Trying to get data for vector " << j << " in multivector"
-                     << " but there are only " << k_ << " vectors!\n";
+        out::error() << "Vector::getData - vector index " << j << " out of range, multivector has only " << k_ << " vectors\n";
         return nullptr;
       }
 
@@ -465,18 +452,14 @@ namespace GridKit
       case HOST:
         if (cpu_updated_[j] == false)
         {
-          out::error() << "Trying to get data for vector " << j << " on the host, "
-                       << "but host data is out of date!\n"
-                       << "Use syncData function to sync host data with the device data!\n";
+          out::error() << "Vector::getData - host data for vector " << j << " is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return &h_data_[j * n_size_];
       case DEVICE:
         if (gpu_updated_[j] == false)
         {
-          out::error() << "Trying to get data for vector " << j << " on the device, "
-                       << "but device data is out of date!\n"
-                       << "Use syncData function to sync device data with the host data!\n";
+          out::error() << "Vector::getData - device data for vector " << j << " is stale. Perhaps you need to call syncData?\n";
           return nullptr;
         }
         return &d_data_[j * n_size_];
@@ -513,16 +496,14 @@ namespace GridKit
       {
         if (gpu_updated_[i] != all_gpu_updated)
         {
-          out::error() << "Trying to sync all multivector data on the device,"
-                       << " but individual vectors were updated differently.\n"
-                       << "Use syncData function for individual vectors instead!\n";
+          out::error() << "Vector::syncData - inconsistent update state across device columns.\n"
+                       << "Use syncData(j, memspace) for individual vectors\n";
           return 1;
         }
         if (cpu_updated_[i] != all_cpu_updated)
         {
-          out::error() << "Trying to sync all multivector data on the host,"
-                       << " but individual vectors were updated differently.\n"
-                       << "Use syncData function for individual vectors instead!\n";
+          out::error() << "Vector::syncData - inconsistent update state across host columns.\n"
+                       << "Use syncData(j, memspace) for individual vectors\n";
           return 1;
         }
       }
@@ -532,12 +513,12 @@ namespace GridKit
       case DEVICE: // cpu -> gpu
         if (gpu_updated_[0])
         {
-          out::error() << "Trying to sync device, but device already up to date!\n";
+          out::error() << "Vector::syncData - device already up to date\n";
           return 1;
         }
         if (!cpu_updated_[0])
         {
-          out::error() << "Trying to sync device with host, but host is out of date!\n";
+          out::error() << "Vector::syncData - host data is stale, cannot sync to device\n";
           return 1;
         }
         if (d_data_ == nullptr)
@@ -551,12 +532,12 @@ namespace GridKit
       case HOST: // gpu -> cpu
         if (cpu_updated_[0])
         {
-          out::error() << "Trying to sync host, but host already up to date!\n";
+          out::error() << "Vector::syncData - host already up to date\n";
           return 1;
         }
         if (!gpu_updated_[0])
         {
-          out::error() << "Trying to sync host with device, but device is out of date!\n";
+          out::error() << "Vector::syncData - device data is stale, cannot sync to host\n";
           return 1;
         }
         if (h_data_ == nullptr)
@@ -598,12 +579,12 @@ namespace GridKit
       case DEVICE: // cpu->gpu
         if (gpu_updated_[j])
         {
-          out::error() << "Trying to sync device, but device already up to date!\n";
+          out::error() << "Vector::syncData - device already up to date\n";
           return 1;
         }
         if (!cpu_updated_[j])
         {
-          out::error() << "Trying to sync device with host, but host is out of date!\n";
+          out::error() << "Vector::syncData - host data is stale, cannot sync to device\n";
           return 1;
         }
         if (d_data_ == nullptr)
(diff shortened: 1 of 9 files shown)


GridKit/LinearAlgebra/Vector/Vector.hpp
```diff
@@ -0,0 +1,81 @@
...
+     *    flags for HOST and DEVICE) or not, depending on how it is used.
+     *
+     * @author Kasia Swirydowicz <kasia.swirydowicz@pnnl.gov>
+     * @author Slaven Peles <peless@ornl.gov>
+     */
+    template <typename ScalarT, typename IdxT>
+    class Vector
+    {
+    public:
+      Vector(IdxT n);
```
PhilipFackler, line 35: This would need a default constructor for use in the phasor dynamics initialization.

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    template <typename ScalarT, typename IdxT>
+    int Vector<ScalarT, IdxT>::setToConst(IdxT j, ScalarT C, memory::MemorySpace memspace)
+    {
+      using namespace memory;
+      switch (memspace)
+      {
+      case HOST:
+        if (h_data_ == nullptr)
+        {
+          h_data_        = new ScalarT[n_capacity_ * k_];
```
PhilipFackler: It doesn't make sense for allocating to be part of the semantics of this function. It certainly wouldn't be obvious to me as a user that this might happen. I'm used to having no implicit memory operations, only explicit (the user has to say so). This should at least be documented if not reconsidered.

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1039 @@
...
+     * @param[in] memspace - Memory space (HOST or DEVICE)
+     *
+     * @warning This function DOES NOT ALLOCATE any data, it only assigns the
+     * pointer.
+     *
+     * @warning This is an expert level method. Use only if you know what
+     * you are doing.
+     */
+    template <typename ScalarT, typename IdxT>
+    int Vector<ScalarT, IdxT>::setData(ScalarT* data, memory::MemorySpace memspace)
```
PhilipFackler, line 120: This would make more sense as part of a constructor. An instance should be constructed with ownership and data (if non-owning) for host and device respectively. Ownership should not be allowed to change. And I think the data you're wrapping in the non-owning case shouldn't be allowed to change either. This function wouldn't exist in that case.

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    template <typename ScalarT, typename IdxT>
+    int Vector<ScalarT, IdxT>::allocate(memory::MemorySpace memspace)
+    {
+      using namespace memory;
+      switch (memspace)
+      {
+      case HOST:
+        delete[] h_data_;
+        h_data_        = new ScalarT[n_capacity_ * k_];
+        owns_cpu_data_ = true;
```
PhilipFackler: This ignores the case where we're already non-owning. See my comment above. If ownership was settled at construction there could be tighter invariants for the rest of these functions.

PhilipFackler (commented):
For an all-in-one solution like this, I would prefer some revisions as explained in my line comments. None of those is a deal-breaker if this is meant only for "advanced" use-cases, but I still think it would be helpful.

GridKit/LinearAlgebra/Vector/Vector.hpp
pelesh: With default constructor, we would need allocate and/or reallocate methods, as well.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: Yes, you are right, this should be removed.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: I think it would be good to have `allocate` and `reallocate` methods, both refusing to operate on data if not owned by the vector object.
- `allocate` should return error message and do nothing if data pointer is not null.
- `reallocate` should check if the new size is larger than capacity and if so delete previous and allocate new data.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: I agree, changing ownership after construction is looking for trouble.

This method was experimental and should not be a part of the vector class.

pelesh: This method was used to get a vector from a multivector. Better solution is to construct non-data-owning vector object instead, or create a separate `VectorView` object as @lukelowry suggested.

pelesh:
> For an all-in-one solution like this, I would prefer some revisions as explained in my line comments. None of those is a deal-breaker if this is meant only for "advanced" use-cases, but I still think it would be helpful.

All of your suggestions are spot on and should be incorporated before this PR is merged.

GridKit/LinearAlgebra/Vector/Vector.hpp
PhilipFackler: If every bus/component has a constant (hard-coded) `size_` then the default constructor is not necessary.

If some components might not know their size until run time, it would make more sense to have a default-constructed vector from the component constructor and then just replace the instance in the `allocate` method when you know the size (`y_ = Vector(n);`), rather than "reallocate"ing.

But again, if this will never happen, we don't have to worry about it.

pelesh: If we add option to fault each bus/component then there will be not one but two different hard-coded sizes. One for clear and another for faulted object. I can't think of anything else at the moment. @lukelowry or @abirchfield may have additional use cases where bus/component might need to change its model size.

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1039 @@
...
+     * @return pointer to the vector data (HOST or DEVICE). In case of multivectors,
+     * vectors are stored column-wise.
+     *
+     * @note This function gives you access to the pointer, not to a copy.
+     * If you change the values using the pointer, the vector values will
+     * change too. Make sure to use setDataUpdated function to set the update
+     * flags correctly after changing the values.
+     */
+    template <typename ScalarT, typename IdxT>
+    ScalarT* Vector<ScalarT, IdxT>::getData(memory::MemorySpace memspace)
```
lukelowry, line 318: `getData` and its overloads may trigger a memory copy, which I would advise we avoid at all costs once the system is initialized. I see how we could functionally achieve zero-copy with this `Vector` class, but I do not want to give Components the chance to copy data in this way.

To prevent our future selves from having access to functions like this in simulation-time functions like `evaluateResidual,` I strongly recommend that we enforce zero-copy *de facto* after initialization to force us to design scalably. (This is the motivation of `VectorView,` which I'll draft up so you can get a better picture of what I am imagining)

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1039 @@
...
+     */
+    template <typename ScalarT, typename IdxT>
+    const ScalarT* Vector<ScalarT, IdxT>::getData(memory::MemorySpace memspace) const
+    {
+      using memory::DEVICE;
+      using memory::HOST;
+
+      switch (memspace)
+      {
+      case HOST:
```
lukelowry, line 360: I think this was mentioned in the meeting earlier today, but the `Host` and `Device` vocabulary and GPU vocab are not desirable IMHO

lukelowry (commented):
I think we need a zombie approach if we do merge this implementation. We should take the best parts of this and graft them onto a more scalable implementation that can isolate the Component in memory (is "closure" the right word here?).

GridKit/LinearAlgebra/Vector/Vector.hpp
lukelowry: Everything I can think of would be deterministically sized at initialization. I am struggling to think of a case where we would resize the vector after initialization. I think the only exception I can see that would be tangible to resize after initialization is the network itself or related operations by the SystemModel

pelesh: This is a common issue with all scientific software that I can think of. You need to compute your workspace before you can allocate it. If you know the size of the vector is 124 a priori, then you can instantiate and allocate it like this:
```c++
Vector v(124);
v.allocate(HOST);
```
Otherwise you need to do something like
```c++
Vector v;

// compute size n

v.resize(n);
v.allocate(HOST);
```
The `resize` method is there just to give you some flexibility.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: Zero copy will be enforced.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: Domain folks do not need to see that. We could implement all functions with `HOST` memory space as the default, e.g.
```c++
int allocate(MemSpace m = HOST);
```
That way, if you don't specify memory space everything will run on CPU (host). Only if you want to run something on GPU, you would need to add memory space argument, but in that case you would already know what the host and device are.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: I restricted this function to be able to operate only on vectors with data not allocated. Intended use would be:
```c++
Vector v(n)
...
double* data = new double[n];
...
v.setData(data);
```
The danger is still that user needs to set data that matches the vector size. This is a makeshift solution to create vector views for now.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: If we are allowed to use a default constructor, the ownership cannot be decided on construction. I suggest we limit instances/methods that change ownership instead. I think this is sufficiently addressed in e24f31de66e09253ca4b14492d6a7bec39b9f756.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: Fixed in https://github.com/ORNL/GridKit/pull/418/commits/fdc2db91d31d018c0867656bb0ad7e8f93a0327f.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: Addressed in https://github.com/ORNL/GridKit/pull/418/commits/368e363a1e15007c2879789cdfc01fc1fe0717e5

tests/UnitTests/LinearAlgebra/Vector/VectorTests.hpp
```diff
@@ -0,0 +1,436 @@
...
+#include <iomanip>
+#include <iterator>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include <GridKit/LinearAlgebra/MemoryUtils.hpp>
+// #include <resolve/vector/Vector.hpp>
+// #include <resolve/vector/VectorHandler.hpp>
+// #include <resolve/workspace/LinAlgWorkspace.hpp>
```
nkoukpaizan: I recommend removing references to `resolve`, even if commented.

tests/UnitTests/LinearAlgebra/Vector/runVectorTests.cpp
```diff
@@ -0,0 +1,52 @@
...
+    result += test.copyFromExternal(50);
+    // result += test.copyToExternal(50);
+
+    result += test.resize(100, 50);
+
+    result += test.setToConst(50);
+    result += test.setToConst(50);
+  }
+
+#ifdef GRIDKIT_USE_GPU
```
nkoukpaizan: I don't recall that we've defined this yet. Now may be a good time. Need to clarify `GRIDKIT_ENAGLE_` vs `GRIDKIT_USE_` if we choose to differentiate between CMake options and pre-cprocessor directives.

GridKit/LinearAlgebra/Vector/Vector.cpp
PhilipFackler: This is why I think it should be part of the constructor.

```c++
Vector v;
...
double* data = new double[n];
...
v = Vector(n, data);
```

GridKit/LinearAlgebra/Vector/Vector.hpp
PhilipFackler: If the only use for `resize` is to set a size for an initially empty vector, then just reconstruct it.

```c++
Vector v;

// compute size n

v = Vector(n);
v.allocate(HOST);
```

And I would go further, based on what I understand the use-case for this class to be. Ideally (if possible) the internal configuration of the vector (not the elements) should be "immutable". From this perspective I think allocation should happen at construction also. The constructor should specify the entire configuration for the vector. If that needs to change, then you're talking about a different vector.

Unless there is a really good reason to make it truly dynamic, you're just adding ways to shoot yourself in the foot.

GridKit/LinearAlgebra/Vector/Vector.cpp
```diff
@@ -0,0 +1,1007 @@
...
+        setHostUpdated(true);
+        setDeviceUpdated(false);
+        break;
+      case DEVICE:
+        setHostUpdated(false);
+        setDeviceUpdated(true);
+        break;
+      }
```
PhilipFackler, line 175: What if the memspace not passed to the function had already been tagged as updated?

GridKit/LinearAlgebra/Vector/Vector.cpp
lukelowry: okay, sounds great!

GridKit/LinearAlgebra/Vector/Vector.cpp
lukelowry: thank you that was my biggest concern

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: It will be tagged as out of date. This is desired behavior.

nkoukpaizan (approved):
A few minor comments. 

I'm not convinced the class in GridKit needs to mirror Re::Solve, and I tend to prefer solutions that are driven by the application needs first, but I'm willing to see what happens when we try to use this in GridKit.

GridKit/LinearAlgebra/Vector/Vector.hpp
pelesh: @PhilipFackler, please create an issue detailing your suggested improvement.

tests/UnitTests/LinearAlgebra/Vector/runVectorTests.cpp
pelesh: I would use the same names for both.

tests/UnitTests/LinearAlgebra/Vector/runVectorTests.cpp
```diff
@@ -0,0 +1,52 @@
...
+    GridKit::Testing::VectorTests test(GridKit::memory::DEVICE);
+
+    std::cout << "Running Testing on GPU:\n";
+    result += test.vectorConstructor(50, 5);
+    result += test.vectorConstructor(50);
+
+    result += test.setData(50);
+
+    result += test.copyFromExternal(50);
+    // result += test.copyToExternal(50);
```
shakedregev, line 40: Why is this test commented out?

shakedregev (commented):
Address the comment and clarify that this is ready for testing.

GridKit/LinearAlgebra/Vector/Vector.cpp
pelesh: As I said, this is a makeshift solution for this PR. My preference would be to remove this method altogether if better replacement exists. Please submit an issue with your proposed solution.

tests/UnitTests/LinearAlgebra/Vector/VectorTests.hpp
pelesh: Fixed in https://github.com/ORNL/GridKit/pull/418/commits/d2625f15221fc66c992b65abcf2828aeb7132d2e

tests/UnitTests/LinearAlgebra/Vector/runVectorTests.cpp
pelesh: Fixed in https://github.com/ORNL/GridKit/pull/418/commits/d2625f15221fc66c992b65abcf2828aeb7132d2e

tests/UnitTests/LinearAlgebra/Vector/VectorTests.hpp
```diff
@@ -0,0 +1,433 @@
...
+        bool success = true;
+
+        if (memspace_ == memory::DEVICE)
+        {
+          x.syncData(memory::HOST);
+        }
+
+        for (IdxT i = 0; i < x.getSize(); ++i)
+        {
+          // std::cout << x->getData("cpu")[i] << "\n";
```
abdourahmanbarry, line 383: I believe this commented line was used for debugging. We can remove it before merging.

PR #423, opened 2026-05-31
Remove GridElement class

pelesh:
## Description
 
 Simplify inheritance structure for `PhasorDynamics` buses and components.  
 
 ## Proposed changes
 
Remove intermediary `GridElement` class and let BusBase inherit directly from `Evaluator`. While this does introduce some minor code duplication, we expect bus and component classes to get more specialized and the common code between them to change going forward. As we move to implement EMT models, having simpler inheritance structure will help, as well.

Also, I expect `BusBase` class to be removed altogether and have `Bus` inherit directly from `Evaluator`. Class `BusInfinite` should be a standalone class, similar to `SignalNode` and not inherit from `Evaluator` at all.

 
 ## Checklist
 

- [x] All tests pass.
 
 ## Further comments


PhilipFackler (approved):
It works :)

PR #438, opened 2026-06-05
Use system base for governor-machine power signals

lukelowry:
## Description

Use system-base values for governor-machine mechanical power signals.

This is needed because governor ratings are not generally the same as the associated machine MVA base. Governor `pmech` signals, or equivalent governor-machine signals, should therefore specify their power base (Turbine rating in this context). This PR introduces `Trate` as the TGOV1 component base and converts at the governor signal boundary.

## Proposed changes

- Add `Trate` as the TGOV1 component-base parameter.
- Convert TGOV1 mechanical power between system base and TGOV1 component base.
- Convert GENROU and GENSAL `pmech` signal interaction to use system-base values.
- Update example cases so TGOV1 `Trate` matches the associated machine `mva`.
- Update TGOV1 tests and tiny example inputs for the new base semantics.

## Checklist

Changelog changes N/A

## Further comments

No functional change. This is just a component contract change that I need to establish before I add more complex governors

I made this change as small as possible to fix the TGOV1 issue. Long term a more general base conversion might be prefered but this by no means locks us into a convention.


GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
```diff
@@ -491,10 +491,10 @@ namespace GridKit
                + G_ * (vd * -std::cos(delta) + vq * std::sin(delta));
 
       ScalarT Te = y_[12];
-      pmech_set_ = Te;
+      pmech_set_ = toSystemBase(Te);
       if (signals_.template isAttached<GenrouExternalVariables::PM>())
```
pelesh, line 499: The code is well written and almost self-documenting but it would be good to add comments here describing what is done here (a unit conversion).

pelesh (approved):
Good to merge. I would suggest adding more comments to the code where appropriate.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROUwS/GenrouImpl.hpp
lukelowry: Done. Rebased and added comments where conversion occurs so theres no confusion about what each is doing moving forward

PR #439, opened 2026-06-05
Support zero IEEET1 transducer time constant

lukelowry:
## Description

Support `Tr = 0` in the IEEET1 exciter by treating the sensed terminal voltage state as algebraic when the transducer time constant is zero.

This removes the need for small artificial `Tr` values in validation and example cases.

## Proposed changes

- Tag IEEET1 `Vts` as algebraic when `Tr = 0`.
- Update the IEEET1 README equations to match the implementation and standard notation
- Update IEEET1 example, validation, and parser-test data so prior tiny `Tr` placeholders are zero.
- Add unit-test for the `Tr = 0` algebraic behavior.

## Checklist

Changelog changes N/A

## Further comments

Early development known issue that did not matter until detailed validation of the larger models.


GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -261,15 +261,15 @@ namespace GridKit
...
-        tag_[0] = true;  // y0 - vts  - Sensed term volt
-        tag_[1] = true;  // y1 - vr   - Voltage reg
-        tag_[2] = true;  // y2 - efdp - Efd pre mult
-        tag_[3] = true;  // y3 - vfx  - Exciter feedback
-        tag_[4] = false; // y4 - vtr  - Term Volt Err
-        tag_[5] = false; // y5 - vf   - Feedback volt
-        tag_[6] = false; // y6 - ve   - Excit. Cntrl Volt
-        tag_[7] = false; // y7 - efd  - Efd
-        tag_[8] = false; // y8 - ksat - Saturation
+        tag_[0] = (Tr_ != 0.0); // y0 - vts  - Sensed term volt
```
nkoukpaizan: There are potential issues with comparing a floating point value to 0 like this.

lukelowry: I agree, so I added the smallest possible inequality using `epsilon()`.

There are many models that use zero-valued parameters to control the model's behavior (e.g., a control block bypass). A good way to describe them would be 'parametrically piecewise'. In the fixed time step solvers like PowerWorld, it doesn't even allow $T_r\leq \Delta t/2$ for numeric stability, but of course, we are not constrained by the time step here.  We should probably establish a convention for these 'parametrically piecewise' models as they become more common.

We might actually want a way to wire in the minimum time step for IDA into the algebraic flag $T_r\leq \Delta t$, so if we ever allow the user to set a minimum or fixed time step, this model is still stable. This can probably be done for a lot of the equations in other models that take this form (as Power World does). cc @Steven-Roberts

lukelowry:
Added a fix to the exciter scaling now that $\sigma(x)$ is fixed

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -262,15 +263,17 @@ namespace GridKit
...
-        tag_[2] = true;  // y2 - efdp - Efd pre mult
-        tag_[3] = true;  // y3 - vfx  - Exciter feedback
-        tag_[4] = false; // y4 - vtr  - Term Volt Err
-        tag_[5] = false; // y5 - vf   - Feedback volt
-        tag_[6] = false; // y6 - ve   - Excit. Cntrl Volt
-        tag_[7] = false; // y7 - efd  - Efd
-        tag_[8] = false; // y8 - ksat - Saturation
+        static constexpr RealT ZERO_TOL = std::numeric_limits<RealT>::epsilon();
+
+        tag_[0] = (Tr_ > ZERO_TOL); // y0 - vts  - Sensed term volt
```
pelesh: I would use here the `isEqual` function we defined for testing purposes. That way we have a more uniform approach across GridKit to handling floating point numbers comparisons.
```suggestion
        tag_[0] = (!isEqual(Tr_, ZERO<RealT>)); 
```
Also, a better comment explaining what has been done here would be helpful.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/README.md
```diff
@@ -73,7 +73,7 @@ Generally, this system has two solutions. The non-extraneous solution is as foll
 
 Symbol    | Units  | Description                       | Note
 ----------|--------|-----------------------------------|-------
-$V_{ts}$  | [p.u.] | Sensed terminal voltage           |
+$V_{ts}$  | [p.u.] | Sensed terminal voltage           | Algebraic when $T_R=0$
 $V_R$     | [p.u.] | Voltage regulator                 |
```
pelesh: This will create an issue when implementing interface for equations in Hessenberg form. I think a proper solution would be instantiating different models rather than having switch from algebraic to differential using a parameter. When an equation is turned from differential to algebraic, the numerical treatment may be different. In case of IDA solver, it does not matter, but any solver taking equations in Hessenberg form would require different methods for evaluating model equations.

pelesh (commented):
This is a patchwork solution that works only for DAEs defined in fully implicit form. It will not work for any solver taking equations in Hessenberg form. 

If this is essential feature needed, I suggest creating an issue flagging this model as potential blocker for Hessenberg-form solvers before merging it to `develop`.

Also, some minor comments below.

pelesh:
CC @alexander-novo

GridKit/Model/PhasorDynamics/Exciter/IEEET1/README.md
lukelowry: Could you expound on the meaning of 'different models'? As in we have Ieeet1A and Ieeet1B? That feels like it could be hard to maintain, possibly. I am not opposed, however. 

I would say that this type of conditional algebraic/differential is very common, and in the larger cases, you could have some exciters/models with $T=0$ and the other half $T>0$. If we did a variant model for each variable like this, we would need a permutation for each, which could be quite a few....

If you could clarify intent

pelesh: We want to eventually have separate methods for evaluating differential and algebraic equations, so instead of one residual function for implicit DAEs, we would have differential equations rhs and algebraic equations residual functions. In that case when the equation changes from differential to algebraic, it should be implemented in different method. 

Therefore, we would need separate methods for cases when $T=0$ and $T>0$. The boolean variable `Tr_ ! = 0.0` would then decide which pair of methods (differential equation rhs and algebraic equation residual) to use.

lukelowry: That is helpful, thank you. I will open an issue for this and list which models need to be addressed. 

I am in favor of this form for now, so we don't have to have duplicate equations or helper functions.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
lukelowry: Fixed. Thank you

GridKit/Model/PhasorDynamics/Exciter/IEEET1/README.md
lukelowry: Opened  #463 for this issue

lukelowry:
I believe I resolved the issue by adding an auxiliary algebraic variable, as suggested in today's technical discussion.

@pelesh rebased as well

lukelowry:
Okay, per discussion in #463, I changed it so $T_R\geq 0.001$ and made that a constant in the header of the model.

Rebased and good now!

nkoukpaizan (approved):
This looks clean to me.

pelesh (approved):

PR #440, opened 2026-06-06
Use ports for PhasorDynamics signal connectivity

### LINKED ISSUE #407: Associate signals with ports rather than variables
### Summary

As discussed in the comments in #401 change the enums that are used to define signal nodes on a component.

### Rationale

Signal nodes should be associated with input/output ports rather than internal/external variables.

### Description

- Split the `<component>Ports` enum into `<component>InputPorts`, `<component>OutputPorts`, and `<component>BusPorts` (not sure about the name of that last one).
- Add a `SIZE` member to the same (required for use in `ComponentSignals`)
- Use input/output port enums as template arguments to `ComponentSignals`
- Modify `SignalComponents` to use `SIZE` rather than `MAXIMUM` (more accurate)
- Update input file documentation and component parser to specify ports separately

### Additional information

For this change I would keep `ComponentSignals` basically as it is. But in future, it needs to be reworked a little to make it easier to use (and understand) in the code. This will be a good first step.

lukelowry:
## Description

This PR updates PhasorDynamics component wiring so signal nodes are associated with input/output ports instead of internal/external variable enums.

Closes #407

@PhilipFackler, what are your thoughts? Managed to do it with no `*.case.json` changes, which lets us do the INPUT_FORMAT changes and documentation with a few changes in code once we get to it

## Proposed changes

- Split component connection data into terminals, input ports, and output ports.
- Keep the existing flat JSON `"ports"` format, but parse entries into the new internal maps.
- Update `ComponentSignals`, `SystemModel`, components, examples, and tests to use port enums for signal wiring.
- Remove the old variable-enum signal helpers.

## Checklist

- [x] All tests pass for the focused coverage.
- [ ] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR.

## Further comments

I organized the commits so we can roll back if the scope is too big, but it was mechanical and straightforward after the main changes were in.

### DIFF
--- a/GridKit/Model/PhasorDynamics/ComponentDataJSONParser.hpp
+++ b/GridKit/Model/PhasorDynamics/ComponentDataJSONParser.hpp
@@ -1,7 +1,8 @@
 #pragma once
 
 #include <sstream>
-#include <stdexcept>
+#include <string>
+#include <type_traits>
 
 #include <magic_enum/magic_enum.hpp>
 #include <nlohmann/json.hpp>
@@ -20,12 +21,23 @@ namespace GridKit
     template <typename RealT,
               typename IdxT,
               typename Parameters,
-              typename Ports,
+              typename Terminals,
+              typename InputPorts,
+              typename OutputPorts,
               typename MonitorableVariables>
       requires std::is_enum_v<Parameters>
-               && std::is_enum_v<Ports>
+               && std::is_enum_v<Terminals>
+               && std::is_enum_v<InputPorts>
+               && std::is_enum_v<OutputPorts>
                && std::is_enum_v<MonitorableVariables>
-    void from_json(const json& j, ComponentData<RealT, IdxT, Parameters, Ports, MonitorableVariables>& c)
+    void from_json(const json&                          j,
+                   ComponentData<RealT,
+                                 IdxT,
+                                 Parameters,
+                                 Terminals,
+                                 InputPorts,
+                                 OutputPorts,
+                                 MonitorableVariables>& c)
     {
       j.at("class").get_to(c.device_class);
 
@@ -72,19 +84,37 @@ namespace GridKit
         }
       }
 
+      // WARNING --- TEMPORARY --- : the C++ model data is
+      // intentionally split into terminals, input ports, and output ports.
+      // This backward-compatible flat JSON "ports" parsing SHOULD NOT become
+      // the supported design; it only avoids changing every tracked case in
+      // this PR. The JSON schema should be split once the ingress changes are merged
       for (auto& raw_port : j.at("ports").items())
       {
-        auto key = magic_enum::enum_cast<Ports>(raw_port.key());
-        if (key.has_value())
+        auto terminal = magic_enum::enum_cast<Terminals>(raw_port.key());
+        if (terminal.has_value() && terminal.value() != Terminals::SIZE)
         {
-          raw_port.value().get_to(c.ports[key.value()]);
+          raw_port.value().get_to(c.terminals[terminal.value()]);
+          continue;
         }
-        else
+
+        auto input_port = magic_enum::enum_cast<InputPorts>(raw_port.key());
+        if (input_port.has_value() && input_port.value() != InputPorts::SIZE)
         {
-          Log::error() << "\n\tInvalid port mapping: \"" << raw_port.key()
-                       << "\" has no value." << error_context.str()
-                       << std::endl;
+          raw_port.value().get_to(c.input_ports[input_port.value()]);
+          continue;
         }
+
+        auto output_port = magic_enum::enum_cast<OutputPorts>(raw_port.key());
+        if (output_port.has_value() && output_port.value() != OutputPorts::SIZE)
+        {
+          raw_port.value().get_to(c.output_ports[output_port.value()]);
+          continue;
+        }
+
+        Log::error() << "\n\tInvalid port mapping: \"" << raw_port.key()
+                     << "\" has no value." << error_context.str()
+                     << std::endl;
       }
 
       if (j.contains("mon"))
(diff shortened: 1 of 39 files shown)


GridKit/Model/PhasorDynamics/ComponentSignals.hpp
```diff
@@ -11,20 +11,6 @@ namespace GridKit
...
-      MAXIMUM
-    };
-
-    /// Concept requiring an enum to have a `MAXIMUM` variant and that it have
-    /// the underlying type of `size_t`. This does not ensure the variant is
-    /// the actual maximum
-    template <typename T>
-    concept EnumHasMaximumValueAndIsSizeT = std::is_enum_v<T>
-                                            && std::is_same_v<std::underlying_type_t<T>, size_t>
-                                            && requires { T::MAXIMUM; };
```
superwhiskers, line 26: i think we should keep this concept. above, in the json parsing code, you rely upon the existence of a `SIZE` variant in the enums. just swap the `T::MAXIMUM` requirement for `T::SIZE`.

additionally, there's many cases in this diff where you don't have any ports, and just leave an empty enum for one of the input/output/terminal port enums. it may be beneficial to keep the `NoVariables` enum---perhaps renamed---for those cases. that's less important, though.

superwhiskers (commented):
i'd recommend we make this one change. everything else seems fine. i have not attempted to compile or test the diff.

GridKit/Model/PhasorDynamics/ComponentSignals.hpp
lukelowry: Thank you for the feedback. Regarding your second comment about `NoVariables`: My rational was that some of these models can be implemented with or without internal variables (e.g. `Branch` and `LoadZIP`). I agree that having the model internal variables enum just contain `SIZE` is a bit useless, but it makes adding variables/ports easier and makes the model implementation more copy-and-paste-able.

What are your thoughts?

superwhiskers: i'm of the opinion that boilerplate exists to be minimized, so i don't view a copy-and-pasteable implementation to be a good thing. however, i can see how it could be useful for people unfamiliar with the codebase or c++ generally to be able to easily see how to add internal variables in their own model without cross-referencing more.

the former opinion is what leads me to suggest retaining a `NoVariables`-like placeholder. if you disagree with this stance, and want to make it simpler to copy-and-paste an implementation as a template wholesale, then go ahead with that. this was a minor concern anyway.

pelesh:
As far as the naming conventions go, I would suggest using either "port" or "terminal" for the connectors but not a mixture of the two. The way I see it, we have three types of terminals (ports): bus, signal out and signal in.

I also suggest when making a wholesale name changes to have a purely name changing pull request to separate it from any design changes in the code. This makes reviewing PRs and tracking back changes easier.

PhilipFackler:
> As far as the naming conventions go, I would suggest using either "port" or "terminal" for the connectors but not a mixture of the two. The way I see it, we have three types of terminals (ports): bus, signal out and signal in.

Maybe this should be a question for users? I don't know power systems, but we do treat the bus connection differently. If it is conceptually different from signal connections, then it would make sense (to me) to call it something else to distinguish the two. But at the end of the day, what makes the most intuitive sense for users is what matters in this case.

lukelowry:
@pelesh @PhilipFackler This branch has been updated to only have the naming changes, now following the pattern:
- `*SignalInputs`
- `*SignalOutputs`
- `*Buses`

Ready for review. mostly mechanical change

PhilipFackler:
@lukelowry Just double-checking. I thought from our meeting, it was decided to use the word ports for all of them:
- *SignalInputPorts
- *SignalOutputPorts
- *BusPorts

It appears we did not write it down, though. :)

@pelesh do you remember?

pelesh:
> @lukelowry Just double-checking. I thought from our meeting, it was decided to use the word ports for all of them:
> 
> * *SignalInputPorts
> * *SignalOutputPorts
> * *BusPorts
> 
> It appears we did not write it down, though. :)
> 
> @pelesh do you remember?

We agreed to have 
* `*SignalInputs`
* `*SignalOutputs`
* `*Buses`

and if there is a common name (i.e. base class) needed that would be `*Ports`.

PhilipFackler:
Alright, sorry about my confusion.

pelesh:
By looking at the implementation, the three are very different, so it doesn't seem we need `Ports` object in the code, but in the documentation we could refer to them as ports for brevity.

GridKit/Model/PhasorDynamics/ComponentDataJSONParser.hpp
```diff
@@ -72,19 +84,37 @@ namespace GridKit
         }
       }
 
+      // WARNING --- TEMPORARY --- : the C++ model data is
+      // intentionally split into buses, signal inputs, and signal outputs.
+      // This backward-compatible flat JSON "ports" parsing SHOULD NOT become
+      // the supported design; it only avoids changing every tracked case in
+      // this PR. The JSON schema should be split once the ingress changes are merged
```
PhilipFackler, line 91: Thanks for leaving this comment. Once we get the code settled, we can have an effort to update input file format.

PhilipFackler (approved):
Looks good. Tested locally.

GridKit/Model/PhasorDynamics/ComponentDataJSONParser.hpp
lukelowry: absolutely! We will want to make some changes for the official v0.1 of the phasor dynamics input format, and I think this is one of the more important things to do before we release next, so the code, docs, and input format are all consistent :)!

pelesh (approved):

PR #437, opened 2026-06-05
Separate compilation units and targets for `SystemModel`

PhilipFackler:
## Description
 
Move `SystemModel` implementation to dedicated compilation unit(s).
 
 _Closes #408_
 
 ## Proposed changes
 
Move implementation code for `SystemModel` to `SystemModelImpl.hpp` and declares class template instantiations in `SystemModel.cpp` and `SystemModelDependencyTracking.cpp`.
 
 ## Checklist
 
- [X] All tests pass.
- [X] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [X] The new code follows GridKit™ style guidelines.
- [N/A] There are unit tests for the new code.
- [N/A] The new code is documented.
- [X] The feature branch is rebased with respect to the target branch.
- [N/A] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.

PR #444, opened 2026-06-09
Fix phasor_dynamics_systemmodel targets.

nkoukpaizan:
## Description
 
This fixes a bug where PhasorDynamics components symbols are not found when building the `phasor_dynamics_systemmodel` target. (This shows up on macOS, but not on Linux. I thought I had tested #437 on macOS, but evidently not).
 
 ## Proposed changes
 
- Construct the `phasor_dynamics_systemmodel` targets after component models have been constructed.
- Replace the `phasor_dynamics_components` target with `phasor_dynamics_systemmodel` where appropriate.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments


PhilipFackler (approved):
Makes sense. Thanks for catching that.

PR #447, opened 2026-06-11
Nuke COO_Matrix class

nkoukpaizan:
## Description

This removes `COO_Matrix` class.

 ## Proposed changes
- Removed the`COO_Matrix` class
- Replaced the component-level use of `COO_Matrix` with the construction of `CooMatrix`, a thin wrapper around the allocated buffers. 
- Removing the `axpy` method (which was not thread-safe) and deduplication at the component-level required changing how `Bus` and `BusFault` are handled.
- Components now store duplicates of the bus Jacobians instead. These are deduplicated at the system-level.
- BusFault now owns voltage variables, so the indices are always present in the system-level Jacobian, but the values are only stored when `status_`.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
- Nuking `DenseMatrix` will be done in a subsequent PR. It is only used in examples and not in the core. These examples should be converted to uni tests as well.
- A subsequent PR will also consolidate the component-level Jacobian evaluation boilerplate code.


GridKit/Model/PhasorDynamics/Bus/Bus.hpp
```diff
@@ -99,6 +98,35 @@ namespace GridKit
...
+            if (J_rows_buffer_[i] > num_rows)
+            {
+              num_rows = J_rows_buffer_[i];
+            }
+            if (J_cols_buffer_[i] > num_cols)
+            {
+              num_cols = J_cols_buffer_[i];
+            }
+          }
+          coo_jac_ = new CooMatrixT(num_rows, num_cols, nnz_);
```
pelesh, line 119: If rows/columns are zero-indexed, shouldn't it be:
```c++
coo_jac_ = new CooMatrixT(num_rows + 1, num_cols + 1, nnz_);
```
?

GridKit/Model/PhasorDynamics/Bus/Bus.hpp
```diff
@@ -99,6 +98,35 @@ namespace GridKit
...
+          coo_jac_->setDataPointers(J_rows_buffer_, J_cols_buffer_, J_vals_buffer_, LinearAlgebra::memory::HOST);
+        }
```
pelesh, line 121: I would use hijacking constructor here instead of `setDataPointers` method. That way, buffers can be local variables and only `coo_jac_` will remain a member variable.

GridKit/Model/PhasorDynamics/BusFault/BusFaultImpl.hpp
```diff
@@ -161,18 +200,40 @@ namespace GridKit
      */
     template <typename scalar_type, typename index_type>
     __attribute__((always_inline)) int BusFault<scalar_type, index_type>::evaluateBusResidual(
-        [[maybe_unused]] ScalarT* y, [[maybe_unused]] ScalarT* yp, ScalarT* wb, ScalarT* h)
+        ScalarT*                  y,
+        [[maybe_unused]] ScalarT* yp,
+        [[maybe_unused]] ScalarT* wb,
+        ScalarT*                  h)
```
pelesh: I think it would be good idea if inputs are defined as `const`
```suggestion
        const ScalarT*                  y,
        [[maybe_unused]] const ScalarT* yp,
        [[maybe_unused]] const ScalarT* wb,
        ScalarT*                  h)
```

GridKit/Model/PhasorDynamics/Component.hpp
```diff
@@ -189,10 +259,12 @@ namespace GridKit
...
-      MatrixT J_;
-      IdxT*   J_rows_buffer_{nullptr};
-      IdxT*   J_cols_buffer_{nullptr};
-      RealT*  J_vals_buffer_{nullptr};
+      IdxT*       J_rows_buffer_{nullptr};
+      IdxT*       J_cols_buffer_{nullptr};
+      RealT*      J_vals_buffer_{nullptr};
+      IdxT*       map_to_csr_{nullptr};
+      CsrMatrixT* csr_jac_{nullptr};
+      CooMatrixT* coo_jac_{nullptr};
```
pelesh, line 267: Why do we need CSR and COO matrices as component members?

GridKit/Model/PhasorDynamics/Load/LoadZ/LoadZEnzyme.cpp
```diff
@@ -23,63 +23,61 @@ namespace GridKit
...
-      J_.zeroMatrix();
       if (J_rows_buffer_ == nullptr)
       {
-        J_rows_buffer_ = new IdxT[4];
-        J_cols_buffer_ = new IdxT[4];
-        J_vals_buffer_ = new RealT[4];
+        auto size        = static_cast<size_t>(size_);
+        auto bus_size    = static_cast<size_t>(bus_->size());
+        auto buffer_size = size * size + 2 * size * bus_size;
+        J_rows_buffer_   = new IdxT[buffer_size];
```
pelesh, line 34: How is the buffer size calculated? Perhaps we should simply require developer to provide NNZ for now?

pelesh (commented):
As the first step, I would set `nnz_` to constant value upon component instantiation. That should make local Jacobian instantiation and allocation a little bit simpler.

We could also make a unit test that checks `nnz_` set by developer is the same Enzyme would evaluate.

GridKit/Model/PhasorDynamics/Bus/Bus.hpp
nkoukpaizan: You are correct. I had fixed it elsewhere, but forgot to update here. Addressed in [a49092f](https://github.com/ORNL/GridKit/pull/447/commits/a49092f86accd2fada0f8fb06f932c23cac489af), e.g., `num_rows = J_rows_buffer_[i] + 1`.

GridKit/Model/PhasorDynamics/BusFault/BusFaultImpl.hpp
nkoukpaizan: Addressed in [418827f](https://github.com/ORNL/GridKit/pull/447/commits/418827f46c25d2672bb8e2cb5891635af5e0312f).

lukelowry:
On my system, this causes a small slowdown. Is that from increasing the tolerance, or something else? (In DynamicSimulation app)

pelesh (approved):

nkoukpaizan:
> On my system, this causes a small slowdown. Is that from increasing the tolerance, or something else? (In DynamicSimulation app)

The tolerance does change the runtime, but these were the tolerances set in `SystemModel` a while back. It is more likely due to duplicating bus Jacobian entries.

pelesh:
> On my system, this causes a small slowdown. Is that from increasing the tolerance, or something else? (In DynamicSimulation app)

We would need to profile code to answer that. Thanks for reporting possible performance regression. This PR is just replacing some obsolete code.

PR #449, opened 2026-06-12
Use relative errors by default when comparing output files

### LINKED ISSUE #368: Need relative errors in aggregate errors
### Summary

Users should have the option of using relative or absolute tolerance for errors. (Currently `ErrorAggregate`s are computed as absolute errors and must be compared against an absolute tolerance.)

### Rationale

Relative errors are generally more useful than absolute errors.

PhilipFackler:
## Description
 
Enable relative and absolute errors (for CSV comparison) and use relative by default.
 
 _Closes #368_
 
 ## Proposed changes
 
This enables aggregating absolute errors or relative errors. Optional parameters are added to the JSON solver input file. Scaling relative errors depends on the optional parameter "abs_err_threshold" below which the errors will switch to absolute (so we don't divide by zero).
 
 ## Checklist
 
- [X] All tests pass.

### DIFF
--- a/GridKit/Testing/AggregateErrors.hpp
+++ b/GridKit/Testing/AggregateErrors.hpp
@@ -21,31 +21,45 @@ namespace GridKit
      */
     struct TemporalNormAggregate
     {
+      /// Name of variable
       std::string label;
-      double      max{0.0};
-      double      max_time{0.0};
+      /// L-inf norm
+      double      max_value{0.0};
+      /// Time at which max value occurred
+      double      max_value_time{0.0};
+      /// L2 norm
       double      L2{0.0};
 
-      void push(double err, double time)
+      /**
+       * @brief Add a new value for the variable to be aggregated
+       */
+      void push(double val, double time)
       {
-        if (err > max)
+        if (val > max_value)
         {
-          max      = err;
-          max_time = time;
+          max_value      = val;
+          max_value_time = time;
         }
-        L2 += err * err;
+        L2 += val * val;
       }
 
+      /**
+       * @brief Finalize the calculation(s)
+       */
       void wrap()
       {
         L2 = std::sqrt(L2);
       }
 
+      /**
+       * @brief Scale by a reference value (only if the current value is above
+       * the given threshold (used internally for relative errors)
+       */
       void scale(const TemporalNormAggregate& ref, double threshold)
       {
-        if (ref.max > threshold)
+        if (ref.max_value > threshold)
         {
-          max /= ref.max;
+          max_value /= ref.max_value;
         }
         if (ref.L2 > threshold)
         {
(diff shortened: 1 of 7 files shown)


GridKit/Testing/AggregateErrors.hpp
```diff
@@ -11,96 +13,196 @@ namespace GridKit
...
 
+    inline constexpr double DEFAULT_ABS_ERROR_THRESHOLD =
+        std::numeric_limits<double>::epsilon();
+
     /**
-     * @brief Aggregate error data for a single variable
+     * @brief Aggregate norm data for a single variable over time
      */
-    struct ErrorAggregate
+    struct TemporalNormAggregate
```
lukelowry, line 22: I approve of this vocabulary

GridKit/Testing/CSV.hpp
```diff
@@ -20,7 +20,11 @@ namespace GridKit
      *
      * @note This assumes a "time" variable is always in the first column.
      */
-    ErrorSet compareCSV(const std::string& f_a, const std::string& f_b);
+    std::unique_ptr<ErrorSet> compareCSV(
+        const std::string& test_file,
+        const std::string& reference_file,
+        ErrorType          error_type    = ErrorType::RELATIVE,
+        double             abs_threshold = DEFAULT_ABS_ERROR_THRESHOLD);
```
lukelowry, line 27: Instead of `abs_threshold` how about `err_threshold` since the type can change?

lukelowry (commented):
Just some comments. Thank you for making my life easier for validation. Much appreciated

GridKit/Testing/CSV.hpp
PhilipFackler: This is specific to computing relative error. If the reference value is lower than the threshold, you don't scale, so you switch to absolute.

GridKit/Testing/AggregateErrors.hpp
```diff
@@ -11,96 +13,196 @@ namespace GridKit
...
     {
       std::string label;
-      double      max_error{0.0};
-      double      max_error_time{0.0};
-      double      error_norm_L2{0.0};
+      double      max{0.0};
+      double      max_time{0.0};
+      double      L2{0.0};
```
pelesh: Please revert to more verbose names and add comments explaining what each variable represents. Also naming something `max` may not be the best choice given there is a math function `max`.

pelesh (approved):
Local variables in functions need to be better commented. Otherwise good to merge.

GridKit/Testing/AggregateErrors.hpp
PhilipFackler: Resolved in [d21d101](https://github.com/ORNL/GridKit/pull/449/commits/d21d101b46ac911ef7c28472a739aae16e8ff2d7)

PR #421, opened 2026-05-29
REECB Documentation + Implementation

lukelowry:
## Description

Implementation and documentation for the `REECB` PhasorDynamics converter model.

## Proposed changes

- Added `REECB` model implementation, documentation, and diagram.
- Added `REECB` JSON parsing and `SystemModel` assembly support.
- Added focused `PhasorDynamicsConverterReecbTest` unit coverage.
- Added minimal CommonMath compatibility needed for REECB dependency tracking/Jacobian paths.
- Updated converter registration, input-format docs, and changelog.

## Checklist

- [x] All tests pass (except known PE failure)
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [x] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR.

## Further comments

Intended to be merged after #436

PR #415, opened 2026-05-26
REGCA Implementation

lukelowry:
## Description

Adds the REGCA renewable generator/converter model to PhasorDynamics, including system-model JSON parsing, signal wiring for optional current-command inputs, monitor support, and Enzyme/dependency-tracking Jacobian coverage.

## Proposed changes

- Add `Converter/REGCA` model, data type, CMake
- Wire `Regca` into `ComponentLibrary`, `SystemModelData`, `SystemModel`, and JSON parsing.
- Add REGCA unit tests for construction, validation, initialization, residuals, bus injection, JSON/system assembly, and Jac consistency

## Checklist

- [x] All tests pass.
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [x] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR.

## Further comments

Won't add this to any `examples` until I have `REECA` since they go together in all cases we have in PowerWorld

PR #452, opened 2026-06-16
Global Views in PhasorDynamics SystemModel using new Vector

lukelowry:
## Description

This PR updates PhasorDynamics so buses and components use their slices of the `SystemModel` state vectors directly. This removes the local state copy path and keeps component accessors consistent with the integrator-updated system state.

Partially addresses #96. Related to #145.

## Proposed changes

- Added `StateVector` support so `GridElement` can own standalone state or alias `SystemModel` state.
- Updated `SystemModel` allocation to bind each bus/component to global state and verify element sizes remain fixed.
- Updated Enzyme call sites to use the explicit `yData()` accessor for live state pointers.
- Added a PhasorDynamics system test for element/system state aliasing.

| Case | `develop` (s) | This branch (s) | Improvement |
| --- | ---: | ---: | ---: |
| Hawaii | 0.201302 | 0.165187 | 17.9% |
| NewEngland | 0.272596 | 0.220406 | 19.1% |
| Illinois | 0.664262 | 0.592922 | 10.7% |
| WECC | 1.203380 | 1.045540 | 13.1% |
| Texas | 9.063860 | 7.515690 | 17.1% |

## Checklist

- [x] All tests pass. (except the known PE failure)
- [x] N/A: no [CHANGELOG.md](/CHANGELOG.md) update is needed for this internal PhasorDynamics state-storage change.

## Further comments

This work was motivated by the addition of IEEEST stabilizers: adding 333 stabilizers to the Texas case dramatically reduced performance, and this change removes avoidable state-copy overhead in PhasorDynamics system evaluation.

NOTE: I had to make a new PR but this is revival of #411


PhilipFackler:
I like the `VectorView` class. I don't think this is the right way to use it. We could use it in the way that, while touching a lot of files outside PhasorDynamics, would only change the API (not the implementation) for those classes. I'm going to explain my thinking even though, given your comments in the last meeting, you've probably had these ideas already. :)

1. The abstract functions defined for `Evaluator` should return one of these.

That is,

```c++
      virtual std::vector<ScalarT>&       y()       = 0;
      virtual const std::vector<ScalarT>& y() const = 0;

```
becomes
```c++
      virtual VectorView<ScalarT>       y()       = 0;  // no `&` needed
      virtual VectorView<const ScalarT> y() const = 0;

```

This means that the `Evaluator` API doesn't care if a subclass is owning or non-owning.

2. `VectorView` would need (non-`explicit`) constructor(s) to convert _from_ `[const] std::vector<T>&`. This would allow for `Evaluator`s _outside_ PhasorDynamics to still return their `std::vector` members from those `Evaluator` functions.

3. In PhasorDynamics, `Component` (and `BusBase`) can use Slaven's `Vector` directly for these variables. `SystemModel` as a `Component` would own its data in the normal use-case, and components would not. (This also makes conceivable the possibility of a `SystemModel` also being non-owning in a hierarchical system.)

4. Then we would just need a conversion from `Vector` to `VectorView`; either another constructor for `VectorView` or a method on `Vector` (`Vector::getView()`)...or both.

pelesh:
I suggest removing `View` class altogether and using `Vector` instead. If additional features are required in `Vector` class, please create an issue.

lukelowry:
@pelesh @PhilipFackler thank you both for your suggestions. I tried my hardest to take the best of both solutions to get the best of both. Please let me know what you think.

lukelowry:
Rebased, but the speedup is not as significant because the scatter of the jacobian is is now much larger from recent changes in develop :(

pelesh (commented):
For some reason I cannot build this branch. When building from scratch, I get following compile error:
<details>
```
[  5%] Built target phasor_dynamics_signal_dependency_tracking
  "GridKit::LinearAlgebra::Vector<double, int>::getData(GridKit::LinearAlgebra::memory::MemorySpace)", referenced from:
      AnalysisManager::Sundials::Kinsol<double, int>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, int>&) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, long>::getData(GridKit::LinearAlgebra::memory::MemorySpace)", referenced from:
      AnalysisManager::Sundials::Kinsol<double, long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, long>&) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, unsigned long>::getData(GridKit::LinearAlgebra::memory::MemorySpace)", referenced from:
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, unsigned long>&) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, int>::getData(GridKit::LinearAlgebra::memory::MemorySpace) const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, int>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::copyVec(GridKit::LinearAlgebra::Vector<double, int> const&, _generic_N_Vector*) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, int>::getSize() const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, int>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::copyVec(GridKit::LinearAlgebra::Vector<double, int> const&, _generic_N_Vector*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, int>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, int>&) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, long>::getData(GridKit::LinearAlgebra::memory::MemorySpace) const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::copyVec(GridKit::LinearAlgebra::Vector<double, long> const&, _generic_N_Vector*) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, long>::getSize() const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::copyVec(GridKit::LinearAlgebra::Vector<double, long> const&, _generic_N_Vector*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, long>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, long>&) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, unsigned long>::getData(GridKit::LinearAlgebra::memory::MemorySpace) const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::copyVec(GridKit::LinearAlgebra::Vector<double, unsigned long> const&, _generic_N_Vector*) in Kinsol.cpp.o
  "GridKit::LinearAlgebra::Vector<double, unsigned long>::getSize() const", referenced from:
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::Residual(_generic_N_Vector*, _generic_N_Vector*, void*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::getDefaultInitialCondition() in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::copyVec(GridKit::LinearAlgebra::Vector<double, unsigned long> const&, _generic_N_Vector*) in Kinsol.cpp.o
      AnalysisManager::Sundials::Kinsol<double, unsigned long>::copyVec(_generic_N_Vector*, GridKit::LinearAlgebra::Vector<double, unsigned long>&) in Kinsol.cpp.o
[  5%] Built target utilities_logger
make[2]: *** [GridKit/Solver/SteadyState/libgridkit_solvers_steady.0.1.0.dylib] Error 1
make[1]: *** [GridKit/Solver/SteadyState/CMakeFiles/solvers_steady.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

</details>

lukelowry:
@pelesh i think it was a linker issue, fixed in recent commit. Could you try again?

pelesh:
> @pelesh i think it was a linker issue, fixed in recent commit. Could you try again?

That fixed it. Thanks!

GridKit/LinearAlgebra/Vector/Vector.hpp
```diff
@@ -53,6 +55,38 @@ namespace GridKit
...
+      ScalarT* data(memory::MemorySpace memspace = memory::HOST)
+      {
+        return getData(memspace);
+      }
+
+      const ScalarT* data(memory::MemorySpace memspace = memory::HOST) const
+      {
+        return getData(memspace);
+      }
+
```
pelesh: I think having an alias here is counterproductive. We either use `getData` or `data` but having both can be confusing.

GridKit/LinearAlgebra/Vector/Vector.hpp
```diff
@@ -53,6 +55,38 @@ namespace GridKit
...
+      {
+        assert(i < size());
+        return data()[i];
+      }
+
+      const ScalarT& operator[](std::size_t i) const
+      {
+        assert(i < size());
+        return data()[i];
+      }
```
pelesh: I don't think it is a good idea to have this operator overload because it will work only with host memory.

pelesh (commented):
This is great work and I suggest we make it top priority for merging.

I suggest we complete all necessary `Vector` class and other infrastructure changes in #477 and have this PR address redundant copies. That way we avoid scope creep here.

lukelowry:
I will rebase as soon as I get a chance, high priority

GridKit/LinearAlgebra/Vector/Vector.hpp
lukelowry: okay, I removed these, but it required a lot of downstream changes.

GridKit/LinearAlgebra/Vector/Vector.hpp
lukelowry: Also removed, causes a lot more churn downstream tho

lukelowry:
@pelesh I am marking this ready for review, this is one of the last big things before I can get the new models ready for final review

lukelowry:
This is the last item blocking the completion of the other PR models, (#415 , #421 , etc). Once merged I will open a validation PRs for the example cases and update/standardize all the cases and model implementation for readability!

tests/UnitTests/PhasorDynamics/runSystemTests.cpp
```diff
@@ -10,6 +10,7 @@ int main()
 
   result += test.constructor();
   result += test.composer();
+  result += test.elementStateAliasesSystemState();
```
nkoukpaizan, line 13:
```suggestion
  result += test.componentStateAliasesSystemState();
```

tests/UnitTests/PhasorDynamics/LoadZIPTests.hpp
```diff
@@ -181,30 +184,37 @@ namespace GridKit
...
-          bus.y()[i].setVariableNumber(i + load.size());
+          bus_y[i].setVariableNumber(i + load.size());
         }
 
         bus.evaluateResidual();
-        load.evaluateResidual();
-        auto residual_y = load.getResidual();
+        load.evaluateResidual(); ///< Computes the residual and the Jacobian values by tracking
+                                 ///< the dependencies
+        auto&                                     residual_y_view = load.getResidual();
```
nkoukpaizan, line 201: I'm not sure I understand the `_view` in the variable name in the context of the current version of the PR.

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
```diff
@@ -343,33 +352,38 @@ namespace GridKit
...
-          exciter.yp()[i].setVariableNumber(i); ///< Exciter independent variables
+          exciter_yp[i].setVariableNumber(i); ///< Exciter independent variables
         }
 
         bus.evaluateResidual();
         exciter.evaluateResidual(); ///< Computes the residual and the Jacobian values by tracking
                                     ///< the dependencies
-        std::vector<DependencyTracking::Variable> residual_yp = exciter.getResidual();
+        auto&                                     residual_yp_view = exciter.getResidual();
+        std::vector<DependencyTracking::Variable> residual_yp(residual_yp_view.getData(), residual_yp_view.getData() + residual_yp_view.getSize());
```
nkoukpaizan, line 386: Probably out of scope for this PR, but we'll need to remove the `std::vector<DependencyTracking::Variable>`. I'll give that some thought.

GridKit/Solver/Dynamic/Ida.hpp
```diff
@@ -235,9 +236,8 @@ namespace AnalysisManager
 
     private:
       // static void copyMat(Model::Evaluator::Mat& J, SlsMat Jida);
-      static void copyVec(const N_Vector x, std::vector<ScalarT>& y);
-      static void copyVec(const std::vector<ScalarT>& x, N_Vector y);
-      static void copyVec(const std::vector<bool>& x, N_Vector y);
```
nkoukpaizan, line 240: If `VectorT` is `GridKit::LinearAlgebra::Vector<ScalarT, IdxT>`, what's the behavior for the `tag` (`bool`)? An implicit conversion seems dangerous to me, but looks like [IDA is expecting a floating point value anyway](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDASetId). cc @Steven-Roberts

GridKit/Model/PowerFlow/SystemModel.hpp
```diff
@@ -683,6 +809,11 @@ namespace GridKit
     }
 
   private:
+    static ScalarT* dataOrNull(VectorT& vector)
```
nkoukpaizan, line 812: This is not as elegant as what is being done in PhasorDynamics to explicitly check the size. 

Trying to access the `nullptr` is undefined behavior. We'd have to check if the return value is `nullptr` before using.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -16,6 +17,49 @@
...
+  /**
+   * Writes a vector to a file in Matrix Market format
+   *
+   * @param vec The vector to write
+   * @param filename The name of the output file
+   * @param header Additional header information/comments
+   * @return true if the write was successful, false otherwise
+   */
+  template <typename VectorT>
+  void writeVectorToMatrixMarket(const VectorT& vec, const std::string& filename, const std::string& header)
```
nkoukpaizan, line 29: I thought I had deleted this in #447. Rebasing issue perhaps?

I think this belongs either in the Vector class or a separate utility. Tracking the task of adding it back in #467.

GridKit/Model/PhasorDynamics/SystemModel.hpp
```diff
@@ -92,7 +95,11 @@ namespace GridKit
       int evaluateResidual() override;
       int evaluateJacobian() override;
 
-      void updateVariables();
+      CsrMatrixT* getCsrJacobian() const override
```
nkoukpaizan, line 98: I believe I moved this to the `Component` base class and it should not need an override here.

examples/Enzyme/PowerElectronics/main.cpp
```diff
@@ -189,8 +189,8 @@ int main()
 
   for (size_t i = 0; i < dg->getExternSize(); i++)
   {
-    y[i]   = dg->y()[i];
-    res[i] = dg->getResidual()[i];
+    y[i]   = dg->y().getData()[i];
```
nkoukpaizan, line 192: The pattern `auto* y_data = dg->y().getData(); y_data[i]...` (used elsewhere) looks cleaner to me. `.getData()[1]` looks cleaner if we are only accessing a specific index.

GridKit/Model/PhasorDynamics/SystemModelImpl.hpp
```diff
@@ -436,45 +436,62 @@ namespace GridKit
...
+      for (const auto& component : components_)
+      {
+        component->bind(y_, yp_, f_, tag_, abs_tol_, child_offset);
+        child_offset += component->size();
+      }
+
+      assert(child_offset == size_);
 
-      // Default variable and residual index mapping to local index
+      variable_indices_.resize(size);
```
nkoukpaizan, line 476: It may be a good idea to make these `GridKit::LinearAlgebra::Vector<IdxT, IdxT>` right away and allocate them with the others.

GridKit/Model/PhasorDynamics/Component.hpp
```diff
@@ -305,52 +294,52 @@ namespace GridKit
         throw NotImplementedError(__func__);
       }
 
-      [[noreturn]] std::vector<ScalarT>& yB() override
+      [[noreturn]] VectorT& yB() override
```
nkoukpaizan, line 297: The member variables should also be updated, though they are not used. Please add a TODO around those member variables, so we know to pay attention to them later.

nkoukpaizan (commented):
Mostly looks good to me. A few relatively minor comments.

pelesh (commented):
This PR addresses a long overdue work and optimizes model evaluation code.

A couple of issues I suggest we address before merging:
- There is a significant scope creep and the changes in the code go beyond what was stated as the goal of the PR. Consider breaking the PR into a couple of more focused PRs:
    - Replace `std::vector` with `GridKit::LinearAlgebra::Vector`.
    - Eliminate copies between local component variables and system variables (this PR stated goal).
    - All other suggested changes changes.
- Note that `GridKit::Model::Evaluator` class is intended to be pure virtual. Vectors `y_`, `yp_`, `f_`, etc. should be implemented in `GridKit::PhasorDynamics::Component` and `GridKit::PhasorDynamics::BusBase` classes. Some repetition does not hurt right now as these two base classes will further evolve.

GridKit/Solver/Dynamic/Ida.hpp
pelesh: Yes, IDA is expecting floating point values because `N_Vector` is a vector of floating point types. That is why there was an overload `copyVec(const std::vector<bool>& x, N_Vector y)`.

lukelowry:
Thank you all for the reviews I am going to be splitting this into a few smaller PRs with the suggested edits

PR #457, opened 2026-06-19
`ConstantSignalSource` component and example with `BusToSignalAdapter`

PhilipFackler:
## Description
 
Introduce new component model as a simple example of sending currents to buses via the bus signal adapter
 
 ## Proposed changes
 
`ConstantSignalSource` takes complex parameter and makes it available on output signal nodes (real and imag).
 
 ## Checklist
 
- [X] All tests pass.
- [p] There are unit tests for the new code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 - Added off-nominal tap ratio and phase shift support to the PhasorDynamics `Branch` model.
 - Added portable Vector class to GridKit
 - Added support for running IDA with fixed time steps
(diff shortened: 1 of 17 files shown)


pelesh:
I suggest we document this better and mark it ready for review. @lukelowry might have additional suggestions how to make the usage example more relevant for domain experts.

pelesh (approved):
There are issues with how signals are handled in GridKit that are inherited in this PR but are out of scope here (see #407).

For now please rebase and we can merge this PR.

PR #410, opened 2026-05-20
Fix (most) template formatting issues in phasor dynamics

PhilipFackler:
## Description
 
 See #366 
 
## Proposed changes
 
Use `snake_case` for template parameters. and `UpperCamelT` for public interface types.
 
 ## Checklist
 
- [X] All tests pass.
- [X] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [N/A] There are unit tests for the new code.
- [N/A] The new code is documented.
- [X] The feature branch is rebased with respect to the target branch.
- [N/A?] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.

PR #460, opened 2026-06-22
IEEEST Stabilizer Improvements/Corrections

lukelowry:
## Description

Fixes and cleans up the `IEEEST` stabilizer implementation and documentation.

This updates the older initial implementation style so that reduced-order notch filters and zero denominator time constants are handled consistently

## Proposed changes

- Replace the branchy/safe-inverse notch logic with a fixed DAE form using differential tags.
- Initialize `IEEEST` from the attached input signal instead of forcing all states to zero.
- Document the implemented equations, initialization, derived parameters, and unsupported compatibility parameters.
- Tests for initialization, residuals, invalid wiring, invalid notch parameters, zero time constants, and Jacobian consistency.

## Checklist

- [x] All tests pass for the focused PhasorDynamics coverage.

Changelog changes N/A

## Further comments

This is cleanup/correction work for the existing `IEEEST` model, not a new model


lukelowry:
@pelesh this fix is also ready to be merged I forgot to make a PR for this one.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -204,7 +191,7 @@ namespace GridKit
           ret += 1;
         }
 
-        if (a4_ == 0 && a3_ == 0 && a2_ == 0 && a1_ != 0)
+        if (a2_ == ZERO<RealT> && a3_ == ZERO<RealT> && a4_ == ZERO<RealT> && a1_ != ZERO<RealT>)
         {
```
pelesh: Please use `isEqual` function to compare floating point numbers. Direct comparisons are unsafe.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -294,19 +315,17 @@ namespace GridKit
...
         f[5]  = -T4_ * x6_dot - x6 + v5;
         f[6]  = -T6_ * x7_dot - x7 + v6;
-        f[7]  = -v4 + bypass_notch_ * u + use_notch_ * (x1 + A5_ * x2 + (use_4th_order_ + use_3rd_order_) * A6_ * x3);
-        f[8]  = use_T2_block_ * (-T2_ * (v5 - x5) + T1_ * (v4 - x5)) + bypass_T2_block_ * (v4 - v5);
-        f[9]  = use_T4_block_ * (-T4_ * (v6 - x6) + T3_ * (v5 - x6)) + bypass_T4_block_ * (v5 - v6);
-        f[10] = use_T6_block_ * (-T6_ * v7 + Ks_ * T5_ * (v6 - x7)) + bypass_T6_block_ * (Ks_ * v6 - v7);
+        f[7]  = -v4 + x1 + A5_ * x2 + A6_ * x3;
+        f[8]  = tag_[4] * (-T2_ * (v5 - x5) + T1_ * (v4 - x5)) + (1 - tag_[4]) * (v4 - v5);
+        f[9]  = tag_[5] * (-T4_ * (v6 - x6) + T3_ * (v5 - x6)) + (1 - tag_[5]) * (v5 - v6);
+        f[10] = tag_[6] * (-T6_ * v7 + Ks_ * T5_ * (v6 - x7)) + (1 - tag_[6]) * (Ks_ * v6 - v7);
```
pelesh: I am not sure if this is a good idea. We are changing equations based on whether we defined other equations to be differential or algebraic. Furthermore, we rely here on implicit bool-to-integer and integer-to-floating point conversions. This logic needs to be implemented in a different way.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -222,19 +209,53 @@ namespace GridKit
...
-        tag_[4]  = (T2_ != 0.0);
-        tag_[5]  = (T4_ != 0.0);
-        tag_[6]  = (T6_ != 0.0);
+        tag_[0]  = (a2_ != ZERO<RealT> || a3_ != ZERO<RealT> || a4_ != ZERO<RealT>);
+        tag_[1]  = tag_[0];
+        tag_[2]  = (a3_ != ZERO<RealT> || a4_ != ZERO<RealT>);
+        tag_[3]  = (a4_ != ZERO<RealT>);
+        tag_[4]  = (T2_ != ZERO<RealT>);
+        tag_[5]  = (T4_ != ZERO<RealT>);
+        tag_[6]  = (T6_ != ZERO<RealT>);
```
pelesh: The logic implemented this way will break when we add solvers that require DAEs implemented in Hessenberg form.

pelesh (changes requested):
The logic governing changing differential into algebraic equations and vice versa does not seem to be sustainable. This requires more offline discussion.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: I am not sure how to implement this without introducing template-y solutions. Do we want to move toward something like this where we define the model conditionally?

```c++
template <class ScalarT, typename IdxT, Order order, Block t2, Block t4, Block t6>
```

But there are 31 different permutations, so there would be a lot of duplicate code. I don't know what the general approach should be, so I am open to any suggestions.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Fixed!

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Opened an issue #463 where hopefully we can discuss the general approach to take since this impacts almost every model

lukelowry:
I have rebased and adjusted the formulation to align with #463 .

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -156,19 +151,20 @@ namespace GridKit
...
       {
+        size_     = static_cast<IdxT>(IeeestInternalVariables::MAXIMUM);
         auto size = static_cast<size_t>(size_);
-        f_.resize(size);
-        y_.resize(size);
-        yp_.resize(size);
-        tag_.resize(size);
-        abs_tol_.resize(size);
+
+        f_.assign(size, ScalarT{0});
```
nkoukpaizan: `resize` makes more sense to me in the `allocate` method. Initialization occurs later.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -216,30 +215,65 @@ namespace GridKit
...
-        tag_[9]  = false;
-        tag_[10] = false;
-        tag_[11] = false;
+        auto index = [](IeeestInternalVariables variable)
+        {
+          return static_cast<size_t>(variable);
+        };
+
+        std::fill(tag_.begin(), tag_.end(), false);
+        tag_[index(IeeestInternalVariables::X1)] = true;
```
nkoukpaizan: This is a lot less readable to me than the previous implementation.

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -22,312 +24,505 @@ namespace GridKit
...
+          model.yp()[5] = 0.06;
+          model.yp()[6] = 0.07;
+
+          model.evaluateResidual();
+
+          for (size_t i = 0; i < test.expected.size(); ++i)
           {
-            std::cout << "Non-zero residual at index " << i << ": " << f[i] << "\n";
-            success = false;
+            auto test_tol = (i == 11) ? loose_tol : tol_;
```
nkoukpaizan: A comment would be helpful when changing tolerances.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -271,54 +305,89 @@ namespace GridKit
...
 
         return 0;
       }
 
       template <typename scalar_type, typename index_type>
       int Ieeest<scalar_type, index_type>::evaluateResidual()
       {
+        const auto U = static_cast<size_t>(IeeestExternalVariables::U);
+
+        std::fill(ws_.begin(), ws_.end(), ZERO<RealT>);
```
nkoukpaizan: I would argue `std::fill` is a overkill here. The size is 1.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -271,54 +305,89 @@ namespace GridKit
...
+        const auto X7  = static_cast<size_t>(IeeestInternalVariables::X7);
+        const auto V4  = static_cast<size_t>(IeeestInternalVariables::V4);
+        const auto V5  = static_cast<size_t>(IeeestInternalVariables::V5);
+        const auto V6  = static_cast<size_t>(IeeestInternalVariables::V6);
+        const auto V7  = static_cast<size_t>(IeeestInternalVariables::V7);
+        const auto VSS = static_cast<size_t>(IeeestInternalVariables::VSS);
+        const auto U   = static_cast<size_t>(IeeestExternalVariables::U);
+
+        const ScalarT x1  = y[X1];
+        const ScalarT x2  = y[X2];
```
nkoukpaizan, line 282: Same here, the indirection from the enum is less readable to me. We should discuss a sustainable way to use the enums.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -6,8 +6,9 @@
  * @brief Definition of the IEEEST Power System Stabilizer.
  */
 
-#include <cmath>
+#include <algorithm>
 #include <iostream>
+#include <variant>
```
nkoukpaizan, line 11:
```suggestion
```

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -44,106 +47,98 @@ namespace GridKit
...
+        T4_inv_ = ONE<RealT> / T4_;
+        T6_inv_ = ONE<RealT> / T6_;
+
+        a1_ = A1_ + A3_;
+        a2_ = A2_ + A4_ + A1_ * A3_;
+        a3_ = A1_ * A4_ + A2_ * A3_;
+        a4_ = A2_ * A4_;
+
+        order_ = 0;
+        if (a1_ != ZERO<RealT>)
```
nkoukpaizan: Usual comment about comparing to 0.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -44,106 +47,98 @@ namespace GridKit
...
         {
-          Lsmax_ = std::get<RealT>(data.parameters.at(ModelDataT::Parameters::Lsmax));
+          order_ = 1;
```
nkoukpaizan: `order_` being an integer, it will be more useful for checks. If there's a way to make the `order_` the main input and derive others from it, this would be a lot cleaner.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -179,16 +175,19 @@ namespace GridKit
         if (signals_.template isAssigned<IeeestInternalVariables::VSS>())
         {
           signals_.template getSignalNode<IeeestInternalVariables::VSS>()->set(
-              &y_[11], &(this->getVariableIndex(11)));
+              &y_[static_cast<size_t>(IeeestInternalVariables::VSS)],
+              &(this->getVariableIndex(static_cast<IdxT>(IeeestInternalVariables::VSS))));
         }
 
+        tagDifferentiable();
```
nkoukpaizan: `tagDifferentiable` calls from outside the model should be sufficient. If this is a necessary workaround, please add a comment.
```suggestion
```

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -336,8 +405,13 @@ namespace GridKit
       void Ieeest<scalar_type, index_type>::initializeMonitor()
       {
         using Variable = typename ModelDataT::MonitorableVariables;
-        monitor_->set(Variable::vss, [this]
-                      { return y_[11]; });
+        auto index     = [](IeeestInternalVariables variable)
```
nkoukpaizan: I've seen this pattern a few times. Consider defining once if it is really needed. I would argue it's less readable.

tests/UnitTests/PhasorDynamics/runStabilizerIeeestTests.cpp
```diff
@@ -6,9 +6,10 @@ int main()
 
   GridKit::Testing::StabilizerIeeestTests<double, size_t> test;
 
-  result += test.constructor();
-  result += test.zeroInitialResidual();
```
nkoukpaizan, line 10: Why remove the `constructor` and the `zeroInitialResidual` tests?

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -6,6 +6,10 @@
 
 #pragma once
 
+#include <cstddef>
+#include <memory>
+#include <vector>
```
nkoukpaizan, line 11: Pretty sure `vector` should already be included. Could you comment on the need for `cstddef` and `memory`?

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -26,7 +27,8 @@ namespace GridKit
       template <typename scalar_type, typename index_type>
       Ieeest<scalar_type, index_type>::Ieeest()
       {
-        size_ = 12;
+        size_ = static_cast<IdxT>(IeeestInternalVariables::MAXIMUM);
```
nkoukpaizan, line 30: I prefer the hard-coded size, but that's minor. Perhaps add an assert for the enum.

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -22,312 +24,505 @@ namespace GridKit
...
+          const std::vector<ScalarT>  expected;
+        };
 
-        // Wire: stabilizer reads u_node as input, writes vss_node as output
-        stab.getSignals().template attachSignalNode<PhasorDynamics::Stabilizer::IeeestExternalVariables::U>(&u_node);
-        stab.getSignals().template assignSignalNode<PhasorDynamics::Stabilizer::IeeestInternalVariables::VSS>(&vss_node);
+        const std::vector<ResidualCase> cases = {
+            {"baseline",
+             [](DataT&) {},
+             {0.19, 0.28, 0.37, 1.0975, 0.25, 0.24, -0.01, -0.42, -0.25, -0.31, 1.15, 0.0}},
```
nkoukpaizan: This combination of `std::vector` and `lambda` functions is not very readable to me. Consider splitting.

nkoukpaizan (commented):
I have a few comments here. Perhaps worth discussing at our next meeting.

lukelowry:
Testing Ground for template parameters `<Order N>`

lukelowry:
general implementation of the template parameter done

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Resolved in [1ddb243](https://github.com/ORNL/GridKit/pull/460/commits/1ddb24306fb73877b0b5aed86cbcf308f3f43ec4), thank you!

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Resolved in [1ddb243](https://github.com/ORNL/GridKit/pull/460/commits/1ddb24306fb73877b0b5aed86cbcf308f3f43ec4), thank you for  notcing that

lukelowry:
@nkoukpaizan what are your thoughts on my template implementation?

nkoukpaizan:
> @nkoukpaizan what are your thoughts on my template implementation?

The model implementation is cleaner with the `order` as a template parameter. However, the order remains a runtime selection based on floating point comparisons, which is what I wanted to avoid.

lukelowry:
@nkoukpaizan we could possibly require the order to be passed as a parameter of IEEEST model, would that resolve your concern? 

"Order" is not usually directly exposed as a model parameter and so would require some knowledge of how IEEEST works to pass valid parameters. 

Can you give an example where the runtime comparison is bad/fails? I might be able to resolve both our concerns if I understand better.

nkoukpaizan:
> @nkoukpaizan we could possibly require the order to be passed as a parameter of IEEEST model, would that resolve your concern?
> 
> "Order" is not usually directly exposed as a model parameter and so would require some knowledge of how IEEEST works to pass valid parameters.
> 
> Can you give an example where the runtime comparison is bad/fails? I might be able to resolve both our concerns if I understand better.

Yes, exposing the order as a parameter to the model is what I had in mind. Understood that that's not the domain practice. That's why as was asking whether this is possible as a question.

Floating point values are stored with limited precision. We'd have to guarantee that the value returned by the parser is bit-for-bit equal to zero. That may be the case for [nlohmann-json](https://json.nlohmann.me/features/types/number_handling/#zeros), but considering #227 and the possibility of using a different parser, that's not guaranteed.

lukelowry:
@nkoukpaizan i see the issue now thank you, that was very clarifying. After reflecting on my past experience and discussing with others at TAMU, I think deriving the order from the parameters alone is not intuitive or safe, either.

As discussed, we should add an integer parameter to the IEEEST model, "order" and use that for the StabilizerFactory (rename to IeeestFactory). Then, our parameter validation can gate for zero division relevant to each static order (e.g. $A_2 + A_3>0$). Additionally the model json may then contain some coefficients that are only used at a certain "order", which would be a feature IMO since you can retain information needed for higher order but choose to operate at smaller order.

I will make this change then we can reasess.

lukelowry:
@nkoukpaizan I have made the proposed edits. For future models with many different "orders" for subsystems the template arguments might get long, but it works well for one argument here

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -30,311 +42,730 @@ namespace GridKit
...
+      }
 
-        u_value = 0.5;
-        setStatePointDep(stab);
+      /// Symmetric notch filter (A1 = A3 = 0): a standard fourth-order
+      /// configuration with zero interior denominator coefficients
+      /// (a1 = a3 = 0)
+      ///
+      /// The Enzyme-vs-dependency-tracking Jacobian comparison is not run for
+      /// this configuration: Enzyme's `sparse_store` drops exact-zero entries
```
nkoukpaizan: Enzyme should not drop non-structural zeros. May be worth looking into it if the Jacobians differ. I recommend starting with an implementation of the test with an `expectFailure` outcome.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -23,233 +23,263 @@ namespace GridKit
...
-      int Ieeest<scalar_type, index_type>::verify() const
+      template <typename scalar_type, typename index_type, size_t order>
+      int Ieeest<scalar_type, index_type, order>::validateLeadingCoefficient(
+          RealT       value,
+          const char* name)
       {
-        int ret = 0;
-
-        if (signals_.template isAttached<IeeestExternalVariables::U>())
+        if (value != ZERO<RealT>)
```
nkoukpaizan: I think `isEqual` would be appropriate here. I recommend a PR that moves it from the `Testing` namespace to, e.g., `Math`. @lukelowry Let me know if you can do that or if you'd rather me to.

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -30,311 +42,730 @@ namespace GridKit
...
+        const std::string name = orderedName(__func__, order);
+        return success.report(name.c_str());
       }
 
-#ifdef GRIDKIT_ENABLE_ENZYME
-      /**
-       * @brief Compare DependencyTracking Jacobian against Enzyme Jacobian.
-       */
-      TestOutcome jacobian()
+      TestOutcome verify()
```
nkoukpaizan: Verify currently logs some errors. Make sure to message the user about expected errors, similar to #500.

tests/UnitTests/PhasorDynamics/runStabilizerIeeestTests.cpp
```diff
@@ -7,10 +7,42 @@ int main()
   GridKit::Testing::StabilizerIeeestTests<double, size_t> test;
 
   result += test.constructor();
-  result += test.zeroInitialResidual();
-  result += test.residual();
+
+  result += test.init<0>();
```
nkoukpaizan: It might be cleaner to template the test class rather than individual member functions. 

If I understand correctly, some tests are manipulating stabilizer objects of different orders. Consider two test objects: 
- One templated to check all orders
- Another one that manipulates/compares objects of different orders.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestEnzyme.cpp
```diff
@@ -19,10 +19,11 @@ namespace GridKit
        *
        * @tparam ScalarT - Scalar data type
        * @tparam IdxT    - Index data type
```
nkoukpaizan: Note the changes in templating style in #410 and #505.  Something to be careful with when rebasing. We'll have to make sure new instances of the old format are not introduced in new PRs.

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -30,311 +42,730 @@ namespace GridKit
...
+        TestStatus success = true;
 
-        auto        tol    = 10 * std::numeric_limits<RealT>::epsilon();
-        const auto& f      = stab.getResidual();
-        const auto* f_data = f.getData();
-        for (size_t i = 0; i < f.getSize(); ++i)
+        using Params = PhasorDynamics::Stabilizer::IeeestParameters;
+
+        // The second case exercises the TIME_CONSTANT_MINIMUM flooring.
+        const std::vector<std::array<RealT, 3>> time_constant_cases = {
```
nkoukpaizan: I would split into two different tests.

nkoukpaizan (commented):
The template for the stabilizer object looks a lot cleaner to me!

I'm having a hard time reading the `StabilizerIeeestTests` at the moment, especially where multiple cases are defined as vectors of arrays. I would split the cases into separate tests and implement specific, reusable private methods.

lukelowry:
Marking as draft until after release because this is rather big change to introduce

lukelowry:
Since IEEEST is already not in Hessenberg form, I am temporarily closing this in favor of #520, which is a significantly cleaner implicit representation of the model. The 'minimum time constant' approach is unfortunately making validation quite difficult. Tracked in #463 

I am justifying this decision because IEEEST can simply be removed from a case if it becomes a blocker for Hessenberg, but right now validation is more urgent. We can reopen after v0.2 is released

lukelowry:
Rebasing ASAP

pelesh:
I suggest to proceed like this:
- Define model order as an integer parameter.
- Use switch statement in initialization and residual functions to select correct equations with the order parameter as the switch.

I would avoid templates for now. I don't think we should use hacks as in #520.

lukelowry:
@pelesh please see most recent version of this branch, where I use a switch. How does this look?

lukelowry:
Confirmed working in validation of three cases against PowerWorld. See #461

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -191,25 +92,69 @@ namespace GridKit
...
+
+        check(std::isfinite(a1_) && std::isfinite(a2_)
+                  && std::isfinite(a3_) && std::isfinite(a4_),
+              "expanded denominator coefficients must be finite");
+
+        const IdxT numerator_order = A6_ != ZERO<RealT>
+                                         ? static_cast<IdxT>(2)
+                                     : A5_ != ZERO<RealT>
+                                         ? static_cast<IdxT>(1)
+                                         : static_cast<IdxT>(0);
```
pelesh: You can assume that a floating point number is never exactly zero. The comparisons above are likely to return true most of the time.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -237,18 +226,9 @@ namespace GridKit
...
-        tag_[5]  = (T4_ != 0.0);
-        tag_[6]  = (T6_ != 0.0);
-        tag_[7]  = false;
-        tag_[8]  = false;
-        tag_[9]  = false;
-        tag_[10] = false;
-        tag_[11] = false;
+        std::fill(tag_.begin(), tag_.end(), false);
+        constexpr auto X7 = static_cast<size_t>(IeeestInternalVariables::X7);
+        std::fill_n(tag_.begin(), X7 + 1, true);
```
pelesh, line 235: You need a switch statement here to define differential variables for each order.

pelesh: Also, for clarity, I would avoid using `std::fill`. Just spell them out.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -191,25 +92,69 @@ namespace GridKit
...
+          break;
+        case 4:
+          leading_coefficient = a4_;
+          leading_name        = "a4";
+          break;
+        }
 
-        if (a4_ == 0 && a3_ == 0 && a2_ == 0 && a1_ != 0)
+        if (leading_name != nullptr && leading_coefficient == ZERO<RealT>)
         {
```
pelesh, line 159: Operator `==` or `!=` should not be used to compare floating point numbers.

pelesh (commented):
Looks much better!!

A few things to fix quickly before more thorough review.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -113,13 +105,19 @@ namespace GridKit
...
-            const ScalarT*,
-            const ScalarT*,
-            const ScalarT*,
-            const ScalarT*,
-            ScalarT*);
+            const ScalarT* y,
+            const ScalarT* yp,
+            const ScalarT* wb,
+            const ScalarT* ws,
+            ScalarT*       f);
```
nkoukpaizan: Variable names aren't strictly necessary here. Is there a motivation for the change other than style?

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
nkoukpaizan: I find syntactic sugar for nested conditions hard to read. Consider expanding the conditionals.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -280,60 +260,123 @@ namespace GridKit
...
+        const ScalarT x4_dot = yp[X4];
+        const ScalarT x5_dot = yp[X5];
+        const ScalarT x6_dot = yp[X6];
+        const ScalarT x7_dot = yp[X7];
+
+        const ScalarT u = ws[U];
+
+        switch (order_)
+        {
+        case 0:
```
nkoukpaizan, line 306: I'm not sure having the switch inside the residual evaluation is better than other alternatives we've explored. I suspect that is the motivation for the new `-mllvm -simplifycfg-sink-common=false` flag.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/README.md
```diff
@@ -1,114 +1,251 @@
...
+      x_1 + \dfrac{A_5}{a_1}(-x_1+u) & n=1 \\
+      x_1 + A_5x_2
+        + \dfrac{A_6}{a_2}(-x_1-a_1x_2+u) & n=2 \\
+      x_1 + A_5x_2 + A_6x_3 & n\in\{3,4\}
+    \end{cases} \\
+  0 &= -v_5 + x_5 + \dfrac{T_1}{T_2}(v_4-x_5) \\
+  0 &= -v_6 + x_6 + \dfrac{T_3}{T_4}(v_5-x_6) \\
+  0 &= -v_7 + K_s\dfrac{T_5}{T_6}(v_6-x_7) \\
+  0 &= -V_{\mathrm{ss}}
+    + \operatorname{clamp}(v_7,L_s^{\min},L_s^{\max}).
```
nkoukpaizan: Rendering issue on Github.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/README.md
```diff
@@ -1,114 +1,251 @@
...
-$u$.
+### Input Initialization
+
+```math
+u \leftarrow \text{stabilizer input signal}.
+```
+
+### Internal Initialization
+
+```math
```
nkoukpaizan, line 226: Rendering issue on Github.

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -1,433 +1,918 @@
...
 
         return success.report(__func__);
       }
 
-      /**
-       * @brief All states initialize to zero (stabilizer at rest).
-       * With u = 0, all residuals should be zero.
-       */
-      TestOutcome zeroInitialResidual()
+      TestOutcome initializationAndSignals()
```
nkoukpaizan, line 238: This generates some errors without notifying that those are expected.
```
55: [ERROR] Ieeest: initial input signal U must be finite
55: [ERROR] Ieeest: initial input signal U must be finite
55: [ERROR] Ieeest: initial input signal U must be finite
55: [ERROR] Ieeest: required input signal U is not attached
55: [ERROR] Ieeest: cannot initialize with invalid configuration
55: --- PASS: Test initializationAndSignals
```

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
```diff
@@ -1,433 +1,918 @@
...
+          success         *= (dependency_jacobian.size() == enzyme_jacobian.size());
+          const auto rows  = std::min(dependency_jacobian.size(), enzyme_jacobian.size());
+          for (size_t row = 0; row < rows; ++row)
+          {
+            if (!isEqual(dependency_jacobian[row], enzyme_jacobian[row], kTol))
+            {
+              std::cout << "IEEEST Jacobian row " << row << " for order " << order
+                        << " differs between dependency tracking and Enzyme\n";
+              printDependencyMap(dependency_jacobian[row], "dependency tracking");
+              printDependencyMap(enzyme_jacobian[row], "Enzyme");
```
nkoukpaizan: Since there are only two calls to the `printDependencyMap` helper, I would spell it out and use `DependencyTracking::Variable::print()`.

nkoukpaizan (commented):
A few comments on the current state of the PR.

I'll defer to @pelesh if the switch really is what we want to go with.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
lukelowry: Addressed in [fa453fa](https://github.com/ORNL/GridKit/pull/460/commits/fa453fac103275fee099c45ca761b68be93cde23), I removed them. thank you!

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Agree. I am used to using ternary operators but its ugly. Remvoed in 2 places in [8909d47](https://github.com/ORNL/GridKit/pull/460/commits/8909d47c2243f96e40263299fc93927ebafbb8d1)

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/README.md
lukelowry: Thank you [2809549](https://github.com/ORNL/GridKit/pull/460/commits/28095495a59df6b2638cebd8ae5f9231fa20bc27). Perpetrator was `\operatorname` once again

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/README.md
lukelowry: Addressed in [2809549](https://github.com/ORNL/GridKit/pull/460/commits/28095495a59df6b2638cebd8ae5f9231fa20bc27)

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
lukelowry: Thank you good catch, addressed here [bd1fcad](https://github.com/ORNL/GridKit/pull/460/commits/bd1fcadf29cd2345bd210e7a5fed28a71f8a2790)

tests/UnitTests/PhasorDynamics/StabilizerIeeestTests.hpp
lukelowry: Removed in [292c07c](https://github.com/ORNL/GridKit/pull/460/commits/292c07c895a422fd794519baeed29525bb453b9b)

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Yes I had to add that flag for Enzyme to construct correctly. I will wait for @pelesh to chime in

lukelowry:
@nkoukpaizan thank you for the timely review. pending @pelesh and @abirchfield now

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Tracking in #530 . I attached an IEEE specification there that might better communicate my choices here

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Addressed, thank you!

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: Tracking in https://github.com/ORNL/GridKit/issues/530 .

abirchfield (approved):

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
pelesh: No, this is not going to work as implemented (I rushed to celebration too early). You need a separate residual for each order and separate Jacobian for each residual. Jacobian derivatives are evaluated at compile time, so they have to have residual equations available at compile time.

The switch statement can be used to select residual function of the right order and its matching Jacobian but they need to be implemented separately. The rest of model infrastructure could be reused, I think.

pelesh (changes requested):
The current implementation does not build on my machine. Enzyme seems to be trying to get derivatives for all runtime combinations and eventually runs out of memory.

The issue we are dealing here is not the modeling of a device but rather trying to fit five different mathematical model in one component. The documentation in README is misleading in that regard.

I suggest implement stabilizer for only one order as a temporary solution.

lukelowry:
I would rather do the other branch we had for IEEEST #520 than do a single-order implementation as a temporary solution then, because that is much cleaner its just not in Hessenberg form. I can't validate the cases in #461 without variable order IEEEST

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -191,25 +92,73 @@ namespace GridKit
...
+          break;
+        case 3:
+          leading_coefficient = a3_;
+          leading_name        = "a3";
+          break;
+        case 4:
+          leading_coefficient = a4_;
+          leading_name        = "a4";
+          break;
+        }
```
superwhiskers, line 155: this stringification of the leading coefficient variable name seems kind of suspicious to me. i get that it's only for the error message, but the way this is done kind of seems odd for a reason that i cannot quite articulate.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -191,25 +92,73 @@ namespace GridKit
...
-          if (!signals_.template isLinked<IeeestExternalVariables::U>())
+          if (!condition)
           {
-            Log::error() << "Ieeest: input signal U attached with no linked source\n";
+            Log::error() << "Ieeest: " << message << '\n';
             ret += 1;
           }
-        }
-        else
+        };
```
superwhiskers, line 103: don't we have [`assert`](https://www.cppreference.com/cpp/error/assert) for this?

superwhiskers: or are these checks always necessary (`assert` functions go away when `NDEBUG` is defined)?

as an additional note, this should not increment `ret`. it should just return an error from the function. continuing onward after a failed assertion is a bad idea.

superwhiskers: additionally, if this change is to stay, this should go into a utility header or something like that, not an ad-hoc thing inside of a function. this is better for consistency across the codebase

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -345,12 +392,145 @@ namespace GridKit
...
+        if (load_real(Params::Vcu, Vcu_, "Vcu") && Vcu_ != ZERO<RealT>)
+        {
+          Log::warning() << "Ieeest: nonzero Vcu requests an upper input cutout, "
+                            "but input cutout is not implemented and Vcu is ignored\n";
+        }
+        if (load_real(Params::Tdelay, Tdelay_, "Tdelay") && Tdelay_ != ZERO<RealT>)
+        {
+          Log::warning() << "Ieeest: nonzero Tdelay requests an input delay, "
+                            "but input delay is not implemented and Tdelay is ignored\n";
+        }
```
superwhiskers, line 470: we should come up with a way to convert a parameter enum to a string to avoid hardcoded strings for the purpose of logging

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -345,12 +392,145 @@ namespace GridKit
...
+
+      template <typename scalar_type, typename index_type>
+      void Ieeest<scalar_type, index_type>::setDerivedParameters()
+      {
+        auto check_non_negative = [&](RealT value, const char* name)
+        {
+          if (value < ZERO<RealT>)
+          {
+            Log::error() << "Ieeest: " << name << " must be non-negative\n";
+            ++parameter_error_count_;
```
superwhiskers, line 494: while the signature of the method may be forcing this, i'd prefer we don't continue with this pattern of keeping an error count

superwhiskers (changes requested):
haven't built and tested yet, but there are a few things that stand out as necessary to correct prior to this being merged

lukelowry:
temp drafting before release

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -122,6 +114,12 @@ namespace GridKit
             ScalarT*);
 
       private:
+        void initializeParameters(const ModelDataT& data);
+        void initializeMonitor();
+        void setDerivedParameters();
+
+        static constexpr RealT TIME_CONSTANT_MINIMUM = static_cast<RealT>(1.0e-3);
```
superwhiskers, line 121: it looks like this same constant is duplicated across many header files. is there any reason it cannot be placed in a more "global" place so we aren't copy-pasting it all over?

nkoukpaizan: Consolidating is desirable and possible. But at this point, it's still not clear that we want to maintain this minimum time constant.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -16,6 +21,9 @@ namespace GridKit
 {
   namespace PhasorDynamics
   {
+    template <typename scalar_type, typename index_type>
+    class SignalNode;
+
```
superwhiskers: why are we forward declaring this here?

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -44,41 +52,42 @@ namespace GridKit
...
+        using ModelDataT         = IeeestData<RealT, IdxT>;
+        using SignalNodeSetT     = SignalNodeSet<ScalarT, IdxT>;
+        using SignalPortsT       = SignalPorts<ScalarT, ModelDataT>;
+        using MonitorT           = Model::VariableMonitor<Ieeest, IeeestData>;
+        using InternalVariablesT = IeeestInternalVariables;
+        using ExternalVariablesT = IeeestExternalVariables;
 
         Ieeest();
-        Ieeest(const ModelDataT& data);
+        explicit Ieeest(const ModelDataT& data);
```
superwhiskers: just curious what motivated this adjustment here

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -190,23 +91,73 @@ namespace GridKit
...
-        auto input_port = ports_.in.template port<IeeestSignalInputs::input>();
-        if (!input_port.connected())
+        auto check = [&](bool condition, const char* message)
+        {
+          if (!condition)
+          {
+            Log::error() << "Ieeest: " << message << '\n';
+            ret += 1;
+          }
+        };
```
superwhiskers: may be worthwhile building on top of your #566 pull request. could rebase this upon that branch

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
superwhiskers: i'd say that we should move the order to a template parameter at minimum. this seems like a good way to resolve the issue. you can then implement the conditional logic using `if constexpr` conditions which will evaluate that logic at compile time

superwhiskers (commented):
nothing significant, just some additional comments & suggestions

PR #475, opened 2026-07-01
Support Bus Parameters

lukelowry:
## Description

Updates PhasorDynamics bus input data so buses use the same `params` convention as component models.

This replaces the old bus-level `v_base` field with param `kv`  since we are using 'mva' for power bases. Cleaner for phasor dynamics, and lets us add other parameters we might need, such as `region` or other labeling 

## Proposed changes

- Add `BusParameters::kv` and store bus parameters in the enum parameter map.
- Parse bus `params` with the same style used by component data parsers.
- Update PhasorDynamics examples and case-format tests from `v_base` to `params.kv`.
- Update `INPUT_FORMAT.md` and clean up the Bus README notation.

## Checklist

- [x] All tests pass (except know PE failure)

Changelog changes N/A


pelesh (approved):
I approve but would like @superwhiskers to take a quick look before we merge.

GridKit/Model/PhasorDynamics/Bus/README.md
```diff
@@ -1,25 +1,30 @@
...
-variables -- real and imaginary voltage $`V_{rk}`$ and $`V_{ik}`$,
-respectively. The bus also owns current balance residual equations for real
-and imaginary currents coming into the bus $`I_{rk}`$ and $`I_{ik}`$,
-respectively. While th bus model owns current residuals, it _does not compute_
-them. Instead, each component connected to the bus is adding its contribution
-to the residual. The bus will merely initialize the residual to zero each time
-numerical integrator requests residual evaluation.
+A bus is a point of interconnection of electrical devices. The bus component
+model also plays a key role in coupling system components. Each bus $k$ owns
+two variables, real and imaginary voltage $V_{rk}$ and $V_{ik}$,
```
superwhiskers: if we're taking writing suggestions which could be entirely irrelevant here: this could be phrased better as "Each bus $k$ owns two variables: real voltage and imaginary voltage, denoted as $V_{rk}$ and $V_{ik}$, respectively."

a similar suggestion also applies to a later sentence, where i would rephrase it as "The bus also owns current-balance residual equations for real and imaginary currents entering the bus, denoted as $I_{rk}$ and $I_{ik}$, respectively."

there may be some other, better way of phrasing these, but this is what came to me.

superwhiskers (approved):
looks fine overall, mostly appears to be updates to json models anyway. my only comments are on some of the documentation, which can be ignored.

GridKit/Model/PhasorDynamics/Bus/README.md
lukelowry: Changed in recent commit. Thank you for feedback! I will do a larger round of polishing of wording/grammar before next release, aswell.

lukelowry:
@pelesh Great!

PR #479, opened 2026-07-07
Add Rosenbrock integrators

alexander-novo:
## Description
 
Add Rosenbrock integrators, which are lighter-weight alternatives to IDA that work well with co-simulation.
 
 ## Proposed changes
 
The current version depends on ReSolve and needs porting work to be used with GridKit vectors. The operations used by the Rosenbrock integrators are:

```c++
vector_handler_.scal();
vector_handler_.axpy();
vector_handler_.abs();
vector_handler_.max();
vector_handler_.diagSolve();
vector_handler_.amax();
```

This PR also need's ReSolve's `SystemSolver` to be ported.

As well, I think some more testing wouldn't go amiss... An integration test on an example power grid would work nicely.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments


GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,999 @@
...
+ * wraps a fallible function call and causes the current function to fail as well,
+ * returning the same error code as the called function.
+ *
+ */
+#define BUBBLE_FAIL(arg) \
+  do                     \
+  {                      \
+    if (int err = (arg)) \
+      return err;        \
+  } while (false)
```
pelesh, line 23: I would also suggest using GridKit logger to return informative error message(s) to the users.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,999 @@
...
+   * @brief Constructs a human-readable report string out of the `StepInfo` object.
+   *
+   * Useful for debug dumps.
+   *
+   */
+  template <class ScalarT, typename IdxT>
+  std::string Rosenbrock<ScalarT, IdxT>::StepInfo::report() const
+  {
+    std::stringstream out;
+    out << std::scientific << std::setprecision(20)
```
pelesh: Please use GridKit logger. It allows you to switch output on and off and also lets you choose between standard output and user defined files.

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
```diff
@@ -0,0 +1,205 @@
...
+    };
+
+    re.alpha_sum[0] = 0.0;
+    re.alpha_sum[1] = 0.6358126895828704;
+    re.alpha_sum[2] = 0.4095798393397535;
+    re.alpha_sum[3] = 0.9769306725060716;
+    re.alpha_sum[4] = 0.4288403609558664;
+    re.alpha_sum[5] = 1.0;
+    re.alpha_sum[6] = 1.0;
+    re.alpha_sum[7] = 1.0;
```
pelesh: All of these are double precision. Nice to have feature would be to have quad precision numbers in case we ever enable that in GridKit. This shouldn't be a show-stopper for merging this PR.

pelesh (commented):
This is really nice work! 

I would suggest merging #480 first and then rebasing so that we avoid introducing Re::Solve dependency.

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
alexander-novo: Unfortunately, I don't think we have access to the exact values for these. Let me grab and insert the citation for them - I believe it only has double precision values.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: This is just for generating a string report - nothing is output here. No need for logger.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: I can do this, but `BUBBLE_FAIL` is used when a different function has already failed (and printed its informative error message). All printing a new message will do is essentially be a stack trace. Is that what we want?

alexander-novo:
@pelesh I've updated the PR to make use of GridKit's own `Vector` and `VectorHandler`. Re::Solve is still used for the lienar solver part, however (see Rosenbrock::time_step)

pelesh:
> @pelesh I've updated the PR to make use of GridKit's own `Vector` and `VectorHandler`. Re::Solve is still used for the lienar solver part, however (see Rosenbrock::time_step)

Using a linear solver from Re::Solve is a good thing :). We can add bare KLU interface in a separate PR, if needed.

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    re.C[0] = 2.0;
+
+    return re;
+  }
+
+  /**
+   * @brief
+   *
+   * @todo Add paper citation.
```
shakedregev: Do you want to just add it, or is it not out yet?

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
shakedregev: So these are particular values for a test or something? Or are these quadrature points?

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1010 @@
...
+   * @todo Document mass matrix construction
+   *
+   * @param t0 The starting simulation time.
+   * @return An error code, with 0 as success.
+   */
+  template <class ScalarT, typename IdxT>
+  int Rosenbrock<ScalarT, IdxT>::initializeSimulation(RealT t0)
+  {
+    current_time_ = t0;
+    BUBBLE_FAIL(y_cur_->copyFromExternal(model_->y().data(), memspace_, memspace_));
```
shakedregev: Is it standard for `BUBBLE_FAIL` to be written like this? I'm not sure we have a convention that covers it.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1010 @@
...
+   * information. This derivative information can be re-used for Hermite interpolation.
+   *
+   * @todo It doesn't really make sense to have the error estimator separate from the step controller, since your choice of one will
+   * affect your choice of the other. The error estimator should probably be inside the step controller, then accessed if needed.
+   *
+   * @todo Return an error when max steps is hit.
+   *
+   * @todo Check if current time is close enough to output time, and skip interpolation
+   *
+   * @todo Configure upper bound for skip lu step size increase
```
shakedregev: Are these for the next PR?

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1010 @@
...
+        if (step_controller.usesError())
+        {
+          if (err_norm_ == nullptr)
+          {
+            std::cerr << "The provided step controller requires the use of an error norm, but none was provided!\n";
+
+            return -1;
+          }
+
+          State& err_vec = error_estimate();
```
shakedregev: function call seems to violate naming scheme

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1010 @@
...
+                .skip_lu        = skip_lu_,
+                .skip_f         = skip_f_,
+                .accepted       = prev_accept,
+            });
+          }
+          stats_.min_step = std::min(stats_.min_step, step_size_);
+          stats_.max_step = std::max(stats_.max_step, step_size_);
+
+          std::swap(y_prev_, y_cur_);
+          std::swap(y_cur_, y_new_);
```
shakedregev: I figured out what's happening here. But I don't think it's obvious.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1010 @@
...
+        // Generate output at the appropriate time.
+        if (tab_.hasDenseOutput())
+        {
+          if (!dense_coefficients_valid_)
+          {
+            BUBBLE_FAIL(calc_dense_coeff());
+            dense_coefficients_valid_ = true;
+          }
+
+          BUBBLE_FAIL(interp_dense(theta));
```
shakedregev: function violates convention

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
alexander-novo: Just added.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: yes

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: No, I think standard is to ignore fallible functions for the most part and just keep on going. Mostly because these functions are marked as fallible but can never really in practice fail. But this code calls A LOT of fallible functions which can actually fail (ReSolve functions fail for a bunch of different reasons).

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
alexander-novo: @shakedregev These are method coefficients. The `alpha_sum` coefficients here are quadrature nodes (abscissae) but the rest of the coefficients are just general Rosenbrock coefficients that describe different things. You can see the `Rosenbrock::timeStep()` and `Rosenbrock::errorEstimate()` documentations to see how some of them are used.

alexander-novo: @pelesh I checked the source paper again - these coefficients are determined through a numerical minimization procedure, so we would have to redo that procedure if we wanted higher precision coefficients.

tests/UnitTests/Solver/Dynamic/CMakeLists.txt
```diff
@@ -1,9 +1,18 @@
...
 
-add_test(NAME IDATest COMMAND $<TARGET_FILE:test_ida>)
+  add_test(NAME IDATest COMMAND $<TARGET_FILE:test_ida>)
 
-install(TARGETS test_ida RUNTIME DESTINATION bin)
+  install(TARGETS test_ida RUNTIME DESTINATION bin)
+endif()
+
+if(GRIDKIT_ENABLE_RESOLVE)
+  add_executable(test_ros runRosenbrockTests.cpp)
```
nkoukpaizan:
```suggestion
  add_executable(test_rosenbrock runRosenbrockTests.cpp)
```
I recommend the more verbose name, so we have a better understanding when running the tests. Update other spots below as well.

GridKit/Solver/Dynamic/Rosenbrock.cpp
shakedregev: I mean the naming convention (with capital letters and an underscore)

alexander-novo: Oh yeah we don't have a convention for this, but it's fairly standard to have macros be screaming snake case.

alexander-novo: See e.g. GRIDKIT_ENABLE_SUNDIALS_SPARSE https://github.com/ORNL/GridKit/blob/c099ac0037b27c03d4ec9964cd3a006429a636eb/GridKit/Solver/Dynamic/Ida.hpp#L56-L58

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1012 @@
...
+
+    return 0;
+  }
+
+  /**
+   * @brief Advance the simulation forward by one step, storing the new state in \ref y_new_.
+   *
+   * Apply the Rosenbrock scheme using the stored tableau. Each stage \f(u_i\f) is calculated as
+   *
+   * \f[\left(J - \frac{1}{h\gamma} M\right)u_i = -f \left(t_0 + \alpha_ih, y_0 + \sum_{j = 1}^{i - 1} a_{ij}u_j\right) - M \sum_{j=1}^{i-1} \left(\frac{c_{ij}}{h}\right)u_j,\f]
```
shakedregev:
```suggestion
   * \f(\left(J - \frac{1}{h\gamma} M\right)u_i = -f \left(t_0 + \alpha_ih, y_0 + \sum_{j = 1}^{i - 1} a_{ij}u_j\right) - M \sum_{j=1}^{i-1} \left(\frac{c_{ij}}{h}\right)u_j,\f]
```

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1012 @@
...
+      // TODO: non-autonomous model
+      BUBBLE_FAIL(workspace_.asum_->copyToExternal(model_->y().data(), memspace_, memspace_));
+      model_->updateTime(t0 + tab_.alpha_sum_[i] * dt, 0.0);
+      BUBBLE_FAIL(model_->evaluateResidual());
+      workspace_.RHS_->copyFromExternal(model_->getResidual().data(), memspace_, memspace_);
+
+      // TODO: examine if this -1 is correct
+      vector_handler_.scal(-1, workspace_.RHS_.get(), memspace_);
+      vector_handler_.scal(workspace_.mass_.get(), workspace_.csum_.get(), memspace_);
+      vector_handler_.axpy(-1, workspace_.csum_.get(), workspace_.RHS_.get(), memspace_);
```
shakedregev: We have MINUS_ONE as a constant. Maybe use that.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1012 @@
...
+      vector_handler_.scal(1 - theta, y_interp_.get(), memspace_);
+    }
+    else
+    {
+      BUBBLE_FAIL(y_interp_->setToZero(memspace_));
+    }
+
+    vector_handler_.axpy(1.0, y_cur_.get(), y_interp_.get(), memspace_);
+    vector_handler_.scal(theta, y_interp_.get(), memspace_);
+    vector_handler_.axpy(1 - theta, y_prev_.get(), y_interp_.get(), memspace_);
```
shakedregev: use constants.hpp

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: `\f[` should be correct here - it is used for inserting a math display rather than inline math. With display:
image

With inline:
image

And it needs to match the final delimiter (currently creates an error).

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
```diff
@@ -0,0 +1,440 @@
...
+
+      std::vector<ScalarT>& y() override
+      {
+        return y_;
+      }
+
+      const std::vector<ScalarT>& y() const override
+      {
+        return y_;
+      }
```
shakedregev, line 195: What's the rationale of const and nonconst functions of the same thing?

shakedregev (approved):
Fix the minor issues and good to merge. I tested it.

GridKit/Solver/Dynamic/Rosenbrock.cpp
shakedregev: Oh, where are you seeing the docs this is generating? It seemed to me like you had an opening \left with a [ and then a closing \right with a )?

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
alexander-novo: Well here it's because I'm implementing an interface which requires both (`GridKit::Model::Evaluator`).

But in general in C++ this is to allow for the flexibility of modifying the resulting vector if you have a non-`const` reference to the model (top function). If you only have the bottom function, you won't be able to modify the vector the reference is pointing to. If you only have the top function, you won't be able to call `y()` on a `const` model (even if you aren't going to modify it later).

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1012 @@
+#include "Rosenbrock.hpp"
+
+#include <format>
+#include <iomanip>
+#include <sstream>
+
+#include <sundials/sundials_types.h>
```
nkoukpaizan:
```suggestion

```
I'm getting an error because SUNDIALS is not linked to this solver through CMake. As far as I can tell, this is only used to get `sunrealtype` in the template instantiation, so I recommend removing the dependency and using `double` directly.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: If you run the `GridKitDocs` target in cmake, it will generate the Doxygen docs in your output directory.

In Doxygen you have to demark a latex math block with the special commands `\f( ... \f)` (for inline) or `\f[ ... \f]` (for display). So that first brackets was just a Doxygen command which is followed by a `\left(` which is a latex macro.

README.md
```diff
@@ -28,6 +28,7 @@ You should have all of the following installed before installing GridKit™
       that the use of Enzyme is experimental, and some versions of it have been found to break GridKit code.
         - [LLVM](https://github.com/llvm/llvm-project) >= 15.x. GridKit is
           currently tested with LLVM 16.
+    - [ReSolve](https://github.com/ORNL/ReSolve) >= commit `03d11bdf9e3b41bc2ca7ecf7dd94f7eb8a73f2e2` (optional)
```
nkoukpaizan: There have been some interface-breaking changes in Re::Solve since that commit, so current `develop` does not work. I would recommend updating this PR for the current Re::Solve develop. We'll have to be careful here if we do add the dependency. In the README, we may want a `=` rather than `>=` for now. A release will be preferable in due time.
```
GridKit/GridKit/Solver/Dynamic/Rosenbrock.cpp:349:49: error: too few arguments to function call, single argument 'side' was not specified
  349 |     BUBBLE_FAIL(lin_solver_.preconditionerSetup());
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
GridKit/GridKit/Solver/Dynamic/Rosenbrock.cpp:20:20: note: expanded from macro 'BUBBLE_FAIL'
   20 |     if (int err = (arg)) \
      |                    ^~~
buildsystem/spack/opt/spack/linux-zen3/resolve-develop-zgkswmnetw6ht6hkr5rxjfedxaje6qkh/include/resolve/SystemSolver.hpp:64:9: note: 'preconditionerSetup' declared here
   64 |     int preconditionerSetup(std::string side);
      |         ^                   ~~~~~~~~~~~~~~~~
```

alexander-novo: @pelesh What makes sense to do here? Rosenbrock doesn't care about preconditioning - this function is only called so that if the user configured preconditioning then it runs properly. But I don't want to add a left/right option to Rosenbrock because that isn't something Rosenbrock cares about.

alexander-novo: I talked with Slaven and we're going to push this down the line until ReSolve can tell us whether or not there's a preconditioner configured.

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
```diff
@@ -0,0 +1,204 @@
+#include <sundials/sundials_types.h>
```
nkoukpaizan:
```suggestion
```

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
```diff
@@ -0,0 +1,204 @@
...
+    re.H_[2 * 8 + 3] = 80.12511834622643;
+    re.H_[2 * 8 + 4] = -52.072871366152654;
+    re.H_[2 * 8 + 5] = -59.78993625266729;
+    re.H_[2 * 8 + 6] = -0.15582684282751913;
+    re.H_[2 * 8 + 7] = 4.883087185713722;
+
+    return re;
+  }
+
+  template class Rosenbrock<sunrealtype, int>;
```
nkoukpaizan:
```suggestion
  template class Rosenbrock<double, int>;
```

tests/UnitTests/Solver/Dynamic/CMakeLists.txt
```diff
@@ -1,9 +1,18 @@
...
+  add_test(NAME IDATest COMMAND $<TARGET_FILE:test_ida>)
 
-install(TARGETS test_ida RUNTIME DESTINATION bin)
+  install(TARGETS test_ida RUNTIME DESTINATION bin)
+endif()
+
+if(GRIDKIT_ENABLE_RESOLVE)
+  add_executable(test_rosenbrock runRosenbrockTests.cpp)
+  target_link_libraries(test_rosenbrock GridKit::rosenbrock GridKit::testing)
+  add_test(NAME ROSTest COMMAND $<TARGET_FILE:test_rosenbrock>)
```
nkoukpaizan:
```suggestion
  add_test(NAME RosenbrockTest COMMAND $<TARGET_FILE:test_rosenbrock>)
```

GridKit/MemoryUtilities/ResolveMemoryUtils.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace GridKit
+{
+  namespace memory
+  {
+    /**
+     * @brief Converts a GridKit \ref MemorySpace to its corresponding Re::Solve MemorySpace
+     *
+     */
+    inline ReSolve::memory::MemorySpace memorySpaceAsResolve(MemorySpace memspace)
```
pelesh: This should not be necessary. GridKit should not see Re::Solve memory space and vice versa.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,760 @@
...
+namespace Integrator
+{
+
+  /**
+   * @brief Define control flow for `StepController`s to be able to control the step size of a `Rosenbrock` integrator.
+   *
+   */
+  template <typename RealT>
+  struct StepControl
+  {
```
pelesh: This structure and everything that is not `Rosenbrock` class should be in separate file. For each class or structure there should be a separate *.hpp and *.cpp files.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,760 @@
...
+
+  /**
+   * @brief Interface for step size controllers. Used by `Rosenbrock` integrators to decide when to accept/reject steps and
+   *        what size each step should be.
+   *
+   * @todo It may be best to have \ref usesError() return a reference to the \ref ErrorNorm that should be used.
+   */
+  template <typename RealT>
+  class StepController
+  {
```
pelesh: Same here. Please move this to a separate file.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,760 @@
...
+     *
+     */
+    Stats stats_;
+
+  public:
+    Rosenbrock(Tableau&&                                             tab,
+               GridKit::Model::Evaluator<ScalarT, IdxT>*             model,
+               ReSolve::SystemSolver&                                lin_solver,
+               GridKit::LinearAlgebra::VectorHandler<ScalarT, IdxT>& vector_handler,
+               const ErrorNorm<ScalarT, IdxT>*                       err_norm,
```
pelesh: `VectorHandler` and `ErrorNorm` probably should be members of `Rosenbrock` class. I am not sure why these are passed as constructor parameters.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,760 @@
...
+  };
+
+  /**
+   * @brief A simple textbook adaptive `StepController` which seeks to meet a relative and absolute tolerance
+   *        based on an error estimate.
+   *
+   */
+  template <typename RealT>
+  class AdaptiveStep : public StepController<RealT>
+  {
```
pelesh: This class should be in a separate file.

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
```diff
@@ -0,0 +1,204 @@
+#include <sundials/sundials_types.h>
+
+#include "Rosenbrock.hpp"
+
+namespace Integrator
+{
+  template <class ScalarT, typename IdxT>
+  Rosenbrock<ScalarT, IdxT>::Tableau Rosenbrock<ScalarT, IdxT>::Tableau::linImplicitEuler()
+  {
+    constexpr size_t num_stages = 1;
```
pelesh: I would keep tableaus in the `Rosenbrock.cpp` file.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1028 @@
...
+   * @param err \f(\hat{e}\f) in the above formula.
+   * @param y \f(y_1\f) in the above formula.
+   * @param yprev \f(y_0\f) in the above formula.
+   * @param handler The handler to be used for performing linear algebra operations.
+   * @param memspace The memory space to be used for performing linear lagebra operations.
+   * @see `Rosenbrock::errorEstimate()`
+   */
+  template <class ScalarT, typename IdxT>
+  InfNorm<ScalarT, IdxT>::RealT InfNorm<ScalarT, IdxT>::errorNorm(State& err, State& y, State& yprev, GridKit::LinearAlgebra::VectorHandler<ScalarT, IdxT>& handler, ReSolve::memory::MemorySpace memspace) const
+  {
```
pelesh: Re::Solve memory space should not show up here. It is internal to Re::Solve and should not be used in GridKit's `VectorHandler`.

GridKit/Solver/Dynamic/Rosenbrock.cpp
```diff
@@ -0,0 +1,1028 @@
...
+    y_interp_                   = std::make_unique<State>(size);
+    workspace_.asum_            = std::make_unique<State>(size);
+    workspace_.csum_            = std::make_unique<State>(size);
+    workspace_.RHS_             = std::make_unique<State>(size);
+    workspace_.RHS_first_stage_ = std::make_unique<State>(size);
+    workspace_.dFdt_            = std::make_unique<State>(size);
+    workspace_.mass_            = std::make_unique<State>(size);
+
+    resolve_rhs_ = std::make_unique<ReSolve::vector::Vector>(size);
+    resolve_lhs_ = std::make_unique<ReSolve::vector::Vector>(size);
```
pelesh: Re::Solve vector should not appear in Rosenbrock class.

pelesh (commented):
Very nice compact implementation.

Before merging it would be helpful to create clear interfaces between Rosenbrock integrator and GridKit and between Rosenbrock integrator and linear solver.

Also, please implement each class/structure in a separate source/header files.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: How do I use a ReSolve linear solver without using ReSolve vectors?

GridKit/Solver/Dynamic/RosenbrockTableaus.cpp
alexander-novo: Are you sure about that? It's pretty standard to put tableau coefficients in a separate file from the implementation because they aren't really code meant to be looked at and can get quite bulky. See SUNDIALS ARKODE: https://github.com/llnl/sundials/blob/main/src/arkode/arkode_butcher_erk.def

GridKit/Solver/Dynamic/Rosenbrock.hpp
alexander-novo: They are members of `Rosenbrock`. How do I construct them if I don't pass them in as constructor parameters?

GridKit/MemoryUtilities/ResolveMemoryUtils.hpp
alexander-novo: I need a memory space for the Re::Solve vectors and matrices which are pointing to GridKit vectors and matrices. So the Re::Solve memory space needs to be the same memory space as the GridKit one.

pelesh: This is better done through the linear solver interface. You would pass the CSR matrix from Rosenbrock to Re::Solve as pointers to the three CSR matrix data arrays, nnz, and n, and you would pass rhs vector as a raw pointer to the vector data and the size n. Re::Solve returns the solution in the same manner.

The interface should use only Re::Solve `SystemSolver` API.  That way, you don't need to include Re::Solve specific memory management in Rosenbrock and/or GridKit.

Perhaps we should leave it for #489?

GridKit/Solver/Dynamic/ErrorNorm.hpp
```diff
@@ -0,0 +1,35 @@
...
+     * @param y The state that the error was calculated from. Can be used for proper relative error normalization.
+     * @param yprev The state from the previous step. Can be used for proper relative error normalization.
+     * @param handler A vector handler which can be used to facilitate vector operations.
+     * @param memspace The memory space which vector operations should be performed in/.
+     * @return The error.
+     *
+     * @todo Allow this method to fail, since it will likely involve linear algebra calls.
+     */
+    virtual RealT errorNorm(State& err, State& y, State& yprev, GridKit::LinearAlgebra::VectorHandler<ScalarT, IdxT>& handler, GridKit::memory::MemorySpace memspace) const = 0;
+  };
```
pelesh: Why is this pure virtual? Also, have you considered adding `errorNorm` as a method in Rosenbrock proper?

GridKit/Solver/Dynamic/AdaptiveStep.hpp
```diff
@@ -0,0 +1,69 @@
...
+     * @brief Parameters for the step controller.
+     *
+     */
+    struct Parameters
+    {
+      /**
+       * @brief The minimum multiple by which the step size can be multiplied to obtain the new step size.
+       *        Increasing this can allow the integrator to be slightly more conservative in selecting the step size
+       *        - decreasing the number of steps taken but increasing the risk of failing the next step.
+       *
```
pelesh: Do we really need a nested `struct` in this class? The class itself is small, nesting a structure with 3 real numbers looks like an overkill to me.

GridKit/Solver/Dynamic/InfNorm.hpp
```diff
@@ -0,0 +1,75 @@
...
+      RealT rel_tol_;
+    } params_;
+
+    InfNorm(Parameters&& params)
+      : params_(std::move(params))
+    {
+    }
+
+    RealT errorNorm(State& err, State& y, State& yprev, GridKit::LinearAlgebra::VectorHandler<ScalarT, IdxT>& handler, GridKit::memory::MemorySpace memspace) const final;
+  };
```
pelesh: We added `VectorHandler` and `MemoryManager` to facilitate development of portable solvers native to GridKit. In this case I would suggest making these class members. 
```suggestion
    RealT errorNorm(State& err, State& y, State& yprev) const final;

private:
    GridKit::LinearAlgebra::VectorHandler<ScalarT, IdxT> handler_;
    GridKit::memory::MemorySpace memspace_;
  };
```

GridKit/Solver/Dynamic/InfNorm.hpp
```diff
@@ -0,0 +1,75 @@
...
+       */
+      std::unique_ptr<State> abs_tol_;
+
+      /**
+       * @brief The relative tolerance. The norm will attempt to reject any error larger in percentage of
+       *        the solution's maximum element than this.
+       *
+       */
+      RealT rel_tol_;
+    } params_;
```
pelesh: As previously suggested, a nested `struct` may be an overkill here.

GridKit/Solver/Dynamic/Rosenbrock.cpp
pelesh: xSDK guidelines require that all libraries must be able to turn off all the outputs. That is why we are streaming everything through the logger.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,514 @@
...
+     * @brief The model being simulated.
+     *
+     */
+    GridKit::Model::Evaluator<ScalarT, IdxT>*             model_;
+    /**
+     * @brief The linear solver to be used during integration in \link timeStep() \endlink.
+     *
+     */
+    GridKit::LinearAlgebra::LinearSolver<ScalarT, IdxT>&  lin_solver_;
+    /**
```
pelesh: Okay, I understand now where you are aiming with this. The idea is for GridKit to provide _all_ linear algebra tooling and Rosenbrock to leverage that, correct?

In this case, you also do not need to know about Re::Solve's internal memory management. This will be wrapped within `GridKit::LinearAlgebra::LinearSolver` interface.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,514 @@
...
+  /**
+   * @brief A Rosenbrock integrator designed to simulate a `GridKit::Model::Evaluator` model.
+   *        Rosenbrock integrators are best for quick medium-accuracy simulation and
+   *        boast a large amount of customization to simulating a given model.
+   *
+   *        For the list of available Rosenbrock methods, see `Rosenbrock::Tableau`.
+   */
+  template <class ScalarT, typename IdxT>
+  class Rosenbrock
+  {
```
pelesh: I think either Rosenbrock or its public interface should inherit from `DynamicSolver` class. The `DynamicSolver` should be the one to own the GridKit model.

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
```diff
@@ -0,0 +1,445 @@
...
+          sol_norm += std::pow(tanh(final_time), 2);
+
+          errors.push_back(std::sqrt(error) / std::sqrt(sol_norm));
+        };
+
+        for (size_t i = 0; i < num_samples; i++)
+        {
+          double step_size = std::pow(10, step_exponent_lower + static_cast<double>(i) * (step_exponent_upper - step_exponent_lower) / static_cast<double>(num_samples - 1));
+          double num_steps = round((final_time - 0.5) / step_size);
+          step_size        = (final_time - 0.5) / num_steps;
```
pelesh: The unit test code needs to be documented better. It would be helpful to provide more detailed comments to  explain each step in the testing. The code is not "self-documenting" 😄

pelesh (approved):
This is very nice work and we can merge it to the feature branch provided the comments below are addressed:
- `Rosenbrock` class needs to inherit from `DynamicSolver` class so it can be called interchangeably with IDA (and I understand there will be some additional work there but we need to open that door now).
- Files `Rosenbrock.*pp` should stay in `GridKit/Solver/Dynamic` directory. For all other files, I suggest moving them to a new directory `GridKit/Solver/Dynamic/Rosenbrock`.
- `Rosenbrock` and related classes should be in `AnalysisManager` namespace, and the `Integrator` namespace should be within that to make it consistent with IDA interface.
- Consider renaming `Integrator` namespace with something like `NativeDynamicSolver` or something similar to describe these are solvers native to GridKit, not third party ones (such as SUNDIALS).
- File `ResolveMemoryUtils.hpp` should not be needed if the linear solver interface is implemented correctly but we can handle that in #489.
- A few minor comments below.

GridKit/Solver/Dynamic/InfNorm.hpp
pelesh: You could also make this class `friend` of Rosenbrock, so it can access memory and linear algebra utilities there.

GridKit/Solver/Dynamic/ErrorNorm.hpp
alexander-novo: `ErrorNorm` is just an interface which is implemented by `InfNorm`. There are different error norms one could want to use. This PR only includes the infinity norm, but the 2 norm is also quite common.

GridKit/Solver/Dynamic/Rosenbrock.cpp
alexander-novo: There is no output here. Nothing needs to be turned off. There is nothing to log. The function just produces a string variable in memory. Nothing is sent to standard output or error.

alexander-novo:
@pelesh 

I'm not sure what to do about implementing `DynamicSolver` for `Rosenbrock`. The method `setTolerance` doesn't make sense, because the Rosenbrock method doesn't have a tolerance inherently.

GridKit/MemoryUtilities/ResolveMemoryUtils.hpp
alexander-novo: This has already been implemented in #489 - the file was accidentally left here and was unused. I removed it.

GridKit/Solver/Dynamic/AdaptiveStep.hpp
alexander-novo: The nice thing about having a nested Struct is that it is very easy to set default parameters. The user can create an instance of `Parameters` with all of the defaults and then change a single one of them with 2 lines of code. But then once the parameters are given to the step controller, they are no longer editable.

To support the same operation with these as normal class members of `AdaptiveStep` would involve having to write getters/setters for every single one of the parameters.

To me a nested struct isn't "overkill" - it doesn't add any maintenance burden or anything. In my opinion it actually removes maintenance burden in this case.

GridKit/Solver/Dynamic/InfNorm.hpp
alexander-novo: The `friend` approach won't work. These are not Rosenbrock-specific operations and are designed to work with multiple types of integrators. And even if we made them Rosenbrock-specific, the functions would then have to be passed a reference to the entire `Rosenbrock` instance just to access the individual pieces that were being passed originally. It opens a can of ~worms~ cyclic references that I don't think is worth it - what if the `ErrorNorm` calls a `Rosenbrock` function that then calls an `ErrorNorm` function?

alexander-novo: For the originally suggested approach I have to ask: is it fine to copy a `VectorHandler`? Could it in the future have some sort of state in it?

GridKit/Solver/Dynamic/Rosenbrock.cpp
pelesh: Sorry, my bad 😄

GridKit/Solver/Dynamic/InfNorm.hpp
pelesh: Original idea was to have one `VectorHandler` per solver, so copying it is probably not a good idea. I see where you are heading with this development, thanks for clarifying.

PR #481, opened 2026-07-09
Replace ComponentSignals with IOPorts implementation

PhilipFackler:
## Description
 
This is a refactor of `ComponentSignals` to be (hopefully) more intuitive and also make model development a bit easier.
 
 ## Proposed changes
 
Manage signal node connections via input and output ports
- Introduce a set of classes that represent a collection of input and output ports that manage connection to signal nodes
- Make input/output ports directly available to components
- Clear up terminology: distinguish between signal node and signal. Use "connect" rather than "assign" and "attach" (keep "assign" as implementation detail) and "link" rather than "set"
- Introduce `SignalNodeSet` class for managing the signal nodes for the system. This can then be passed to components since it doesn't require access to the `SystemModel` implementation. This maps either signal_id or name to signal nodes.
- Have `IOPorts` automatically connect signal nodes to ports at construction (rather than doing it manually in `SystemModel`
- Remove `ComponentSignals`
 
 ## Checklist
 
- [X] All tests pass.


lukelowry:
This looks awesome!!

lukelowry (approved):
So so helpful for my development. This is very intuitive and helps understand and debug intermodel behavior. I would like this to be merged before the new Phasor Dynamics models so I can use this syntax!

Masterful this helps so much with modeling

lukelowry:
As a related question, for future development in EMT: how difficult would this be to implement for vector-typed signals instead of scalar-typed? In an ideal world, the best possible design for EMT is a N width signal (decided at runtime), which I have no clue how to implement.

PhilipFackler:
> As a related question, for future development in EMT: how difficult would this be to implement for vector-typed signals instead of scalar-typed? In an ideal world, the best possible design for EMT is a N width signal (decided at runtime), which I have no clue how to implement.

Handling the variable should be pretty simple. Especially if it can be wrapped in a non-owning vector. That's what `readSignal` would return. Otherwise it would work exactly the same way.

The more challenging part would be the use of `SignalNode`. Would we need to re-implement (duplicate) things, or could we generalize?

GridKit/Model/PhasorDynamics/IOPorts.hpp
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * @brief Base class representing a port (connection to a SignalNode)
+     *
+     * An input or output port may be connected to one SignalNode
+     */
+    template <typename scalar_type, typename index_type>
+    class Port
+    {
+    public:
+      /// Scalar type used for signals
```
pelesh, line 22: Class `Port` must be in separate source file from `IOPorts`.

GridKit/Model/PhasorDynamics/IOPorts.hpp
```diff
@@ -0,0 +1,281 @@
...
+      /// Instantiation of signal node type
+      using SignalNodeT = SignalNode<ScalarT, IdxT>;
+
+      /**
+       * @brief Connect a signal node to the port
+       *
+       * @param node SignalNode to be connected
+       */
+      void connect(SignalNodeT* node) noexcept
+      {
```
pelesh, line 35: Node and port are often used interchangeably in nodal analysis. What is node in this case?

GridKit/Model/PhasorDynamics/IOPorts.hpp
```diff
@@ -0,0 +1,281 @@
...
+
+      SignalNodeT* signal_node_{nullptr};
+    };
+
+    /**
+     * @brief Port for sending a signal, making it available via a SignalNode
+     */
+    template <typename scalar_type, typename index_type>
+    class OutputPort : public Port<scalar_type, index_type>
+    {
```
pelesh, line 81: I believe we decided to name this `SignalOut`. This should also be in a separate file.

GridKit/Model/PhasorDynamics/IOPorts.hpp
```diff
@@ -0,0 +1,281 @@
...
+      }
+    };
+
+    /**
+     * @brief Port for receiving a signal from a SignalNode
+     */
+    template <typename scalar_type, typename index_type>
+    class InputPort : public Port<scalar_type, index_type>
+    {
+    public:
```
pelesh, line 115: Rename to `SignalIn` or `SignalInput` and implement in a separate file.

GridKit/Model/PhasorDynamics/IOPorts.hpp
```diff
@@ -0,0 +1,281 @@
...
+     * @tparam model_data_type "<model>Data" type specifying variables. This is
+     * expected to be (or derive from) an instantiation of ComponentData. This
+     * provides the SignalInputs and SignalOutputs enumeration types for
+     * specifying each PortGroup
+     */
+    template <typename scalar_type, typename model_data_type>
+    struct IOPorts
+    {
+      /// Scalar type used for signals
+      using ScalarT        = scalar_type;
```
pelesh, line 241: More intuitive name would be `SignalPorts`. Needs to be implemented in a separate file.

tests/UnitTests/PhasorDynamics/ExciterSexsPtiTests.hpp
```diff
@@ -107,13 +109,15 @@ namespace GridKit
...
-        auto                                            data = makeTestData();
-        PhasorDynamics::Exciter::SexsPti<ScalarT, IdxT> exciter(&bus, data);
-        exciter.getSignals().template assignSignalNode<PhasorDynamics::Exciter::SexsPtiInternalVariables::EFD>(&efd_node);
-        exciter.getSignals().template attachSignalNode<PhasorDynamics::Exciter::SexsPtiExternalVariables::VS>(&vs_node);
+        using namespace GridKit::PhasorDynamics::Exciter;
+
+        auto data    = makeTestData();
+        auto exciter = SexsPti<ScalarT, IdxT>(&bus, data);
+        exciter.getPorts().out[SexsPtiSignalOutputs::efd].connect(&efd_node);
+        exciter.getPorts().in[SexsPtiSignalInputs::vs].connect(&vs_node);
```
pelesh, line 120: A better approach would be to have methods `getSignalInput()` and getSignalOutput()`.
```suggestion
        exciter.getSignalOutput[SexsPtiSignalOutputs::efd].connect(&efd_node);
        exciter.getSignalInput[SexsPtiSignalInputs::vs].connect(&vs_node);
```
Similar for other components.

GridKit/Model/PhasorDynamics/SignalNode/SignalNode.hpp
```diff
@@ -19,6 +19,17 @@ namespace GridKit
...
+     * A signal node conceptually carries a signal (scalar value) from an
+     * output port of one component to an input port of another component (or
+     * multiple components).
+     *
+     * (Component):[OutputPort] -> {SignalNode} -> [InputPort]:(Component)
+     *
+     * A SignalNode can be "connected" to a Port. When that port is an
+     * OutputPort, the SignalNode is considered `assigned()` since it can be
+     * connected to only one OutputPort. The SignalNode is considered `linked()`
+     * when the actual signal (scalar variable) has been made available.
```
pelesh, line 31: There seems to be some confusion with naming scheme here: The output port is a signal node. What is called a `SignalNode` here is a signal link.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -99,14 +99,9 @@ namespace GridKit
...
-        auto getSignals()
-            -> ComponentSignals<ScalarT,
-                                IdxT,
-                                IeeestInternalVariables,
-                                IeeestExternalVariables>&
+        IOPortsT& getPorts()
         {
-          return signals_;
+          return ports_;
         }
```
pelesh, line 105: I would create separate methods for input and output signal ports. These are two different types of objects with different connection properties.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/Ieeest.hpp
```diff
@@ -79,15 +76,18 @@ namespace GridKit
...
+        using ModelDataT     = IeeestData<RealT, IdxT>;
+        using SignalNodeT    = SignalNode<ScalarT, IdxT>;
+        using SignalNodeSetT = SignalNodeSet<SignalNodeT>;
+        using IOPortsT       = IOPorts<ScalarT, ModelDataT>;
+        using MonitorT       = Model::VariableMonitor<Ieeest, IeeestData>;
 
         Ieeest();
         Ieeest(const ModelDataT& data);
+        Ieeest(const ModelDataT& data, SignalNodeSetT& signal_nodes);
         ~Ieeest();
```
pelesh, line 91: The constructor should not take reference to `SignalNodeSetT` as the input parameter. All signal connectivity information is already in `ModelDataT`. Similar for other component models.


```suggestion
~Ieeest();
```

pelesh (changes requested):
This is certainly step in the right direction and makes component model signal connections more streamlined.

A few things that need to be addressed before this is ready for review:
- Component model constructors should take only reference to the component model data as input parameters. All signal connectivity information is already in there.
- Treat signal input and outputs separately as these are different types of ports (nodes).
- Implement different classes/structs in separate source files.
- Align naming of the object with what we have agreed earlier.

lukelowry:
> > As a related question, for future development in EMT: how difficult would this be to implement for vector-typed signals instead of scalar-typed? In an ideal world, the best possible design for EMT is a N width signal (decided at runtime), which I have no clue how to implement.
> 
> Handling the variable should be pretty simple. Especially if it can be wrapped in a non-owning vector. That's what `readSignal` would return. Otherwise it would work exactly the same way.
> 
> The more challenging part would be the use of `SignalNode`. Would we need to re-implement (duplicate) things, or could we generalize?

I hope I understand the actual challenge:

(technical info) It is common to formulate EMT in `abc` coordinates, so the signals have `size=3` system-wide. However, the general design would accommodate many physical wiring configurations. At the transmission level, the delta configuration (`size=3`) is expected. At the distribution level, we can see Wye configuration (`size=4`), or single-phase lines (`size=2`) for `Branch` components. Controls would ideally be generalized, but I think at bare minimum we must support vectorized signals of size  `size={1,2,3,4}` . But many real systems have  `size>4` (multiple conductors per phase).

In general, we can expect many different N-dimensional signal sizes to appear within a system, especially in combined T&D systems. So a generalized signal implementation would be very valuable. I hope this is what you were asking. Should probably make an issue for this

PR #483, opened 2026-07-10
Add new solver options to phasor dynamics application

Steven-Roberts:
## Description
 
I think these options will be needed for the fixed versus adaptive time step comparisons.
 
 ## Proposed changes
 
This adds JSON options for tolerances and a fixed time step
 
 ## Checklist

 
- [x] All tests pass.
 
 ## Further comments
 
I noticed the existing `dt` JSON option has a few issues. It is documented as solver time step but it seems to be the output/monitoring time step. Perhaps it should be rename `monitor_dt` and the new `fixed_dt` should be called `dt`. Also with this rounding, the monitoring is not guaranteed to occur every `dt` simulation time.
https://github.com/ORNL/GridKit/blob/c099ac0037b27c03d4ec9964cd3a006429a636eb/application/PhasorDynamics/DynamicSimulation.cpp#L48
Any suggestions on how to handle this @pelesh?

### DIFF
--- a/GridKit/Solver/Dynamic/Ida.cpp
+++ b/GridKit/Solver/Dynamic/Ida.cpp
@@ -270,8 +270,8 @@ namespace AnalysisManager
         return 1;
       }
 
-      const RealT n_est    = (tf - t_init_) / dt_monitor;
-      const RealT epsilon  = std::numeric_limits<RealT>::epsilon()
+      const RealT n_est   = (tf - t_init_) / dt_monitor;
+      const RealT epsilon = std::numeric_limits<RealT>::epsilon()
                             * std::max({std::abs(t_init_), std::abs(tf), RealT(1.0)})
                             / dt_monitor;
       return static_cast<int>(std::ceil(n_est - epsilon));
--- a/GridKit/Solver/Dynamic/Ida.cpp
+++ b/GridKit/Solver/Dynamic/Ida.cpp
@@ -270,9 +270,11 @@ namespace AnalysisManager
         return 1;
       }
 
-      const RealT duration = tf - t_init_;
-      const RealT epsilon  = std::numeric_limits<RealT>::epsilon() * std::max(duration, RealT(1.0));
-      return static_cast<int>(std::ceil((duration - epsilon) / dt_monitor));
+      const RealT n_est    = (tf - t_init_) / dt_monitor;
+      const RealT epsilon  = std::numeric_limits<RealT>::epsilon()
+                            * std::max({std::abs(t_init_), std::abs(tf), RealT(1.0)})
+                            / dt_monitor;
+      return static_cast<int>(std::ceil(n_est - epsilon));
     }
 
     /**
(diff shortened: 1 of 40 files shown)


pelesh:
Yes, `dt` is output monitoring step for adaptive step solvers.

pelesh:
> Also with this rounding, the monitoring is not guaranteed to occur every dt simulation time.

I believe original idea was to have something like this:
```c++
for (i = 0; i < nout; ++i) {
  ...
  t += dt;
}
```
It was originally intended for `nout` to be an approximate number of outputs and output steps to be precisely of size `dt`.

pelesh:
I like the idea of `dt` being the fixed time step and `dt_moinitor` to be output time step. Even if we use fixed time step, which is typically a quarter of a cycle, we probably don't want to output values at each step but say every 10th or so.

CC @lukelowry @abirchfield

lukelowry:
(related note) For validation, we need `dt` to be a quarter cycle so we have comparable samples against PowerWorld. 

The large cases produce a lot of monitored output, @pelesh is right in his suggestion that `dt_monitor`  ~10x the fixed time step `dt` for practical use. These are the same in existing solvers, so the distinction is usually moot. But we should make the distinction here

I like the naming `dt_monitor`, good suggestion. To prevent confusion, would `dt_fixed` be clearer than `dt`?

PhilipFackler (approved):
This is great. Thanks for the correction. I'll let the experts decide what is most intuitive for users, but maybe `dt_solver` would be clearer than `dt_fixed`. Not requesting the change, just throwing in my two cents (pun not intended).

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -268,40 +256,66 @@ namespace AnalysisManager
...
-     * @note The actual time of the final IDA solution should be somewhat
-     * close to `tf`, however due to rounding error the precise final time may
-     * be before or after `tf`.
+     * The final monitor target is pinned exactly to `tf` to avoid roundoff in
+     * repeated time-step arithmetic.
+     */
+    template <class ScalarT, typename IdxT>
+    typename Ida<ScalarT, IdxT>::RealT Ida<ScalarT, IdxT>::getMonitorTime(RealT tf, RealT dt_monitor, int step, int nsteps) const
+    {
+      return step == nsteps ? tf : std::fma((RealT) step, dt_monitor, t_init_);
```
nkoukpaizan, line 289: `fma` might be overkill for a scalar, non critical operation, but not a blocker!

nkoukpaizan (approved):
Looks good!

Will need rabasing.

GridKit/Solver/Dynamic/Ida.cpp
Steven-Roberts: I tried to reduce floating point errors as much as possible after running into a few issues with floating point comparisons. It's probably fine either way.

PR #486, opened 2026-07-10
Distinction of Magnetization vs. Line Shunt  (Validation)

lukelowry:
## Description

Implementation on `develop` treats the shunt as a magnetizing admittance, which should have a distinct parameter from the line shunt admittance, which does not observe the effects of phase/tap changes. 

Before these changes, the Branch model is presumed to be either a XFMR or a line but not a line & XFMR in series. We need to support this edge case

MATPOWER does not seem to make this distinction, but PowerWorld models this correctly. This contributed marginally to errors in magnetization admittances in some cases.
 
Required before `0.2` release.
 
 ## Proposed changes
 
- Add magnetic charging admittance term
- Move G and B contribution outside of tap/phase transformation
 
 ## Checklist

 
- [X] All tests pass.
- NA I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
Need to update the circuit mdoel diagram in future

### DIFF
--- a/GridKit/Model/PhasorDynamics/Branch/Branch.hpp
+++ b/GridKit/Model/PhasorDynamics/Branch/Branch.hpp
@@ -204,6 +204,8 @@ namespace GridKit
       RealT X_{0.0};
       RealT G_{0.0};
       RealT B_{0.0};
+      RealT Gmag_{0.0};
+      RealT Bmag_{0.0};
       RealT tap_{1.0};
       RealT phase_{0.0};
       IdxT  bus1_id_{0};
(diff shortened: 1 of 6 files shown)


lukelowry:
An updated circuit diagram is outside the scope of this PR, so marking ready. The proposed changes also help future proof our `Branch` model so it can be used for XFMR, line, and composite XFMR-line model.

abirchfield (approved):
Yes, this correctly handles the distinction between line shunts and XFMR magnetizing impedance, and is important to get into the code for some of the larger cases.

nkoukpaizan (approved):

PR #488, opened 2026-07-13
Use GridKit Vector for Model Evaluator Interfaces

lukelowry:
## Description

Updates GridKit model interfaces so numerical model vectors use `GridKit::LinearAlgebra::Vector` instead of `std::vector`.

A Vector interface is needed for later zero-copy work while keeping differential/algebraic tags as `std::vector<bool>`.

## Proposed changes

- Add `VectorT` to `Model::Evaluator` and use it for state, derivative, residual, tolerance, parameter, quadrature, and adjoint vector interfaces.
- Update PhasorDynamics, PowerFlow, and PowerElectronics models to allocate and access `Vector` host storage.
- Update IDA, KINSOL, optimization, Enzyme, and dependency-tracking paths for the new interface.
- Add the required `Vector` instantiations and target dependencies, and update examples and tests.

## Checklist

- [x] All tests pass for the focused coverage (except known PE failure).

Changelog changes N/A

## Further comments

This is the Vector interface work split out from #452. It intentionally keeps the differential/algebraic tag as `std::vector<bool>` and does not include the global state-view changes.


GridKit/Model/PhasorDynamics/Component.hpp
```diff
@@ -252,11 +278,13 @@ namespace GridKit
...
 
-      std::vector<ScalarT> y_;
-      std::vector<ScalarT> yp_;
-      std::vector<bool>    tag_;
-      std::vector<ScalarT> abs_tol_;
-      std::vector<ScalarT> f_;
+      VectorT           y_;
+      VectorT           yp_;
+      std::vector<bool> tag_;
+      VectorT           abs_tol_;
```
pelesh, line 271: Let's keep `tag_` as a `std::vector` for now and address this issue in a separate PR.

GridKit/Model/PowerFlow/SystemModelPowerFlow.hpp
```diff
@@ -198,9 +202,14 @@ namespace GridKit
 
       for (const auto& component : components_)
       {
-        for (IdxT j = 0; j < component->size(); ++j)
+        if (component->size() > 0)
         {
-          y_[varOffset + j] = component->y()[j];
+          auto* component_y = component->y().getData();
+
```
pelesh: Shouldn't this be a const pointer? 
```suggestion
          const auto* component_y = component->y().getData();

```

pelesh: When accessing read-only data, it should be accessed through a const pointer. Please check other places in the code.

GridKit/Model/PowerFlow/SystemModelPowerFlow.hpp
```diff
@@ -283,18 +305,28 @@ namespace GridKit
       IdxT resOffset = 0;
       for (const auto& bus : buses_)
       {
-        for (IdxT j = 0; j < bus->size(); ++j)
+        if (bus->size() > 0)
         {
-          f_[resOffset + j] = bus->getResidual()[j];
+          auto* bus_f = bus->getResidual().getData();
+
```
pelesh: This should be a const pointer?

```suggestion
          const auto* bus_f = bus->getResidual().getData();

```

GridKit/Model/PowerFlow/SystemModelPowerFlow.hpp
```diff
@@ -283,18 +305,28 @@ namespace GridKit
...
       }
 
       for (const auto& component : components_)
       {
-        for (IdxT j = 0; j < component->size(); ++j)
+        if (component->size() > 0)
         {
-          f_[resOffset + j] = component->getResidual()[j];
+          auto* component_f = component->getResidual().getData();
+
```
pelesh:
```suggestion
          const auto* component_f = component->getResidual().getData();

```

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -842,52 +839,53 @@ namespace AnalysisManager
...
-        std::cerr << "\nstd::vector size (" << x.size() << ") does not match N_Vector size ("
+        std::cerr << "\nvector size (" << x.getSize() << ") does not match N_Vector size ("
                   << ysize << ").\n\n";
         throw SundialsException();
       }
 
       ScalarT* ydata = N_VGetArrayPointer(y);
-      std::copy(x.cbegin(), x.cend(), ydata);
+      std::copy_n(x.getData(), xsize, ydata);
     }
```
pelesh: This may be redundant but just to emphasize `xdata` is not modified:
```suggestion
      const auto* xdata = x.getData();
      std::copy_n(xdata, xsize, ydata);
}
```

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -842,52 +839,53 @@ namespace AnalysisManager
...
-                  << y.size() << ").\n\n";
+        std::cerr << "\nN_Vector size (" << xsize << ") does not match vector size ("
+                  << y.getSize() << ").\n\n";
         throw SundialsException();
       }
 
       const ScalarT* xdata = N_VGetArrayPointer(x);
-      std::copy_n(xdata, y.size(), y.begin());
+      std::copy_n(xdata, ysize, y.getData());
     }
```
pelesh, line 868: Once data is updated, we need to tell that to `Vector`.
```suggestion
      std::copy_n(xdata, ysize, y.getData());
      y.setDataUpdated();
}
```

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
```diff
@@ -310,33 +315,38 @@ namespace GridKit
...
+          gov_yp[i].setVariableNumber(i); ///< Governor independent variables
         }
 
         bus.evaluateResidual();
         gen.evaluateResidual();
         gov.evaluateResidual(); // Computes the residual and the Jacobian values by tracking
                                 // the dependencies
-        std::vector<DependencyTracking::Variable> residual_yp = gov.getResidual();
+        auto&                                     residual_yp_view = gov.getResidual();
+        std::vector<DependencyTracking::Variable> residual_yp(residual_yp_view.getData(), residual_yp_view.getData() + residual_yp_view.getSize());
```
pelesh: I don't think you need to wrap this data into a `std::vector`. See similar in `SystemTests`.

tests/UnitTests/PhasorDynamics/SystemTests.hpp
```diff
@@ -232,28 +232,30 @@ namespace GridKit
...
 
         // Print the dependencies
-        for (size_t i = 0; i < residual.size(); ++i)
+        for (size_t i = 0; i < residual.getSize(); ++i)
         {
           std::cout << i << "th residual: ";
-          (residual[i]).print(std::cout);
+          residual_data[i].print(std::cout);
           std::cout << "\n";
         }
```
pelesh: I think this approach is preferred to using `std::vector` as in GovernorTgov1Tests.hpp`

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
```diff
@@ -251,23 +265,46 @@ namespace GridKit
...
+        yp_.allocate(memory::HOST);
+        yp_.setToZero(memory::HOST);
+
+        f_.resize(n);
+        f_.allocate(memory::HOST);
+        f_.setToZero(memory::HOST);
+
+        abs_tol_.resize(n);
+        abs_tol_.allocate(memory::HOST);
+        abs_tol_.setToZero(memory::HOST);
```
pelesh: Can't we just use `resize` only?

```suggestion
        y_.resize(n);
        yp_.resize(n);
        f_.resize(n);
        abs_tol_.resize(n);
```

pelesh (commented):
## Summary

Excellent work and well-scoped PR. There are a few minor forward looking suggestions I would like to make. Comments to specific places in the code are just examples. My suggestions are as follows:

### Marking model as allocated

Consider setting the flag `allocated_` in the `allocate` method of the `Component` (and similar) class rather than in `allocateVectors` protected method. Allocating 4 vectors there is not a complete allocation.

### Marking allocation as inadequate

Consider flipping flag `allocated_` to `false` whenever `SystemModel` methods like `addBuss` or `addComponent` are called. If you change the system, then the earlier allocation is not valid anymore.

### Using `Vector::resize()` method

In `allocateVectors()` function the code looks like this:
```c++
yp_.resize(n);
yp_.allocate(memory::HOST);
yp_.setToZero(memory::HOST);
```
The `resize` method should be sufficient here. If not, we can make necessary modifications within this PR.

### Use `const` pointers when accessing vector read-only data
When there is no intent to modify vector data, use const pointer. For example:
```c++
const auto* xdata = v.getData();
for (size_t i =0; i < v.getSize(); ++i)
  x[i] = xdata[i];
```

### Mark vector data as "updated" after modifying it
When you access `Vector`'s raw data and _modify_ it, you need to mark vector data as updated. This has little effect now when all of our computations are on the host, but will matter when we move to GPU devices. For example:
```c++
auto* xdata = v.getData();
for (size_t i =0; i < v.getSize(); ++i)
  xdata[i] = x[i];
v.setDataUpdated();
```
Vector needs to know which data (host or device) has the most recent values. I would argue, this call also improves the code readability because it shows intent explicitly.

CC @PhilipFackler

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
lukelowry: not as `Vector` is written to my understanding, since `resize` doesn't initialize or allocate. We could change the implementaiton of `resize`

GridKit/Model/PhasorDynamics/Component.hpp
lukelowry: It is `std::vector` I think this might be outdated

GridKit/Model/PowerFlow/SystemModelPowerFlow.hpp
lukelowry: Addressed in 0ac49f5d778cd12057e4c3768936f521ed347bbf I think I got them all

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
pelesh: I believe it pretty much does everything you need but please check it for yourself: https://github.com/ORNL/GridKit/blob/c099ac0037b27c03d4ec9964cd3a006429a636eb/GridKit/LinearAlgebra/Vector/Vector.cpp#L898

If any modifications are needed, we can take care of it as a part of this PR.

tests/UnitTests/PhasorDynamics/GovernorTgov1Tests.hpp
lukelowry: Fixed in [24d375c](https://github.com/ORNL/GridKit/pull/488/commits/24d375c8aff2bf95962d22fff2f548e5dbb70bca)

GridKit/Solver/Dynamic/Ida.cpp
lukelowry: Addressed in [cf94c55](https://github.com/ORNL/GridKit/pull/488/commits/cf94c558bcb7298e6ca99e7823dcfeb9202ddc38)

GridKit/Solver/Dynamic/Ida.cpp
lukelowry: Addressed in [0ac49f5](https://github.com/ORNL/GridKit/pull/488/commits/0ac49f5d778cd12057e4c3768936f521ed347bbf).

tests/UnitTests/PhasorDynamics/SystemTests.hpp
lukelowry: Addressed in [0ac49f5](https://github.com/ORNL/GridKit/pull/488/commits/0ac49f5d778cd12057e4c3768936f521ed347bbf).

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
lukelowry: Addressed in [eb37826](https://github.com/ORNL/GridKit/pull/488/commits/eb37826e6cbfb5624f138b0f691203f2133d7276). This does require one change to `Vector` so that unallocated `Vector` instances allocate space before resizing.

lukelowry:
I dislike the `setDataUpdated` its confusing terminology to me, but I otherwise agree on its utility and clarity

@pelesh ready for re-review, your suggestions added more churn, but everything is very consistent now.

GridKit/Model/PhasorDynamics/BusFault/BusFaultImpl.hpp
```diff
@@ -113,16 +113,17 @@ namespace GridKit
...
-      f_.resize(size);
-      y_.resize(size);
-      yp_.resize(size);
-      abs_tol_.resize(size);
       tag_.resize(size);
-      variable_indices_.resize(size);
-      residual_indices_.resize(size);
+
+      this->variable_indices_.resize(size);
+      this->residual_indices_.resize(size);
```
alexander-novo: Why was `this->` added when it was previously unnecessary?

alexander-novo (commented):
There's a lot of `this->` that got added that is maybe unnecessary. For variables which are inherited from parent classes it might be better to `using` them instead to bring them in scope for future developers.

GridKit/Model/PhasorDynamics/BusFault/BusFaultImpl.hpp
lukelowry: For me, it helps make the distinction that it is a member/function with access to the member variables of the `Component` instance, rather than a helper/convenience function (At least this was the convention in my native language, Java!). If this is more confusing than clarifying in C++, I will happily change it back. lmk your thoughts

alexander-novo: At least for properties the naming convention of putting `_` at the end addresses this.

lukelowry: Addressed in [919c1b1](https://github.com/ORNL/GridKit/pull/488/commits/919c1b1256956204938888ae951d6e28ceae9993), thank you for the suggestion that is much more readable.

GridKit/Model/PowerElectronics/CircuitComponent.hpp
```diff
@@ -96,11 +99,12 @@ namespace GridKit
...
-      y_.resize(static_cast<size_t>(size_));
-      yp_.resize(static_cast<size_t>(size_));
-      abs_tol_.resize(static_cast<size_t>(size_));
-      f_.resize(static_cast<size_t>(size_));
+      if (!allocated_)
+      {
+        allocateVectors(size_);
+      }
 
+      allocated_ = true;
```
pelesh, line 107: I am not sure base class should allocate. This is inherited from earlier code, so probably should addressed in a separate issue.

pelesh (approved):
Changes look consistent and tests pass.

I'd appreciate one more approval before we merge.

lukelowry:
Great! We get this merged then I will send the team the knock-out lsit of PRs in order of importance and its just a countdown until we have all the models and cases ready for release

examples/PowerElectronics/Microgrid/Microgrid.cpp
```diff
@@ -229,15 +236,16 @@ int main(int /* argc */, char const** /* argv */)
 
   idas->runSimulation(t_final);
 
-  std::vector<double>& yfinial = sysmodel->y();
+  auto&       yfinial      = sysmodel->y();
+  const auto* yfinial_data = yfinial.getData();
```
superwhiskers: this looks like a misspelling. perhaps it should be `yfinal`?

examples/PowerElectronics/RLCircuit/RLCircuit.cpp
```diff
@@ -104,12 +111,13 @@ int main(int /* argc */, char const** /* argv */)
 
   idas.runSimulation(t_final);
 
-  std::vector<double>& yfinial = sysmodel.y();
+  auto&       yfinial      = sysmodel.y();
+  const auto* yfinial_data = yfinial.getData();
```
superwhiskers: likewise here

superwhiskers (approved):
this change looks pretty straightforward. just two naming comments.

as a larger suggestion, i think we could have avoided the large diff volume here by overloading the indexing operator on the `Vector` type. is there any reason we didn't just do that?

examples/PowerElectronics/RLCircuit/RLCircuit.cpp
lukelowry: Addressed in [bf4d7a7](https://github.com/ORNL/GridKit/pull/488/commits/bf4d7a77e13333549a25c3bd37598a92e0544057), I think this was preexisting, but good catch

examples/PowerElectronics/Microgrid/Microgrid.cpp
lukelowry: Addressed in [bf4d7a7](https://github.com/ORNL/GridKit/pull/488/commits/bf4d7a77e13333549a25c3bd37598a92e0544057)

lukelowry:
> this change looks pretty straightforward. just two naming comments.
> 
> as a larger suggestion, i think we could have avoided the large diff volume here by overloading the indexing operator on the `Vector` type. is there any reason we didn't just do that?

I did that at first, and it was much smaller. the issue that @pelesh raised originally in an earlier verison that this made `HOST` / `DEVICE` ambiguous but looking at the current state of PR I am not sure if that concern is still relevant. unsure, thanks for raising

superwhiskers:
the usual case seems to be that you want to write to host data here. defaulting to that seems fine as long as it's documented.

if that's insufficient, there's probably another way to design it so that it's clearer or so that you may choose. the ergonomic benefits would make it worthwhile, i think.

pelesh:
> the usual case seems to be that you want to write to host data here. defaulting to that seems fine as long as it's documented.
> 

Actually, residual and Jacobian should be evaluated on device, so where we are (hopefully) heading is component models with kernels that update data in raw arrays.

PR #489, opened 2026-07-13
Add LinearSolver interface

alexander-novo:
## Description
 
Add new `LinearSolver` interface and one implementation using Re::Solve to be used by e.g. Rosenbrock (#479)
 
 ## Proposed changes
 
Interface has three lifecycle functions:

- `configureSolver` called whenever the matrix changes sparsity pattern
- `setupSolver` called whenever the matrix changes data
- `solve` to perform a linear solve
 
 ## Checklist
- [x] All tests pass.
 
Please see Rosenbrock tests for unit tests.

 ## Further comments


GridKit/LinearAlgebra/Solver/ResolveSystemSolver.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    template <class ScalarT, typename IdxT>
+    int ResolveSystemSolver<ScalarT, IdxT>::setupSolver(bool reuse_factors)
+    {
+      if (reuse_factors)
+      {
+        if (int err = lin_solver_.refactorize())
+          return err;
+      }
```
pelesh, line 69: I believe there should be a third option `reuse_pivot_sequence`. Then, this conditional would look something like this:
```suggestion
      if (reuse_factors)
      {
        return 0;
      }
      else if (reuse_pivot_sequence)
      {
        if (int err = lin_solver_.refactorize())
          return err;
      }
```
Might be a good idea to have an enum instead of multiple boolean flags.

GridKit/LinearAlgebra/Solver/ResolveSystemSolver.cpp
```diff
@@ -0,0 +1,105 @@
...
+      ReSolve::vector::Vector resolve_lhs(lhs.getSize());
+
+      if (int err = resolve_rhs.setData(rhs.getData(), memspace_))
+        return err;
+
+      if (int err = resolve_lhs.setData(lhs.getData(), memspace_))
+        return err;
+
+      if (int err = lin_solver_.solve(&resolve_rhs, &resolve_lhs))
+        return err;
```
pelesh, line 98: I really don't like `setData` methods as they are ticking bomb but we still carry them along. How about we create a view constructor to make a shallow copy of the vector:

```suggestion
      ReSolve::vector::Vector resolve_rhs(rhs.getSize(), rhs.getData());
      ReSolve::vector::Vector resolve_lhs(lhs.getSize(), lhs.getData());

      if (int err = lin_solver_.solve(&resolve_rhs, &resolve_lhs))
        return err;
```
Even better would be if that constructor is made private and we implement method `createVectorView` that would invoke such constructor and create vector object that does not own the data.

pelesh (commented):
Brief comments on interfacing with Re::Solve

GridKit/LinearAlgebra/Solver/ResolveSystemSolver.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return ReSolve::memory::DEVICE;
+      default:
+        throw "Memory space not supported";
+      }
+    }
+
+    template <class ScalarT, typename IdxT>
+    ResolveSystemSolver<ScalarT, IdxT>::ResolveSystemSolver(ReSolve::SystemSolver& lin_solver, GridKit::memory::MemorySpace memspace)
+      : lin_solver_(lin_solver), memspace_(memorySpaceAsResolve(memspace))
+    {
```
pelesh, line 30: You really don't need this. The `SystemSolver` class will manage host/device memory spaces under the hood.
```suggestion
      : lin_solver_(lin_solver)
    {
```
In the PR the `memspace_` is used only to set vector/matrix views. I would just mark those places as `@todo` to remove them when we improve linear algebra support in GridKit. Better to merge this PR now and address these issues in the next PR.

GridKit/Solver/Dynamic/Rosenbrock.hpp
```diff
@@ -0,0 +1,517 @@
...
+     * @brief A Rosenbrock integrator designed to simulate a `GridKit::Model::Evaluator` model.
+     *        Rosenbrock integrators are best for quick medium-accuracy simulation and
+     *        boast a large amount of customization to simulating a given model.
+     *
+     *        For the list of available Rosenbrock methods, see `Rosenbrock::Tableau`.
+     */
+    template <class ScalarT, typename IdxT>
+    class Rosenbrock
+    {
+      using RealT = typename GridKit::ScalarTraits<ScalarT>::RealT;
```
pelesh, line 34: `Rosenbrock` class should inherit from `DynamicSolver`.

README.md
```diff
@@ -28,6 +28,7 @@ You should have all of the following installed before installing GridKit™
       that the use of Enzyme is experimental, and some versions of it have been found to break GridKit code.
         - [LLVM](https://github.com/llvm/llvm-project) >= 15.x. GridKit is
           currently tested with LLVM 16.
+    - [ReSolve](https://github.com/ORNL/ReSolve) == commit `a93fb6571e3542b2fa779983fef0186a034c53c9` (optional)
```
pelesh: We are about to release v0.99.3 so this is probably better dependency requirement.
```suggestion
    - [ReSolve](https://github.com/ORNL/ReSolve) >= 0.99.3 (optional)
```

pelesh (approved):
This is good to merge. I suggest two minor changes before we merge it:
- Have `Rosenbrock` class inherit from `DynamicSolver` class.
- Rebase to the current `develop`.

PR #493, opened 2026-07-16
Shared Memory for SystemModel and its Components

lukelowry:
## Description
 
 Zero copy/scatter using new `Vector` for PhasorDynamics for mroe scalable implementation and simulation
 
## Monitoring enabled

| Case | `develop` (s) | This branch (s) | Speedup |
| --- | ---: | ---: | ---: |
| Hawaii | 0.196874 | 0.142129 | 27.8% |
| NewEngland | 0.251645 | 0.172580 | 31.4% |
| Illinois | 0.476614 | 0.296092 | 37.9% |
| WECC | 0.838516 | 0.532921 | 36.4% |
| Texas | 10.311555 | 5.333159 | 48.3% |

## Monitoring disabled

| Case | `develop` (s) | This branch (s) | Speedup |
| --- | ---: | ---: | ---: |
| Hawaii | 0.153863 | 0.113780 | 26.1% |
| NewEngland | 0.206425 | 0.138533 | 32.9% |
| Illinois | 0.217926 | 0.145467 | 33.2% |
| WECC | 0.393717 | 0.260323 | 33.9% |
| Texas | 6.920282 | 3.623507 | 47.6% |

 
 ## Proposed changes
 
- Zero copy for component models in SystemModel
 
 ## Checklist
 
- [x] All tests pass.
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [x] There are unit tests for the new code.
- (N/A)The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- (N/A) I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
The benefits are massive for larger cases!

Issue #495, opened 2026-07-17
Testing issue: `PhasorDynamicsSystemTest` but returns error and warning messages


pelesh:
### Failed test outputs or steps to reproduce the failure(s)

`PhasorDynamicsSystemTest` reports all tests passing however returns multiple error and warning messages.
```console
$ 62: Test command: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics/test_phasor_system
62: Working Directory: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics
62: Test timeout computed to be: 1500
62: --- PASS: Test constructor
62: --- PASS: Test composer
62: --- PASS: Test reallocateAfterTopologyChange
62: --- PASS: Test modelVectorsAliasSystemStorage
62: [WARNING] A component has returned a nullptr Jacobian.
62: [WARNING] A bus has returned a nullptr Jacobian.
62: [WARNING] A component has returned a nullptr Jacobian.
62: [WARNING] A bus has returned a nullptr Jacobian.
62: 0th residual: 15 dependencies: [ (0, -0.2) (1, 0.4) ]
62: 1th residual: -20 dependencies: [ (0, -0.4) (1, -0.2) ]
62: Sparse Csr Matrix: System Jacobian
62: 0 0 -2.000000000000000e-01
62: 0 1 4.000000000000000e-01
62: 1 0 -4.000000000000000e-01
62: 1 1 -2.000000000000000e-01
62: --- PASS: Test jacobian
62: [ERROR] Vector::setData - cannot replace owned host data
62: [ERROR] Vector::setData - cannot replace owned host data
62: [ERROR] Vector::setData - cannot replace owned host data
62: [ERROR] Vector::setData - cannot replace owned host data
62: [ERROR] BusBase::bind - failed to bind vectors to system storage
62: [ERROR] Failed to bind bus vectors to system storage
62: --- PASS: Test allocationError
62: [ERROR] Genrou: pmech signal attached with no linked governor
62: [ERROR] Genrou: efd signal attached with no linked exciter
62: [ERROR] Genrou: pmech signal attached with no linked governor
62: [ERROR] Genrou: efd signal attached with no linked exciter
62: [ERROR] Component errors: 4
```

It seems there are multiple issues here:
- Error messages are probably triggered deliberately. There needs to be message before that stating the test will produce those error messages.
- It seems something is incorrect in the Jacobian test. Having "-20" dependencies doesn't seem right.

[...]

PR #496, opened 2026-07-17
Update Rosenbrock test and related CI test

nkoukpaizan:
## Description
This updates the Rosenbrock test with recent Vector type changes and adds the ability to run the tests in CI.
 
 ## Proposed changes
 - Added Re::Solve Spack dependency to test Rosenbrock solvers in CI.
 - Bug fix: #489 missed the model vector type updates from #493
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
+++ b/tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
@@ -33,7 +33,8 @@ namespace GridKit
     class TrigonometricDaeEvaluator : public Model::Evaluator<ScalarT, IdxT>
     {
     public:
-      using RealT = typename Model::Evaluator<ScalarT, IdxT>::RealT;
+      using RealT   = typename Model::Evaluator<ScalarT, IdxT>::RealT;
+      using VectorT = typename Model::Evaluator<ScalarT, IdxT>::VectorT;
 
       constexpr static size_t SIZE = 2;
 
@@ -44,6 +45,10 @@ namespace GridKit
       int allocate() override
       {
         constexpr size_t NNZ = SIZE * SIZE;
+
+        y_.resize(SIZE);
+        yp_.resize(SIZE);
+        f_.resize(SIZE);
 
         IdxT*  row_ptrs = new IdxT[SIZE + 1];
         IdxT*  cols     = new IdxT[NNZ];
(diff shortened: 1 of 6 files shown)


tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
```diff
@@ -69,13 +74,24 @@ namespace GridKit
...
         tag_ = {true, false};
 
-        f_ = {0.0, 0.0};
-        g_ = {0.0, 0.0};
+        f[0] = 0.0;
+        f[1] = 0.0;
+
+        y_.setDataUpdated();
+        yp_.setDataUpdated();
+        f_.setDataUpdated();
```
alexander-novo, line 93: I'm not 100% sure when these functions need to be called... do they also need to be called in `ResolveSystemSolver::solve()` as well?

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
```diff
@@ -139,22 +160,24 @@ namespace GridKit
       {
         RealT* vals = csr_jac_->getValues();
 
-        double y1 = y_[0];
-        double y2 = y_[1];
+        const auto* y = y_.getData();
+
+        ScalarT y1 = y[0];
+        ScalarT y2 = y[1];
```
alexander-novo, line 166:
```suggestion
        RealT y1 = static_cast<RealT>(y[0]);
        RealT y2 = static_cast<RealT>(y[1]);
```

We should probably do all of the math here in `RealT` instead of `ScalarT`

pelesh (approved):
Works well. 

There are some error messages returned by the `PhasorDynamicsSystemTest` but this is handled in #495.

alexander-novo (approved):

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
nkoukpaizan: Yes, I think it should be called after solving as well.

tests/UnitTests/Solver/Dynamic/RosenbrockTests.hpp
nkoukpaizan: Doing the math in `ScalarT` is the intended choice here, because if `ScalarT=DependencyTracking::Variable`, we want to track the operations. Casting is only needed to store `RealT` into the matrix. It doesn't really matter here because `ScalarT=RealT=double`.

alexander-novo: Since everything ends up being cast to `RealT`, the operations being tracked will be discarded.

nkoukpaizan: That is true. The hope is that by being intentional about when we do the casting, someone else can read the code later and understand the purpose (of `ScalarT` and of the casting).

PR #505, opened 2026-07-22
Update all template formatting

### LINKED ISSUE #366: Use different style for template parameters and public interface types
The contributor guidelines state that template parameter names should use the format `ScalarT`, but this is the same as the public interface types defined in classes. This presents a problem in that the set of public interface types is incomplete (or must be made inconsistent in order to have access to template parameters).

For example, with types in the `Evaluator` tree, `ScalarT` and `IdxT` are template parameters and accessible only within the type. But `RealT` is available as part of the public interface. So, say I have another class that takes a model type as a template parameter (expecting it to inherit from `Evaluator`), I can access `RealT` but not `ScalarT` or `IdxT`...

```c++
template <typename ModelT>
class Foo
{
  // this is fine
  using RealT = typename ModelT::RealT;

  // this is an error; IdxT isn't part of the interface (but intuitively it should be)
  using IdxT = typename ModelT::IdxT;
};
```

You can see the problem looking at the structs in the enzyme work (`DfDy` shouldn't need its last two template parameters).

One can work around this problem by using a traits type (although one would have to break the rules for that type), but I think we could solve the problem with a simple set of rules.

1. Use a different style for template type parameters (say... `TScalar` ... doesn't matter to me)
2. Use the current standard for class member type aliases
3. Always define aliases for template parameters and use the alias everywhere instead of the template parameter.

For example:
```c++
template <typename TScalar, typename TIdx>
class Evaluator
{
public:
  using ScalarT = TScalar;
  using IdxT = TIdx;
  //...
};
```

PhilipFackler:
## Description
 
Update templates everywhere to use new style guidelines
 
 _Closes #366_

 ## Checklist

- [X] All tests pass.

### DIFF
--- a/GridKit/LinearAlgebra/Solver/LinearSolver.hpp
+++ b/GridKit/LinearAlgebra/Solver/LinearSolver.hpp
@@ -14,11 +14,13 @@ namespace GridKit
      * @brief An interface for linear solvers to be used in GridKit, such as in \ref Integrator::Rosenbrock.
      *
      */
-    template <class ScalarT, typename IdxT>
+    template <class scalar_type, typename index_type>
     class LinearSolver
     {
     public:
-      using RealT = GridKit::ScalarTraits<ScalarT>::RealT;
+      using ScalarT = scalar_type;
+      using IdxT    = index_type;
+      using RealT   = GridKit::ScalarTraits<ScalarT>::RealT;
 
       /**
        * @brief Configure the solver by analyzing the given matrix. The sparsity pattern of the given matrix must be set, and future calls
@@ -27,7 +29,8 @@ namespace GridKit
        * @param matrix The matrix to configure the solver with.
        * @return int An error code, or 0 if none.
        */
-      virtual int configureSolver(GridKit::LinearAlgebra::CsrMatrix<RealT, IdxT>& matrix)                                       = 0;
+      virtual int configureSolver(GridKit::LinearAlgebra::CsrMatrix<RealT, IdxT>& matrix) = 0;
+
       /**
        * @brief Setup the solver with matrix data.
        *
@@ -36,7 +39,8 @@ namespace GridKit
        * @pre If `reuse_factors` is true, then `setupSolver` must have been called once before with the same sparsity pattern.
        * @return int An error code, or 0 if none.
        */
-      virtual int setupSolver(bool reuse_factors = false)                                                                       = 0;
+      virtual int setupSolver(bool reuse_factors = false) = 0;
+
       /**
        * @brief Perform a linear solve using the configured matrix.
        *
--- a/GridKit/LinearAlgebra/SparseMatrix/CsrMatrix.cpp
+++ b/GridKit/LinearAlgebra/SparseMatrix/CsrMatrix.cpp
@@ -681,7 +681,7 @@ namespace GridKit
       return copyDataFrom(row_data, col_data, val_data, memspace_in, memspace_out);
     }
 
-    template <typename real_type, typename IdxT>
+    template <typename real_type, typename index_type>
     int CsrMatrix<real_type, IdxT>::allocateMatrixData(memory::MemorySpace memspace)
     {
       IdxT nnz_current = nnz_;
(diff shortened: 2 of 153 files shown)


GridKit/LinearAlgebra/Solver/LinearSolver.hpp
```diff
@@ -14,11 +14,13 @@ namespace GridKit
...
      */
-    template <class ScalarT, typename IdxT>
+    template <class scalar_type, typename index_type>
     class LinearSolver
     {
     public:
-      using RealT = GridKit::ScalarTraits<ScalarT>::RealT;
+      using ScalarT = scalar_type;
+      using IdxT    = index_type;
+      using RealT   = GridKit::ScalarTraits<ScalarT>::RealT;
```
pelesh, line 23: Why are these public? Shouldn't these be private as they are intended for use inside the class?

GridKit/LinearAlgebra/SparseMatrix/CsrMatrix.cpp
```diff
@@ -668,21 +668,21 @@ namespace GridKit
...
     {
       destroyMatrixData(memspace_out);
       nnz_ = new_nnz;
       return copyDataFrom(row_data, col_data, val_data, memspace_in, memspace_out);
     }
 
-    template <typename RealT, typename IdxT>
-    int CsrMatrix<RealT, IdxT>::allocateMatrixData(memory::MemorySpace memspace)
+    template <typename real_type, typename IdxT>
+    int CsrMatrix<real_type, IdxT>::allocateMatrixData(memory::MemorySpace memspace)
```
pelesh: I believe this was an omission:
```suggestion
    template <typename real_type, typename index_type>
    int CsrMatrix<real_type, IdxT>::allocateMatrixData(memory::MemorySpace memspace)
```

GridKit/MemoryUtilities/MemoryUtils.hpp
```diff
@@ -26,14 +26,16 @@ namespace GridKit
...
    *
    * @author Slaven Peles <peless@ornl.gov>
    */
-  template <class Policy>
+  template <typename policy>
   class MemoryUtils
   {
   public:
+    using Policy = policy;
+
```
pelesh, line 38: Why this change?

GridKit/Model/PhasorDynamics/ComponentData.hpp
```diff
@@ -19,23 +19,34 @@ namespace GridKit
...
+      using Parameters           = parameters_type;
+      /// Buses enum
+      using Buses                = buses_type;
+      /// Signal inputs enum
+      using SignalInputs         = signal_inputs_type;
+      /// Signal outputs enum
+      using SignalOutputs        = signal_outputs_type;
+      /// Monitorable variables enum
+      using MonitorableVariables = monitorable_variables_type;
+
```
pelesh, line 49: Why is this necessary? I thought the issue was with scalar types?

pelesh (commented):
Some preliminary comments:
- This PR changes the scope of scalar type aliases from private to public. This leads to an unintuitive code like `void function<scalar_type>(ScalarT x);`.
- There is minor scope creep as type aliases other than scalar types were handled in a similar way. What is the motivation for that?

GridKit/LinearAlgebra/Solver/LinearSolver.hpp
PhilipFackler: The intent is to make a consistent use of interface type aliases across all classes. Unless there is a good reason to hide these names from consumer contexts, the interface should be the same.

GridKit/LinearAlgebra/SparseMatrix/CsrMatrix.cpp
PhilipFackler: Good catch!

GridKit/MemoryUtilities/MemoryUtils.hpp
PhilipFackler: The issue is about consistent template parameter formatting. This is to follow the new guidelines.

GridKit/Model/PhasorDynamics/ComponentData.hpp
PhilipFackler: This issue is about consistent template parameter format across all templates. `ScalarT` was used as an example, but the reasoning applies everywhere. Consumer contexts often need access to type aliases defined by the classes being used. It is rare that type aliases need to be restricted to be private to a class implementation.

PR #510, opened 2026-07-28
Add max steps options for phasor dynamics app

Steven-Roberts:
## Description
 
I'm hitting the default limit of 500 steps for some fixed step experiments, so adds an option to change the value.
 
 ## Checklist
 
- [x] All tests pass.

### DIFF
--- a/GridKit/Solver/Dynamic/Ida.cpp
+++ b/GridKit/Solver/Dynamic/Ida.cpp
@@ -1051,7 +1051,9 @@ namespace AnalysisManager
     {
       if (retval < 0)
       {
-        std::cerr << "\nERROR: Function " << functionName << " failed with flag " << retval << "!\n\n";
+        char *msg = IDAGetReturnFlagName(retval);
+        std::cerr << "\nERROR: Function " << functionName << " failed with flag " << msg << "!\n\n";
+        std::free(msg);
         throw SundialsException();
       }
     }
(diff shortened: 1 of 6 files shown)


lukelowry:
This is great I have ran into this too. Also in the future we should have an option to just run to completion since it might not be obvious how many steps it will take

Steven-Roberts:
> This is great I have ran into this too. Also in the future we should have an option to just run to completion since it might not be obvious how many steps it will take

Setting `max_steps` to a negative number will do this.

lukelowry:
> > This is great I have ran into this too. Also in the future we should have an option to just run to completion since it might not be obvious how many steps it will take
> 
> 
> 
> Setting `max_steps` to a negative number will do this.

Problem solved, ha! Thank you.

lukelowry (approved):

GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -1051,7 +1051,9 @@ namespace AnalysisManager
     {
       if (retval < 0)
       {
-        std::cerr << "\nERROR: Function " << functionName << " failed with flag " << retval << "!\n\n";
+        char* msg = IDAGetReturnFlagName(retval);
+        std::cerr << "\nERROR: Function " << functionName << " failed with flag " << msg << "!\n\n";
```
pelesh: Please use `Logger::error()` instead of `std::cerr`. This may require including appropriate header and linking with Logger.

pelesh (approved):
Looks good!

Please replace `std::cerr` with GridKit's logger.

PR #511, opened 2026-07-30
Collapse enzyme wb, ws, h

lukelowry:
## Description

Refactors PhasorDynamics components to use a consistent internal/external variable and residual interface (thank you for the inspo @alexander-novo). This removes the separate `wb`, `ws`, and `h` paths, simplifies Enzyme Jacobian generation, and makes residual ownership clearer without changing model behavior.

## Proposed changes

- Add shared external variable and residual storage to `Component`.
- Split residual evaluation into internal equations and external contributions.
- Consolidate the specialized Enzyme derivative wrappers into generic internal/external forms.
- Update the affected PhasorDynamics models and Jacobian assembly.
- Ignore invalid row and column mappings during sparse Jacobian storage.

 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments
 
This demonstrates that we don't need to give special treatment to `Bus`, and we can have `Bus` inherit from `Component`, which dramatically simplifies `SystemModel` and signal orchestration. Personally, this is much easier to document and implement. MOST IMPORTANTLY: this internal/external formulation can allow us to make the residual evaluation order independent, which (I think) is better for partitioning and parallelization.

cc @nkoukpaizan  

Draft just so others can see, I may need to break this up


lukelowry:
@Steven-Roberts this representation might make partitioning easier aswell

nkoukpaizan:
Thanks @lukelowry for looking at this. Updating these local vector is on our todo list, and this should work, now that we are storing the current injections as local variables. I would still prioritize merging the models you need for the 10K over this change.

lukelowry:
> Thanks @lukelowry for looking at this. Updating these local vector is on our todo list, and this should work, now that we are storing the current injections as local variables. I would still prioritize merging the models you need for the 10K over this change.
> 
> 

I agree, definitely

lukelowry:
@pelesh I believe this formulation of the system gives me the most flexibility in EMT models. On my end, I would be okay with splitting off from Phasor Dynamics after this.

Issue #466, opened 2026-06-25
PowerFlow-PhasorDynamics Interface


lukelowry:
### Summary

Tracking the steady state interface feature to solve the power flow

### Rationale

Dependent on external solvers for PhasorDynamics until we can obtain a power flow solution from PowerFlow

### Description

The translation of branches and generators is unproblematic. We need to resolve:
- PhasorDynamics data does not include Slack/PV bus labeling
- The LoadZ and LoadZIP models have power-flow dependent parameters

### Additional information

Translation should happen at the `*Data` interface, not JSON

Issue #405, opened 2026-05-18
Init section for device json input


lukelowry:
### Summary

The exciter and governor models desperately need a way to initialize their external. We are initializing these models in an unfortunate way because we don't have a way to set the initial values of external states in the case format:

Some of the more advanced models cannot be implemented due to our current approach of intialziation intra-model.

### Rationale

Right now, during initialization, the `Genrou` model writes information to the governor `Tgov1` via the mechanical power signal. And then (the sequencing matters, which is unfortunate) the governor initialization reads this data that landed in its own internal state:
```code
  // Initial mechanical = initial electric torque
  if (signals_.template isAssigned<Tgov1InternalVariables::PM>())
  {
    p0 = y_[2]; ///<- generator needs to be initialized first
  }

  // Input Variables (Parameter for now)
  pref_ = R_ * p0;
```
which is then used to initialize the static values of `pref` which is an external variable of the governor

But the dependency chain is backward, since we can know *a priori* from the power flow solution the value of `pref`. And the case format itself embeds knowledge of the models, so if the machine `Genrou` isn't wired up to an exciter, then the author of the case format already knows to initialize `pmech` with the value we would have given `pref` if it had a governor.

Additionally, this would allow us to fully unit test models without coupling them with other models as we do with the exciter, governor, and machine trio.

### Description

a init section which only accepts names of external variables
```json
  {
      "class": "Tgov1",
      "ports": {
          "bus": 26,
          "speed": 28,
          "pmech": 29
      },
      "id": "26_1_tgov1",
      "init":{
            "pref": 10.0,
       },
      "params": {
          "R": 0.05,
          "T1": 0.5,
          "T2": 2.5,
          "T3": 7.5,
          "Pvmin": 0.0,
          "Pvmax": 1.0,
          "Dt": 0.0
      }
  },
````

### Additional information

I had a conversation a while ago about this with @abirchfield, but I hit a wall with the `REECA` exciter  implementation for IBR because of this issue

lukelowry:
This might be defensible if we had a built-in power flow solver, but since we do not, unfortunately, this is not better.

pelesh:
We do have a [power flow solver](https://github.com/ORNL/GridKit/tree/develop/GridKit/Model/PowerFlow), all we need to do is get a parser for it.

pelesh:
I suggest we keep this open for now just to have it on our to-do.

lukelowry:
[...]

PR #514, opened 2026-07-31
Design for PowerFlow and PhasorDynamics Interoperability

lukelowry:
## Description

Partially addresses #466, #405, and allow us to close related PR #391

Introduce `.state.json` as a shared initialization/state artifact for PowerFlow and PhasorDynamics (and compatible with EMT), separating operating state and dispatch from the static `.case.json` model definition. This provides a file-based interoperability boundary without coupling the model implementations. 

The proposed format is documented in [STATE.md](https://github.com/ORNL/GridKit/blob/lukel/state-init-dev/GridKit/Model/PhasorDynamics/STATE.md). Further justification for the design is provided in that document

## Proposed changes

- Document the proposed state-file structure and migration path.
- Store bus voltages and device current injections under `buses`.
- Store device dispatch, connectivity, tap, phase, and other initialization states under `devices`.
- Add a [Hawaii example](https://github.com/ORNL/GridKit/blob/lukel/state-init-dev/examples/PhasorDynamics/Medium/Hawaii/hawaii.state.json) at nominal steady-state dispatch.

If approved, I would like to include a  `*.state.json`  for each case as an initial artifact. I only committed an example for the Hawaii case.


lukelowry:
Using this approach its very easy to exchange states between PowerFlow, PhasorDynamics, and OPF (experimental), see [this downstream branch](https://github.com/ORNL/GridKit/tree/lukel/state-init-2-dev). I can reliably start in steady state after changing the generator dispatch using OPF! And the only shared knowledge between the `OPF/` and `PhasorDynamics/` is the proposed `*.state.json`. This contract would make initialization of EMT  much easier (which is notoriously hard)

lukelowry:
Drafting since not likely to be done before release but is very important for the development of `EMT` and practical use.

PR #490, opened 2026-07-14
m_viz: geo visualization for 4 cases: hawaii, illinois, texas, and wecc

isatkaus:
## Description

Adds `m_viz.ipynb`, an interactive Plotly geo-visualization notebook for MATPOWER `.m` case files (ACOPF or PF solutions). The notebook parses bus, generator, and branch tables into normalized pandas DataFrames, merges geographic coordinates from `.gic` or `.AUX` files, and produces an interactive map showing bus loads, dispatched generation (fuel-colored, sized by MW), and branch loading percentage (viridis color scale).

Supported cases: Hawaii40, Illinois (ACTIVSg200), Texas (ACTIVSg2000), WECC (ACTIVSg10k).

Also adds `setup_env.md` with environment setup instructions and input file reference for all supported cases, and supporting utility files `m_viz_utils.py` and `gridkit_utils.py`.

## Proposed changes

- `uq-usecase/notebooks/m_viz.ipynb` — new notebook with geo visualization for all 4 TAMU synthetic cases
- `uq-usecase/notebooks/setup_env.md` — conda environment setup and input file reference for all supported cases
- `uq-usecase/py-utils/m_viz_utils.py` — MATPOWER `.m` parsing, colocated bus splitting, generator fan-out, geo merge, and Plotly map plotting
- `uq-usecase/py-utils/gridkit_utils.py` — optional GridKit JSON ID augmentation for hover labels

## Checklist

- [ ] All tests pass.
- [ ] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [ ] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [ ] I have updated CHANGELOG.md to reflect the changes in this PR.

## Further comments

This is a Python/notebook contribution; the C++ compile, unit test, and CHANGELOG checklist items are not applicable. The notebook is self-contained and adds visualization utilities for the `uq-usecase` use case without modifying any existing GridKit C++ code.

PR #516, opened 2026-08-01
Add Python visualization notebooks

pelesh:
## Description

Reorganizes the Python geo-visualization notebook and utilities proposed in #490 (m_viz: geo visualization for Hawaii40, Illinois, Texas, WECC) to match GridKit's existing directory conventions, and adds CMake wiring and unit tests for it.

This PR keeps the same notebook and the same visualization functionality, but reorganizes it before merge.

Supersedes #490.

CC @isatkaus

## Proposed changes

- Moves the notebook and Python utilities to `application/notebooks/`, matching the existing `application/PhasorDynamics` convention.
- Splits `m_viz_utils.py` (1220 lines) into `application/notebooks/m_viz/{parsing,geo,plotting}.py`.
- Splits `gridkit_utils.py` (641 lines) into:
  - `application/notebooks/gridkit_common/ids.py` — the two functions (`attach_json_ids`, `get_case_path_for_editing`) the notebook actually uses.
  - `application/notebooks/uq_sampling/{sampling,runs,dispatch}.py` — the UQ parameter-sampling, run-orchestration, and Genrou dispatch-patching code that was bundled in but unused by this notebook. Kept in this PR, split by concern, and documented as not yet wired to a notebook in this repo.
- Adds `application/notebooks/pyproject.toml`, making the code an installable package (`pip install -e .`) and removing the notebook's `sys.path.insert` hack.
- Adds `tests/UnitTests/Python/` with 18 pytest unit tests for `m_viz.parsing` and `m_viz.geo` (previously untested).
- Adds CMake wiring: a new `GridKit_ENABLE_NOTEBOOKS` option (alongside the existing `GridKit_ENABLE_*` flags) gates `add_subdirectory(application/notebooks)`; a `notebooks_pyenv` custom target creates a venv and installs the package; the ctest Python suite is gated the same way and runs against that venv, mirroring the existing `test_install` before/after build ordering. Documented in `INSTALL.md`.
- Replaces the hardcoded personal paths in the notebook with environment-variable-driven configuration (`GRIDKIT_CASE_DATA_DIR`, `GRIDKIT_REPO`, `GRIDKIT_FIGS_DIR`), consistent with the notebook's existing `NREL_CLUSTER`/`onkestrel` pattern.
- Drops `_build_branch_line_arrays` from the plotting module — confirmed dead code, never called from `plot_grid` or anywhere else.
- No changes to the notebook's visualization output or behavior — this is a reorganization, not a rewrite.

## Checklist

- [x] All tests pass. (18/18 pytest tests pass; verified end-to-end via `ctest` with `-DGridKit_ENABLE_NOTEBOOKS=ON`)
- [ ] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`. (N/A — Python-only change, no C++ added or modified)
- [x] The feature branch is rebased with respect to the target branch. (branched directly from `develop`)
- [ ] I have updated CHANGELOG.md to reflect the changes in this PR.

## Further comments

This branch has two commits: the first brings in #490's content unmodified as a baseline, the second applies the reorganization — so the diff between them documents exactly what moved and what changed. See #490's own description for the visualization feature itself; nothing about that feature changed here.


lukelowry:
We can set this up so that the output of the notebook renders in ReadTheDocs, FYI. Out of scope but worth mentioning

lukelowry:
I can run this, seems to work. I would ask that we wait until the remaining models are merged and I update the cases so this doesn't unintentionally become dead weight. 

Also, I have all of the GIS coordinates of the cases that I plan on adding to the `extension` section of `Bus` so that we only have to parse our own format. Otherwise, this is useful and cool!

pelesh:
> I can run this, seems to work. I would ask that we wait until the remaining models are merged and I update the cases so this doesn't unintentionally become dead weight.
> 
> Also, I have all of the GIS coordinates of the cases that I plan on adding to the `extension` section of `Bus` so that we only have to parse our own format. Otherwise, this is useful and cool!

Let's get release 0.2 out and then merge GIS and the notebooks. Thanks @isatkaus and @lukelowry

PR #519, opened 2026-08-03
CommonMath Polished Docs

### LINKED ISSUE #513: Fix CommonMath documentation
### Summary
Please rigorously examine the 
[Math documentation](https://gridkit.readthedocs.io/en/latest/GridKit/CommonMath.html)
Some issues that I noticed:

1. $`\sigma`$ and "the sigmoid function" refer to the same thing, however the mathematical definition is for the Heaviside. I believe this was the intention. I would define $`\sigma`$ as the approximation and the Heaviside function as something different.
2. The ramp function should follow the same approximation form as the Heaviside function. Why is one logarithmic and the other tanh?
3. What are the extra lines on the quadratic ramp? There is no approximation there, right? Or are you requiring two derivatives? I think not distinguishing sigmoid from Heaviside is confusing here.
4. Add a legend to distinguish the exact line from the approximate lines in all figures.
5. "Usage" is confusing. I think you mean "used by". But it's not clear to me what those are (acronyms?).
6. [Partially singular coefficients are not supported.](https://gridkit.readthedocs.io/en/latest/GridKit/Model/EMT/Component/Load/LoadZ/README.html#submodel-validation) I think you mean "partial rank matrices". A matrix is either singular, or it is not. Probably easiest to write "only full rank matrices or the 0 matrix are supported"

lukelowry:
## Description
 
 Closes #513 , @shakedregev, please let me know if this resolves your concerns. I like it much better now too.

@pelesh Please read the `Ramp` section; I provide a better explanation of what I was trying to say earlier.
 
 ## Proposed changes
 
Better description soon
 
 ## Checklist
 
 N/A
 
 ## Further comments
 
Do we want the .tex files I used to generate these tracked in the repo?


GridKit/Model/EMT/Component/Load/LoadZ/README.md
```diff
@@ -33,9 +33,8 @@ $\mathbf{z}$ | Impedance | [VectorFit](../../../Operators/Rational/VectorFit/REA
 
 ### Submodel Validation
 
-The current is differential for a nonsingular linear coefficient and algebraic
-when the coefficient is zero. Partially singular coefficients are not
-supported.
+The current is differential when the linear coefficient has full rank and
+algebraic when it is the zero matrix. No other coefficients are supported.
```
pelesh: Simpler wording is better. Coefficient does not have a rank, it is a scalar parameter.
```suggestion
The current is differential variable when coefficient <coefficient name> is nonzero; it is algebraic variable otherwise.
```

pelesh (commented):
I would use the same organization as before, just distinguish between discrete function and smooth approximation.

For example, in the section "Step Function" I would define Heaviside function as used in most commercial tools today and after that definition I would explain we are using sigmoid function to approximate the step function. Minimal changes to your original documentation.

Describing all nonsmooth functions in one section and their smooth counterparts in the other section makes documentation harder to read.

GridKit/Model/EMT/Component/Load/LoadZ/README.md
lukelowry: It is not a scalar parameter. I simplified the wording here [5a7d25c](https://github.com/ORNL/GridKit/pull/519/commits/5a7d25ca4d9b748bc8b43422f61c81d2e7403ddc).

lukelowry:
Corrected, I think it is good to go

GridKit/CommonMath.md
```diff
@@ -2,76 +2,126 @@
...
+- [Derived Functions](#derived-functions)
+
 ## Primitives
 
-The scale $\mu=4\cdot f_{\text{sync}}=240$ is chosen so $\sigma$ behaves like a step on inputs while keeping derivatives finite. As $\mu \to \infty$, these functions approach their exact targets.
+| Symbol | Name | API | Description |
+|--------|------|-----|-------------|
+| $\sigma$ | Logistic Function | `sigmoid` | Smooth Heaviside |
+| $\rho$ | Ramp | `ramp` | Smooth ReLU |
+| $q$ | Quadratic Ramp | `qramp` | Smooth ReQU |
```
nkoukpaizan:
```suggestion
| $\sigma$ | Logistic Function | `sigmoid` | Smooth approximation to the Heaviside step function |
| $\rho$ | Ramp | `ramp` | Smooth approximation to the ramp function |
| $q$ | Quadratic Ramp | `qramp` | Smooth approximation to the quadratic ramp function |
```

GridKit/CommonMath.md
```diff
@@ -2,76 +2,126 @@
...
-| `sigmoid` | Step function | `GENSAL`, `GENROU`, `REECA` |
-| `ramp` | Smooth one-sided ramp | `REGCA`, `REECA`, `REPCA` |
-| `qramp` | Exact one-sided quadratic ramp | `IEEET1` |
+The default scale $\mu$ of these functions is chosen such that $\sigma'(0)=f_{\text{sync}}$, where $f_{\text{sync}}$ is the fundamental frequency of the `PhasorDynamics` model. As $\mu\to\infty$, $\rho$ and $q$ converge to $\text{ReLU}$ and $\text{ReQU}$ everywhere, while $\sigma$ converges to $H$ for $x\ne0$.
 
-### $\sigma$ - `sigmoid`
+### Logistic Function
 
-The sigmoid is also known as the logistic function. The equivalent `tanh` form is used for numerical stability because the exponential form can divide by a very large value.
+The Heaviside function $H$ is a unit step, defined:
```
nkoukpaizan:
```suggestion
The Heaviside function $H$ is a unit step, defined as:
```

GridKit/CommonMath.md
```diff
@@ -2,76 +2,126 @@
...
       \begin{cases}
         0 & x\le 0 \\[0pt]
         1 & x\gt 0
-      \end{cases} \\[0pt]
-    &\approx \dfrac{1}{2}\left(1+\tanh\left(\dfrac{\mu x}{2}\right)\right)
+      \end{cases}
+\end{aligned}
+```
+
+The logistic function is evaluated using the equivalent hyperbolic-tangent form below to avoid overflow when evaluating the exponential form directly.
```
nkoukpaizan:
```suggestion
We use the logistic function as a smooth approximation to the Heaviside function. This approximation is evaluated using the equivalent hyperbolic-tangent form below to avoid overflow when evaluating the exponential form directly.
```

GridKit/CommonMath.md
```diff
@@ -2,76 +2,126 @@
...
+
+```math
+\begin{aligned}
+  H(x) &\approx \sigma(x) = \dfrac{1}{2}\left(1+\tanh\left(\dfrac{\mu x}{2}\right)\right)
```
nkoukpaizan:
```suggestion
  \sigma(x) = \dfrac{1}{2}\left(1+\tanh\left(\dfrac{\mu x}{2}\right)\right) &\approx H(x)  
```

nkoukpaizan (commented):
Please see my suggestion for the sigmoid and update the ramp and quadratic ramp functions accordingly. The issue raised in #513 is about using precise mathematical terminology. There are discrete functions and smooth approximations to those function. 

I wouldn't use ReLU and ReQU in this context, as they are terminology used for activation functions.

GridKit/CommonMath.md
lukelowry: accepted suggested change, I like it. thank you

GridKit/CommonMath.md
lukelowry: thank you good catch

lukelowry:
Thank you for suggestions @nkoukpaizan. All applied

lukelowry:
Oh also @nkoukpaizan the models that consume these functions do use these as activation functions primarily, particularly when we use it for rate limiters. Not an important distinction but worth clarifying

nkoukpaizan:
> Oh also @nkoukpaizan the models that consume these functions do use these as activation functions primarily, particularly when we use it for rate limiters. Not an important distinction but worth clarifying

In that case, please add to the documentation. It is currently not clear to me.

lukelowry:
Okay, resolved. I just made a more accurate introductory paragraph describing the general application of these functions thusfar.

GridKit/CommonMath.md
```diff
@@ -1,77 +1,129 @@
...
+    &=
+      \begin{cases}
+        0 & x\le 0 \\[0pt]
+        x & x\gt 0
+      \end{cases}
+\end{aligned}
+```
 
-`ramp` is the softplus approximation to the one-sided ramp. We do not use $x\sigma(x)$ directly because it introduces a negative tail for $x \lt 0$, while softplus stays nonnegative and approaches $\max(x, 0)$ as the smoothing becomes sharp.
+Also known as $\text{softplus}(x)$, the Ramp function is a non-negative approximation of $\text{ReLU}(x)$. We do not use $x\,\sigma(x)$ because it introduces a negative tail for $x \lt 0$, while $\text{softplus}(x)$ stays nonnegative.
```
nkoukpaizan:
```suggestion
We implement the Ramp function as the $\text{softplus}(x)$, non-negative approximation of $\text{ReLU}(x)$. We do not use $x\,\sigma(x)$ because it introduces a negative tail for $x \lt 0$, while $\text{softplus}(x)$ stays nonnegative.
```

GridKit/CommonMath.md
```diff
@@ -1,77 +1,129 @@
...
+    &=
+      \begin{cases}
+        0 & x\le 0 \\[0pt]
+        x^2 & x\gt 0
+      \end{cases}
+\end{aligned}
+```
 
-*Note*: the implementation of the quadratic ramp `q(x)` could be optimized with Enzyme features down the road so that we don't need the smooth approximation.
+This approximation is implemented using the logistic function.
```
nkoukpaizan:
```suggestion
We implement an approximation to $\text{ReQU}$ using the logistic function.
```

nkoukpaizan (commented):
A couple more suggestions. Will wait for @shakedregev to review.

lukelowry:
Thank you @pelesh @shakedregev  @nkoukpaizan for the attention to detail. Implementing these models is basically infeasible without this doc, it is like my Rosetta Stone between the classical representation of these models and our framework

shakedregev (changes requested):
There's still no legend on any of the graphs, and still confusing the smooth approximation and the actual function for the ramps. Some of the functions later like inside and outside should first show the exact equivalence to the sharp function and then the smooth approximation. Like Nicholson, I also do not understand why the linear function has a different smoothing than the others.

lukelowry:
I will try to add a legend once the other models are merged, but we can just describe it with a caption if we get crunched for time.

I'm not certain I follow the rest. `inside`, for example, does show the exact and the smooth approximation:

image

and can you clarify what you mean by `linear function`? if you are referring to `linseg`:

image

I don't understand what you mean by different smoothing.

lukelowry:
I have updated the plots and added the tex files I used

shakedregev:
No, the linear ramp as opposed to Heaviside and quadratic ramp.

GridKit/CommonMath.md
```diff
@@ -1,77 +1,129 @@
...
 
-### $q$ - `qramp`
+Although $\rho(x)$ is real-analytic, the implemented form is an overflow-safe representation of the function:
+
+```math
+\begin{aligned}
+  \rho(x) =
+  \dfrac{x+\lvert x\rvert}{2}
+  + \dfrac{1}{\mu}\ln\left(1+e^{-\mu\lvert x\rvert}\right)
+\end{aligned}
```
shakedregev, line 80: It seems to me that for negative x this is equivalent, but not so for positive x.

shakedregev:
I see now that your index for devices is $`e`$, which looks the same as Euler's constant, which you also use.
This is unlikely to create confusion because context will likely determine which is correct, but it's still better to have something else. You could also just use exp$`(x)`$ instead of $`e^x`$.
PS: $`e`$ also appears as a voltage vector

shakedregev:
[Here and elsewhere](https://gridkit.readthedocs.io/en/latest/GridKit/Model/EMT/Component/Line/LineDistributed/README.html), it seems that the variables have the same non-indexed JSON name. Why is that?

shakedregev:
Remove the comment [here](https://gridkit.readthedocs.io/en/latest/GridKit/Model/EMT/Component/Line/LineDistributed/README.html) about partially singular coefficients. You can say "nonzero yet singular coefficients" if you want to keep it.

lukelowry:
> [Here and elsewhere](https://gridkit.readthedocs.io/en/latest/GridKit/Model/EMT/Component/Line/LineDistributed/README.html), it seems that the variables have the same non-indexed JSON name. Why is that?

The parameters for the subsystem (e.g., `Yc`) are used to for two instances of those subsystems, and so duplicating the data in the input format would be redundant. So they use the same key. Depends on how we implement, I suppose. worth keeping tabs on in #532

lukelowry:
okay I believe `CommonMath` is nearing readiness for merge. Please let me know if there's anything else to adjust/fix. I also adjusted some of the EMT items pointed out.

shakedregev (approved):
Looks good, I would just replace "Discrete" with "Piecewise"

shakedregev (changes requested):
Please address the change I commented about the equivalence of the functions.

GridKit/CommonMath.md
shakedregev: @lukelowry

lukelowry: I amended a more detailed demonstration that I hope is clear. Please let me know

lukelowry:
@shakedregev Addressed both concerns!

shakedregev (approved):
Looks good

PR #520, opened 2026-08-03
Polish + Prepare IEEEST for release

lukelowry:
## Description
 
Added doxygen and made one functional correction to the model (initialization).
 
 ## Proposed changes
 
- Doxygen comments for model and enums
- Initialization fixed to work on signals that are not zero-valued at steady state
- Implicit representation resolves any parameter problems for the purpose of validation until we merge #460
 
 ## Checklist
 
- [x] All tests pass.
- N/A I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
Half of the model is already implicit, so this jsut make it consistent and helps with the ugly parameter validation until a cleaner design is approved.

### DIFF
--- a/GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
@@ -22,19 +22,27 @@ namespace GridKit
     {
       using Log = ::GridKit::Utilities::Logger;
 
+      /**
+       * @brief Construct a sized but unconfigured IEEEST stabilizer.
+       */
       template <typename scalar_type, typename index_type>
       Ieeest<scalar_type, index_type>::Ieeest()
       {
-        size_ = 12;
+        size_ = static_cast<IdxT>(IeeestInternalVariables::MAXIMUM);
       }
 
+      /**
+       * @brief Construct an IEEEST stabilizer from model data.
+       *
+       * @param[in] data Parameters and monitored-variable selections.
+       */
       template <typename scalar_type, typename index_type>
       Ieeest<scalar_type, index_type>::Ieeest(const ModelDataT& data)
         : monitor_(std::make_unique<MonitorT>(data))
       {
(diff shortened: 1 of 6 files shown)


GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
```diff
@@ -238,13 +300,13 @@ namespace GridKit
...
-        tag_[5]  = (T4_ != 0.0);
-        tag_[6]  = (T6_ != 0.0);
+        tag_[0]  = (O1_ > 0.0);
+        tag_[1]  = (O2_ > 0.0);
+        tag_[2]  = (O3_ > 0.0);
+        tag_[3]  = (a4_ > 0.0);
+        tag_[4]  = (T2_ > 0.0);
+        tag_[5]  = (T4_ > 0.0);
+        tag_[6]  = (T6_ > 0.0);
         tag_[7]  = false;
```
pelesh, line 309: I believe we agree that this cannot be the way to configure differential and algebraic variables. This approach will work with implicit DAE formulation but will fail for Hessenberg form.

If we are to pack models of different order in this component, we do need to add separate residual functions for each order. I don't see way around it.

pelesh (changes requested):
I think we need to consider differential variable designation as the compile-time requirement, because it will become so once we introduce Hessenberg form for DAEs.

GridKit/Model/PhasorDynamics/Stabilizer/IEEEST/IeeestImpl.hpp
lukelowry: I agree. I think that the template parameter approach (the other PR) is the cleanest. Think starting with this which shows all the permutations is a good starting point showing the requirements so we can design the compile time version maintainably

pelesh:
Closing in favor of #460

PR #524, opened 2026-08-05
Fix broken documentation references

lukelowry:
## Description

The current Read the Docs build fails during Doxygen generation because `CircuitComponent` contains unresolved `\ref` links to internal data members. Since Doxygen warnings are treated as errors (to ensure we don't accumulate documentation issues), the documentation build stops before Sphinx runs.

This fixes those references and adds a Doxygen build check to the existing pre-commit workflow so these failures are caught before merge. No project dependencies are added.

## Proposed changes

- Replace unresolved member references with inline code formatting.
- Correct the internal and external residual documentation.
- Replace the stale `getSize()` reference with `size()`.
- Build Doxygen in the existing pre-commit workflow.

## Checklist

- [x] Relevant checks pass with `check-yaml` and `doxygen Doxyfile`.
- N/A. No compiled code changed.
- [x] The changes follow GridKit™ style guidelines.
- N/A. No new code requiring unit tests.
- [x] The documentation is updated.
- N/A.  no CHANGELOG needed

## Further comments

The new check only builds the Doxygen XML and does not add a full Sphinx or Read the Docs CI build (which would be preferred but is pending.)

EDIT: Doxygen pre-commit has been deferred for another PR.

### DIFF
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -195,16 +195,16 @@ namespace GridKit
     }
 
     /**
-     * @brief Evaluate all of the residuals of internal variables of the component,
-     * modifying \ref f_.
+     * @brief Evaluate all residuals for the component's internal variables,
+     * writing them through `f_int_`.
      *
      * @return An error code, or 0 if successful.
      */
     virtual int evaluateInternalResidual() = 0;
 
     /**
-     * @brief Evaluate all of the residuals of external variables of the component,
-     * modifying \ref f_
+     * @brief Evaluate all residual contributions for the component's external variables,
+     * accumulating them through `f_ext_`.
      *
      * @return An error code, or 0 if successful.
      */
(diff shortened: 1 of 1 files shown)


.github/workflows/pre_commit.yaml
```diff
@@ -20,3 +20,10 @@ jobs:
         committer_name: GitHub Actions
         committer_email: actions@github.com
         message: Apply pre-commit fixes 
+    - name: Install Doxygen
+      run: |
+        sudo apt-get update
+        sudo apt-get install --yes doxygen
+    - name: Build Doxygen
+      working-directory: docs
```
nkoukpaizan: Out of scope for this PR, but it would be good to build out-of-source. I think the `GridKitDocs` `CMake` target does that, but this won't, and ReadTheDocs currently doesn't.

lukelowry: Okay, worth a shot. I will remove that

lukelowry: Removed in [616a190](https://github.com/ORNL/GridKit/pull/524/commits/616a19078fffe361953df83f34d4bea13ad1dd4e)!

lukelowry:
Okay now its just doxygen comment fixes

nkoukpaizan (approved):

alexander-novo:
I know I'm a bit late to this, but several of the items that references have been removed from just need to have documentation added to them, rather than the references unlinked. Doxygen just can't find them because they haven't been documented.

PR #525, opened 2026-08-05
Model adjustments caught during validation

lukelowry:
## Description
 
These are the last few issues I ran into during validation. Mostly initialization and parameter guard fixes.

Lots of tiny edge cases in the full unmodified cases that we did not account for.
 
 ## Proposed changes
 
- Misc parameter gaurds
- Misc formatting while I was touching files
- Simplified & fixed earlier models that existed before the newer API

 ## Checklist
 
- [x] All tests pass.
- NA I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
Looking very good. I think you all will be very happy with the validation results. Sneak peek:

wecc omega
socar omega

### DIFF
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -341,12 +341,8 @@ namespace GridKit
         const ScalarT vr0  = vfe0;
         const ScalarT vhv0 = vr0 / Ka_;
 
-        ret = vr0 >= Vrmin_ && vr0 <= Vrmax_;
-        if (!ret)
-        {
-          Log::error() << "Esdc1a: initialized VR is outside limits\n";
-          return 1;
-        }
+        const RealT vr0_value       = static_cast<RealT>(vr0);
+        const bool  widen_vr_limits = vr0_value < Vrmin_ || vr0_value > Vrmax_;
 
         // An inactive high-value gate is seeded with the gate input, so the
         // residual reproduces VHV through the same smooth maximum. UEL modes
@@ -369,14 +365,25 @@ namespace GridKit
         const ScalarT vf0   = ScalarT{ZERO<RealT>};
         const ScalarT ev0   = vll0;
         const ScalarT xll0  = ev0;
-        const ScalarT vref0 = ev0 + vc0 + vf0 - vs0 - uel_on_ * vuel0;
+        ScalarT       vref0 = ev0 + vc0 + vf0 - vs0 - uel_on_ * vuel0;
+        const ScalarT ev_residual0 =
+            -ev0 + (vref0 + vs0 + uel_on_ * vuel0 - vc0 - vf0);
+        vref0 -= ev_residual0;
 
         if (automatic_ke)
         {
           Log::warning() << "Esdc1a: Ke is zero and is set so initial VR equals Vrmax/10\n";
         }
+        if (widen_vr_limits)
+        {
+          Log::warning() << "Esdc1a: initial VR is outside [Vrmin, Vrmax] "
+                            "and the limits are widened\n";
+        }
 
-        Ke_     = ke0;
+        Ke_    = ke0;
+        Vrmin_ = std::min(Vrmin_, vr0_value);
+        Vrmax_ = std::max(Vrmax_, vr0_value);
+
         y[EFDP] = efdp0;
         y[VC]   = vc0;
         y[VR]   = vr0;
@@ -793,6 +800,12 @@ namespace GridKit
         check_non_negative(Tr_, "Tr");
         check_non_negative(Ta_, "Ta");
         check_non_negative(Te_, "Te");
+
+        if (Vrmax_ < Vrmin_)
+        {
+          Log::warning() << "Esdc1a: Vrmax is below Vrmin and the limits are swapped\n";
+          std::swap(Vrmin_, Vrmax_);
+        }
 
         if (Tr_ < TIME_CONSTANT_MINIMUM || Ta_ < TIME_CONSTANT_MINIMUM
             || Tb_ < TIME_CONSTANT_MINIMUM || Te_ < TIME_CONSTANT_MINIMUM
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -162,7 +162,6 @@ namespace GridKit
         }
 
         check(Ka_ > ZERO<RealT>, "Ka must be positive");
-        check(Tc_ >= ZERO<RealT>, "Tc must be non-negative");
         check(Vrmin_ <= Vrmax_, "Vrmin must be less than or equal to Vrmax");
         check(UEL_ >= static_cast<IdxT>(0) && UEL_ <= static_cast<IdxT>(3),
               "UEL must be 0, 1, 2, or 3");
@@ -684,23 +683,16 @@ namespace GridKit
           target = parsed_value;
         };
 
-        auto load_switch = [&](auto key, bool& target, const char* name)
+        auto load_switch = [&](auto key, bool& target)
         {
           if (!data.parameters.contains(key))
           {
             return;
           }
 
-          const auto& value = data.parameters.at(key);
-          if (const auto* bool_value = std::get_if<bool>(&value))
-          {
-            target = *bool_value;
-          }
-          else
-          {
-            Log::error() << "Esdc1a: parameter '" << name << "' must be boolean\n";
-            ++parameter_error_count_;
-          }
+          target = std::visit([](auto value)
+                              { return value != 0; },
+                              data.parameters.at(key));
         };
 
         auto load_selector = [&](auto key, IdxT& target, const char* name)
@@ -733,13 +725,13 @@ namespace GridKit
         load_real(Params::Te, Te_, "Te");
         load_real(Params::Kf, Kf_, "Kf");
         load_real(Params::Tf1, Tf1_, "Tf1");
-        load_switch(Params::Spdmlt, Spdmlt_, "Spdmlt");
+        load_switch(Params::Spdmlt, Spdmlt_);
         load_real(Params::E1, E1_, "E1");
         load_real(Params::Se1, Se1_, "Se1");
         load_real(Params::E2, E2_, "E2");
         load_real(Params::Se2, Se2_, "Se2");
         load_selector(Params::UEL, UEL_, "UEL");
-        load_switch(Params::exclim, exclim_, "exclim");
+        load_switch(Params::exclim, exclim_);
         setDerivedParameters();
       }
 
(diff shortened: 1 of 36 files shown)


abirchfield (approved):
These are good cleanup steps that help clarify documentation, increase test coverage, and update a few implementation details.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -452,7 +475,9 @@ namespace GridKit
         }
         if (data.parameters.contains(Parameter::Ispdlim))
         {
-          Ispdlim_ = std::get<RealT>(data.parameters.at(Parameter::Ispdlim));
+          Ispdlim_ = std::visit([](auto value)
+                                { return value != 0 ? ONE<RealT> : ZERO<RealT>; },
+                                data.parameters.at(Parameter::Ispdlim));
```
pelesh, line 496: It is difficult to see what this function does. I suggest rewriting or at least commenting better.

pelesh: The rest of the code below is quite cryptic. It needs to be commented better.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/README.md
```diff
@@ -69,7 +66,19 @@ expected saturation near the operating region:
 \end{aligned}
 ```
 
-Generally, this system has two solutions. The non-extraneous solution is as follows.
+When one saturation value is zero, the normal curve fit uses the corresponding
+voltage as the quadratic knee:
+
```
pelesh: I believe you wanted to say "When only one ...". Also, it is not quite clear to me what the rest of this sentence means.

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -130,6 +131,10 @@ namespace GridKit
       template <typename scalar_type, typename index_type>
       void Tgov1<scalar_type, index_type>::setDerivedParams()
       {
+        if (Pvmax_ < Pvmin_)
+        {
+          std::swap(Pvmin_, Pvmax_);
+        }
```
pelesh, line 138: I don't know what is the purpose of this but I don't see anything good coming from swapping `Pvmax` and `Pvmin`.

GridKit/Model/PhasorDynamics/SynchronousMachine/GenClassical/GenClassicalImpl.hpp
```diff
@@ -368,6 +369,18 @@ namespace GridKit
...
     {
+      if (H_ == ZERO<RealT> && D_ == ZERO<RealT>)
+      {
+        H_inv_ = ZERO<RealT>;
+      }
+      else
+      {
+        H_     = std::max(H_, static_cast<RealT>(0.1));
+        H_inv_ = ONE<RealT> / (TWO<RealT> * H_);
+      }
```
pelesh, line 387: This is in my opinion unnecessary and is a bad modeling practice. Setting `H_inv_` to zero changes the model.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROU/GenrouImpl.hpp
```diff
@@ -697,6 +664,27 @@ namespace GridKit
...
+      if (Xdpp_ > transient_min)
+      {
+        Xdpp_ = static_cast<RealT>(0.8) * transient_min;
+      }
+      Xdpp_ = std::max(Xdpp_, static_cast<RealT>(0.05));
+      if (Xl_ > Xdpp_)
+      {
+        Xl_ = static_cast<RealT>(0.8) * Xdpp_;
+      }
+      Xqpp_ = Xdpp_;
```
pelesh: Changing user input silently is a very bad idea. There has to be justification for that, and this block of code needs comments explaining what is done here and why.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROU/README.md
```diff
@@ -136,62 +136,47 @@ Note that for implementation purposes, some of these equations may be simplified
   0 &= -V_{d} -\psi''_{q}(1+\omega)\\
   0 &= -V_{q}  +\psi''_{d}(1+\omega)\\
   0 &= -T_{elec} +(\psi''_{d} - I_dX_d'')I_q-(\psi''_{q} - I_qX_d'')I_d \\
-  0 &= -k_{sat} + S_B(\psi''-S_A)^2\sigma(\psi''-S_A) \\
+  0 &= -k_{sat} + S_B q(\psi''-S_A) \\
   0 &= -I_d + I_r \sin(\delta) - I_i \cos(\delta) \\
   0 &= -I_q + I_r \cos(\delta) + I_i \sin(\delta) \\
   0 &= -I_r + G (V_d \sin(\delta) + V_q \cos(\delta) - V_r) - B (V_d \cos(\delta) + V_q \sin(\delta) - V_i) \\
   0 &= -I_i + B (V_d \sin(\delta) + V_q \cos(\delta) - V_r) + G (V_d \cos(\delta) + V_q \sin(\delta) - V_i)
```
pelesh: These equations do not render correctly.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENSAL/GensalImpl.hpp
```diff
@@ -520,6 +519,21 @@ namespace GridKit
...
+      }
+      if (Xdpp_ > Xdp_)
+      {
+        Xdpp_ = static_cast<RealT>(0.8) * Xdp_;
+      }
+      Xdpp_ = std::max(Xdpp_, static_cast<RealT>(0.05));
+      if (Xl_ > Xdpp_)
+      {
+        Xl_ = static_cast<RealT>(0.8) * Xdpp_;
+      }
```
pelesh: Again, changing user's input silently is a Bad Idea™.

tests/UnitTests/PhasorDynamics/ExciterIeeet1Tests.hpp
```diff
@@ -23,7 +24,8 @@ namespace GridKit
     class ExciterIeeet1Tests
     {
     public:
-      using RealT = typename PhasorDynamics::Component<ScalarT, IdxT>::RealT;
+      using RealT                   = typename PhasorDynamics::Component<ScalarT, IdxT>::RealT;
+      static constexpr ScalarT tol_ = 100 * std::numeric_limits<ScalarT>::epsilon();
```
pelesh: Why are we using 100 x epsilon tolerance? It would be good to have this value set as global testing requirement rather than set it on a test by test basis.

pelesh (changes requested):
Great work and very helpful expansion of unit tests.

There are some solutions in this PR that I don't quite understand (see comments below) and I would like us to clarify that before we merge.

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
abirchfield: Commercial software often has a practice of trying to determine user intent when bad data is encountered. In this case, the most likely explanation for Pvmax < Pvmin is that the user accidentally switched them. Some users prefer this autocorrection to throwing lots of errors, especially when they don't have detailed familiarity with the models. I'm OK with this or without it.

GridKit/Model/PhasorDynamics/SynchronousMachine/GenClassical/GenClassicalImpl.hpp
abirchfield: Sometimes there is a convention of letting a 0 value represent infinity, for a parameter for which 0 does not make sense. A GENCLS with infinite inertia is just an infinite bus, and this is sometimes used to represent infinite buses.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROU/GenrouImpl.hpp
abirchfield: I think these are the PowerWorld autocorrection rules, trying to fix obviously bad data for a user who may not have the expertise to determine what values should be put for these parameters. I guess there is a general question of whether we should implement autocorrection or not.

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
```diff
@@ -349,6 +370,7 @@ namespace GridKit
         const ScalarT xll0  = ev0;
         const ScalarT vref0 = ev0 + vc0 + vf0 - vs0 - uel_on_ * vuel0;
 
+        Ke_     = ke0;
         y[EFDP] = efdp0;
```
pelesh: Here is another instance of auto-correction. It is not commented and it is unclear why user supplied parameter is changed.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -227,13 +229,34 @@ namespace GridKit
         ScalarT efdp = efd0 / (ONE<RealT> + omega * Ispdlim_);
         ScalarT ksat = SB_ * Math::qramp(efdp - SA_);
         ScalarT ve   = ksat * efdp;
-        ScalarT vr   = Ke_ * efdp + ve;
-        ScalarT vtr  = vr / Ka_;
+
+        RealT ke0 = Ke_;
+        if (ke0 == ZERO<RealT>)
```
pelesh, line 239: Another autocorrection here.

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -228,32 +265,53 @@ namespace GridKit
...
+          Log::error() << "Tgov1: initial signals and states must be finite\n";
+          return 1;
+        }
 
-        // Internal States
-        y[0] = (T3_ - T2_) * p0; // y0 - Ptx (Turbine Power )
-        y[1] = p0;               // y1 - Pv  (Valve Position)
-        y[2] = toSystemBase(p0); // y2 - Pm  (Mech Power, System Base)
+        Pvmin_ = std::min(Pvmin_, static_cast<RealT>(pv0));
+        Pvmax_ = std::max(Pvmax_, static_cast<RealT>(pv0));
```
pelesh: I think this is really dangerous auto-correct. Instead of changing initial value, it changes properties of the governor.

GridKit/Model/PhasorDynamics/SynchronousMachine/GENROU/GenrouImpl.hpp
lukelowry: These are the autocorrection rules from PowerWorld. I will add a logger warning when this autocorrection occurs.

I think this is necessary if we are going to be validating against PowerWorld. If we want to get to solver-tolerance precision against a reference solution, I need to be certain it's not autocorrected parameters causing differences. I also agree with @abirchfield, for his listed reasons

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
lukelowry: This one is actually the intended behavior of the governor. If we solved for the steady-state power flow solution in the equivalent `PowerFlow` or `OPF` model, we would set the constraint $P_{vmin} \leq P_{gen} \leq P_{vmax}$, and this initialization-stage autocorrection would never be invoked in an ideal situation. However, if the models' `PowerFlow` data and `PhasorDynamics` data are different, then it would be possible for the power flow solution to be an infeasible steady-state solution in the transient model.

I'll add a logger warning for this. A lot of the test cases have a power flow solution that invokes this.

Issue #469, opened 2026-06-26
Add a simple co-simulation example with GridKit apps operating as ZMQ client/server.


pelesh:
### Summary

Create a simple example where a 3-bus system operates as a ZMQ client and constant signal source operates as ZMQ server.

CC @alexander-novo @lukelowry 

### Rationale

This example is a prototype for implementing capability for GridKit apps to operate in a co-simulation environment.

### Description

Use 3-bus example in GridKit and modify it to replace one of the generators with a signal adapter. Implement that app as a ZMQ client expecting signal input from a generator implemented as a ZMQ server.

In this initial implementation, use constant signal source as a generator model. The signal source should be implemented as a separate app, a ZMQ server, and should return (constant) values for injection currents on client's request.

### Additional information

Another example, working through [DataBroker](https://github.com/sandialabs/ARCADE) instead of direct connection between the two GridKit apps, should be added, as well.

PR #537, opened 2026-08-13
Add simple one-sided co-sim example taking constant current over zmq

PhilipFackler:
## Description
 
See #469 

Introduce a minimal example using zmq for communication between GridKit processes.
 
 ## Proposed changes
 
This example duplicates the behavior of the `ThreeBusConstantSource` example, but with the `ConstantSignalSource` component removed and its behavior reproduced in the app (external to the system model) with currents being received over zmq from another app.
 
 ## Checklist
 
- [X] All tests pass.
 
 ## Further comments

A lot of comments here. :smile: I have numbered them for reference in discussion. 

1. I verified locally that the solution output is equal to that of the `ConstantSignalSource` example.
2. This example code is temporary; longer-term, this will be generalized and moved to the applications directory.
3. This example uses the "step_callback" in `Ida::runSimulation` rather than managing the stepping itself; this is questionable :), since we are intending to remove that.
4. The app implementation is a bit crude because this proof-of-concept (first attempt) was meant to be non-invasive (no modification to the phasor dynamics libraries). In order to move forward with a more general implementation, we will need some other tools.
    1. The immediate next step would be to add a "branch-terminal-adapter" (not sure what to name it) so that bus voltage (coming in over zmq) could be read from outside the system. This would enable a true co-simulation example with two systems solving in step with one another.
    2. Beyond that, the path diverges depending on how often we want to synchronize: just at every time step OR at every call to `evaluateResidual`
        1. Every residual: Use the "IOPorts" PR replacement of `ComponentSignals`. Extend it to make ports configurable...enabling the zmq exchange to be triggered when signals are read from ports.
        2. Every step
            1. Add a modified version of `ConstantSignalSource` that would use zmq to update its "constant" values.
            2. Use "IOPorts" as-is/was to enable this more naturally from outside the system (no new component needed). This would make the most sense from the application perspective because the application input file could simply specify which signals to assign to the network.

I'll add that in either of the "IOPorts" scenarios the co-simulation "manager" could be set up to exchange all necessary variables in one send/recv.

### DIFF
--- a/GridKit/Model/PhasorDynamics/SystemModelData.cpp
+++ b/GridKit/Model/PhasorDynamics/SystemModelData.cpp
@@ -32,5 +32,18 @@ namespace GridKit
       }
       return parseSystemModelData(stream);
     }
+
+    SystemModelData<double, size_t> parseSystemModelData(const std::string& fileName)
+    {
+      auto stream = std::ifstream(fileName);
+      if (!stream)
+      {
+        std::stringstream ss;
+        ss << "Could not open file: " << fileName;
+        Log::error() << ss.str() << std::endl;
+        throw std::runtime_error(ss.str());
+      }
+      return parseSystemModelData(stream);
+    }
   } // namespace PhasorDynamics
 } // namespace GridKit
--- a/GridKit/Model/PhasorDynamics/SystemModelData.cpp
+++ b/GridKit/Model/PhasorDynamics/SystemModelData.cpp
@@ -32,18 +32,5 @@ namespace GridKit
       }
       return parseSystemModelData(stream);
     }
-
-    SystemModelData<double, size_t> parseSystemModelData(const std::string& fileName)
-    {
-      auto stream = std::ifstream(fileName);
-      if (!stream)
-      {
-        std::stringstream ss;
-        ss << "Could not open file: " << fileName;
-        Log::error() << ss.str() << std::endl;
-        throw std::runtime_error(ss.str());
-      }
-      return parseSystemModelData(stream);
-    }
   } // namespace PhasorDynamics
 } // namespace GridKit
--- a/examples/Network/ThreeBusCoSim/CoSim.hpp
+++ b/examples/Network/ThreeBusCoSim/CoSim.hpp
@@ -4,6 +4,17 @@ namespace GridKit
 {
   namespace CoSim
   {
+    /*
+     * The following is meant to be used to communicate driver information over
+     * the network. It is crude, but it at least gives language in the code that
+     * expresses the meaning of the intent.
+     *
+     * - STEP is used to initiate (and continue) the co-simulation (take a step)
+     * - END is used to end the co-simulation
+     *
+     * Both of these are sent by the client to keep the server in step.
+     */
+
     using StatusT = int;
 
     inline constexpr StatusT STEP = 1;
--- a/examples/Network/ThreeBusCoSim/CoSim.hpp
+++ b/examples/Network/ThreeBusCoSim/CoSim.hpp
@@ -1,0 +1,12 @@
+#pragma once
+
+namespace GridKit
+{
+  namespace CoSim
+  {
+    using StatusT = int;
+
+    inline constexpr StatusT STEP = 1;
+    inline constexpr StatusT END  = 0;
+  } // namespace CoSim
+} // namespace GridKit
--- a/examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
+++ b/examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
@@ -1,0 +1,134 @@
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <GridKit/Model/PhasorDynamics/BusFault/BusFault.hpp>
+#include <GridKit/Model/PhasorDynamics/SignalNode/SignalNode.hpp>
+#include <GridKit/Model/PhasorDynamics/SystemModel.hpp>
+#include <GridKit/Model/PhasorDynamics/SystemModelData.hpp>
+#include <GridKit/Solver/Dynamic/Ida.hpp>
+
+#include "CoSim.hpp"
+#include <zmq.hpp>
+
+using namespace GridKit;
+using namespace GridKit::PhasorDynamics;
+using namespace AnalysisManager::Sundials;
+
+template <typename scalar_type, typename index_type>
+class CoSimClient
+{
+public:
+  using ScalarT      = scalar_type;
+  using IdxT         = index_type;
+  using SystemModelT = SystemModel<ScalarT, IdxT>;
+  using SignalT      = typename SystemModelT::SignalT;
+
+  CoSimClient() = delete;
+
+  CoSimClient(SignalT* vr, SignalT* vi, SignalT* ir, SignalT* ii)
+    : vr_signal_(vr),
+      vi_signal_(vi),
+      ir_signal_(ir),
+      ii_signal_(ii),
+      ctx_{},
+      socket_(ctx_, zmq::socket_type::req)
+  {
+    ir_signal_->set(&ir_, &ir_idx_);
+    ii_signal_->set(&ii_, &ii_idx_);
+    socket_.connect("tcp://0.0.0.0:5556");
+  }
+
+  ~CoSimClient()
+  {
+    std::ostringstream oss;
+    oss << CoSim::END << " " << vr_signal_->read() << " " << vi_signal_->read();
+    zmq::message_t s_msg{oss.str().data(), oss.str().size()};
+    socket_.send(s_msg, zmq::send_flags::none);
+
+    zmq::message_t r_msg;
+    auto           recv_result = socket_.recv(r_msg, zmq::recv_flags::none);
+    if (recv_result)
+    {
+    }
+  }
+
+  void exchange()
+  {
+    // 1. Send data
+    std::ostringstream oss;
+    oss << std::scientific << std::setprecision(16);
+    oss << CoSim::STEP << " " << vr_signal_->read() << " " << vi_signal_->read();
+    // std::cout << "[CLIENT] Sending: " << oss.str() << std::endl;
+
+    zmq::message_t s_msg{oss.str().data(), oss.str().size()};
+    socket_.send(s_msg, zmq::send_flags::none);
+
+    // 2. Receive data from DataBroker
+    zmq::message_t r_msg;
+    auto           recv_result = socket_.recv(r_msg, zmq::recv_flags::none);
+    if (recv_result)
+    {
+      std::istringstream iss(r_msg.to_string());
+      // std::cout << "[CLIENT] Received: " << iss.str() << std::endl;
+      iss >> ir_ >> ii_;
+    }
+  }
+
+private:
+  SignalT*       vr_signal_;
+  SignalT*       vi_signal_;
+  SignalT*       ir_signal_;
+  SignalT*       ii_signal_;
+  ScalarT        ir_{};
+  ScalarT        ii_{};
+  IdxT           ir_idx_{GridKit::INVALID_INDEX<IdxT>};
+  IdxT           ii_idx_{GridKit::INVALID_INDEX<IdxT>};
+  zmq::context_t ctx_;
+  zmq::socket_t  socket_;
+};
+
+using ScalarT = double;
+using RealT   = double;
+using IdxT    = std::size_t;
+
+int main()
+{
+  // Instantiate system
+  auto                       data = parseSystemModelData("ThreeBusCoSimClient.case.json");
+  auto                       sys  = SystemModel<ScalarT, IdxT>(data);
+  CoSimClient<ScalarT, IdxT> client(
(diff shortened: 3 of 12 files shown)


pelesh (commented):
It would be good to document code and add a README file with description of the cosimulation app and the example.

GridKit/Model/PhasorDynamics/SystemModelData.cpp
```diff
@@ -32,18 +32,5 @@ namespace GridKit
...
-      auto stream = std::ifstream(fileName);
-      if (!stream)
-      {
-        std::stringstream ss;
-        ss << "Could not open file: " << fileName;
-        Log::error() << ss.str() << std::endl;
-        throw std::runtime_error(ss.str());
-      }
-      return parseSystemModelData(stream);
-    }
```
pelesh, line 47: Why is this function removed?

PhilipFackler: It caused a compiler error for my app because it's redundant/ambiguous.

pelesh (changes requested):
Please document the code per developer guidelines. Also, please add a README file explaining what the co-sim application does.

If changes to the `SystemModel` class are needed that needs to be tracked in an issue. Modifying `SystemModel` is out of scope of this PR.

GridKit/Model/PhasorDynamics/SystemModelData.cpp
PhilipFackler: I understand the issue here. I will use a workaround in this PR and create an issue for the change.

PhilipFackler: This is resolved in [9a9df2d](https://github.com/ORNL/GridKit/pull/537/commits/9a9df2d046faf057d0a5bbe3b3897b61e03f8d63)

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // Introduce fault and run for the next 0.1s
+  sys.getBusFault(0)->setStatus(true);
+  ida.initializeSimulation(1.0);
+  ida.runSimulation(1.1, dt, step_cb);
+
+  // Clear the fault and run until t = 10s.
+  sys.getBusFault(0)->setStatus(false);
+  ida.initializeSimulation(1.1);
+  ida.runSimulation(10.0, dt, step_cb);
+
```
nkoukpaizan, line 181: A few diagnostic messages to show what the app is doing something at runtime would be helpful. Understood that the commented-out debug statements are too much information. 
- Message for client instantiation
- Message for server instantiation
- Message for first "wait" stating that exchange will be continuous
- Message for run completion

examples/Network/ThreeBusCoSim/ThreeBusCoSimServer.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+      // 2. Respond with new data
+      std::ostringstream oss;
+      oss << std::scientific << std::setprecision(16);
+      oss << ir_signal_->read() << " " << ii_signal_->read();
+      // std::cout << "[SERVER] Sending: " << oss.str() << std::endl;
+
+      zmq::message_t reply{oss.str().data(), oss.str().size()};
+      socket_.send(reply, zmq::send_flags::none);
+    } while (status == CoSim::STEP);
```
nkoukpaizan, line 103: Ha, the trick is in this `do while`. Almost missed it. I recommend adding the behavior of `CoSim::STEP` and `CoSim::END` to the documentation.

nkoukpaizan (commented):
I have a few requests for better documentation and (sparse) diagnostic messages. Otherwise looks good.

examples/Network/ThreeBusCoSim/CoSim.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace GridKit
+{
+  namespace CoSim
+  {
+    using StatusT = int;
+
+    inline constexpr StatusT STEP = 1;
+    inline constexpr StatusT END  = 0;
```
superwhiskers, line 21: we should document the semantics of these here at a minimum

examples/Network/ThreeBusCoSim/run_cosim.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
superwhiskers:
```suggestion
#!/usr/bin/env sh
```

none of this is bash-specific. `/bin/sh` also works, generally portable, but `/usr/bin/env` will perform a `PATH` lookup and this is a development script. both are preferable to `/bin/bash` though, which buys us nothing in this case.

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
```diff
@@ -0,0 +1,176 @@
...
+    : vr_signal_(vr),
+      vi_signal_(vi),
+      ir_signal_(ir),
+      ii_signal_(ii),
+      ctx_{},
+      socket_(ctx_, zmq::socket_type::req)
+  {
+    ir_signal_->set(&ir_, &ir_idx_);
+    ii_signal_->set(&ii_, &ii_idx_);
+    socket_.connect("tcp://0.0.0.0:5556");
```
superwhiskers, line 66: doesn't matter as much right now but in the future we should really not hardcode the port, and in a test/example we shouldn't be exposing this on all interfaces by listening on `0.0.0.0`.

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
```diff
@@ -0,0 +1,176 @@
...
+   * @brief Send voltage to and receive current from server-side instance for a
+   * single time step.
+   */
+  void exchange()
+  {
+    // 1. Send data
+    std::ostringstream oss;
+    oss << std::scientific << std::setprecision(16);
+    oss << CoSim::STEP << " " << vr_signal_->read() << " " << vi_signal_->read();
+    // std::cout << "[CLIENT] Sending: " << oss.str() << std::endl;
```
superwhiskers: we shouldn't leave these commented out in the code. they should either be logged using a logging interface or removed. i think the former is more useful here for debugging.

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
PhilipFackler: I have implemented all but the third bullet. I'm not following your intent with that one. Can you clarify where this should happen and what should be communicated?

examples/Network/ThreeBusCoSim/ThreeBusCoSimServer.cpp
PhilipFackler: Should be resolved in [0b930a2](https://github.com/ORNL/GridKit/pull/537/commits/0b930a2ca88ac2299b00ac737c1b92dfced677fd)

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
PhilipFackler: Agree. Almost everything about this needs to be further generalized. :)

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
nkoukpaizan: `Log::summary() << "CLIENT: Ending simulation\n";` and `Log::summary() << "SERVER: Simulation stopped\n";` cover the last bullet.

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
PhilipFackler: In [0b930a2](https://github.com/ORNL/GridKit/pull/537/commits/0b930a2ca88ac2299b00ac737c1b92dfced677fd) I updated these to use the logger.

examples/Network/ThreeBusCoSim/CoSim.hpp
PhilipFackler: Resolved in [382aa63](https://github.com/ORNL/GridKit/pull/537/commits/382aa63bfc52929cf13a71ed6a07ceb690de9f06)

examples/Network/ThreeBusCoSim/ThreeBusCoSimClient.cpp
PhilipFackler: Would you recommend a specific change for this PR, or can we merge?

superwhiskers: i'll merge it after i've reviewed the updates and the ci checks have passed. we can leave this alone for now

examples/Network/ThreeBusCoSim/run_cosim.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+./CoSimServer &
+
+./CoSimClient
```
superwhiskers, line 9: one more comment in this file, we should not expect `PWD` to contain these files. i think it would be an improvement to:

- copy this file to the `build/` directory
- change the current working directory to the directory containing the file (retrieved using `dirname(1)` over `realpath(1)` over `$0`)
- then do `./CoSimServer`, `./CoSimClient`

or do `COSIM_DIRECTORY=$(dirname $(realpath $0))` and then `${COSIM_DIRECTORY}/CoSimServer`, etc. either works

superwhiskers (changes requested):
just one more comment and this should be good to go. runs fine on my machine

examples/Network/ThreeBusCoSim/run_cosim.sh
superwhiskers: alternatively, you could substitute in the `build/...` directory using cmake. your choice

PhilipFackler: I added this script as an easy way to run both apps together in the build directory. I'm thinking now that it might be better if it were removed altogether. When the time comes for a full-fledged co-simulation application, there should be a discussion about how users may best launch it. What do you think?

superwhiskers: the way i would probably go about doing it is to manually launch a bunch of workers on the hardware you would like to run a simulation on, then have a "frontend" program that is configured with a list of addresses to workers that submits a job to the worker services.

alternatively, perhaps one of these workers is delegated a "leader" and maintains a list of all of the details of a job or something like that and only serves as a way for other workers to retrieve parts of a job to simulate. this would be the only worker the frontend needs to connect to as it could be polled by other workers for more work to do.

maybe this doesn't answer your question, but this is my first thought as to how this could be launched and how it could operate. maybe the direction you were going in was different.

PhilipFackler: Oh, yes. :smile: My question was more "should we remove this for now?". But these are definitely interesting ideas for the (future) discussion I was anticipating.

superwhiskers: i think we should leave it for now as the test uses it. just make the minor change here and i'll merge it. the file will be removed later, though

PhilipFackler: It wouldn't make sense (at this point) to run this script from a different working directory since the apps in this example open a hard-coded file name (expected to be present).

PhilipFackler: It also doesn't make sense to place this script at the top-level build directory since this is (at the moment) just an isolated poc example.

superwhiskers: > [...] the apps in this example open a hard-coded file name (expected to be present).

changing the current working directory to the directory of the script---which was my first suggestion---would ensure it is always present because the script directory is where those files are

> It also doesn't make sense to place this script at the top-level build directory since this is (at the moment) just an isolated poc example.

if the reason for not wanting to make the script more robust is because it is a proof-of-concept, then sure, i guess. but the motivation behind these suggestions is to make it a little less annoying to run it out of the test harness.

PhilipFackler: Based on these comments, it seems that I misunderstood your original suggestions. I am already copying the script to the same binary dir where the apps are built. What you're suggesting then is to enable running
```
sh ./examples/Network/ThreeBusCoSim/run_cosim.sh
```
rather than
```
cd examples/Network/ThreeBusCoSim
sh ./run_cosim.sh
```
Am I understanding now?

superwhiskers: double-checked and yeah, i missed that it was already copied. what you are suggesting there is what i was requesting.

PhilipFackler: OK, this should be resolved now in [d4236ec](https://github.com/ORNL/GridKit/pull/537/commits/d4236ecb96f615ff2f957e007ca5e8760939dd2a)

superwhiskers (approved):

PR #461, opened 2026-06-23
10k Bus Case + Three Fully Validated Cases

lukelowry:
## Description

Three validated cases against PowerWorld with amazing accuracy & added ACTIVSg10k bus case.
 
Requires #525 #460, WECC case validation is failing until those merge.

### PhasorDynamics Cases

- [ACTIVSg200](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/ACTIVSg200/README.md)
- [ACTIVSg500](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/ACTIVSg500/README.md)
- [ACTIVSg2000](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/ACTIVSg2000/README.md)
- [ACTIVSg10k](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/ACTIVSg10k/README.md)
- [Hawaii](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/Hawaii/README.md)
- [IEEE39](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/IEEE39/README.md)
- [WECC240](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/cases/PhasorDynamics/WECC240/README.md)

wecc generator_omega

### Validation Cases

I chose these as the validation cases because they are fully, 100% modeled in GridKit.

- [ACTIVSg200 Validation](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/examples/PhasorDynamics/validation/ACTIVSg200/README.md)
- [ACTIVSg500 Validation](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/examples/PhasorDynamics/validation/ACTIVSg500/README.md)
- [WECC240 Validation](https://github.com/ORNL/GridKit/blob/lukel/tenK-dev/examples/PhasorDynamics/validation/WECC240/README.md)

Relative Error Against PowerWorld

| Field | ACTIVSg200 | ACTIVSg500 | WECC240 |
|---|---:|---:|---:|
| Generator real power, P | 0.031% / 0.116% | 0.016% / 0.066% | 0.005% / 0.022% |
| Generator reactive power, Q | 0.066% / 0.137% | 0.133% / 0.179% | 0.099% / 0.165% |
[...]

PR #538, opened 2026-08-13
Minimal corrections to Phasor Dynamics models for validation

lukelowry:
## Description
 
Validated against PowerWorld. Less than 0.5% relative error on WECC 240 case, ACTIVSg200 case, and ACTIVSg500 case.

 ## Further comments
 
Other open-source grid modeling frameworks do not even come close to this low a validation error against their validation tools, especially not for cases this large.  I don't want perfect to be the enemy of good + we can do cleanup more confidently after merging  #461.

### DIFF
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -222,6 +222,16 @@ namespace GridKit
        * preserving the seeded `efd`, latches attached Known inputs, and
        * publishes the reference to an attached `vref` signal.
        *
+       * @warning IEEE Std 421.4-2014 states: “In some programs, if
+       *          \f$K_{E}\f$ is entered as zero, \f$K_{E}\f$ is automatically
+       *          calculated by the program to represent a self-excited shunt
+       *          field and a trimmed rheostat as its initial condition.” GridKit
+       *          preserves the configured \f$K_{E}\f$ and resolves
+       *          \f$K_{E}^{\mathrm{eff}}\f$ using the PSS/E-compatible
+       *          \f$V_R = V_R^{\max}/10 = 0.1 V_R^{\max}\f$ rule. The divisor
+       *          10 is unitless and represents 10% of the maximum regulator
+       *          output.
+       *
        * @return Zero on success; nonzero when the configuration or operating point is rejected.
        */
       template <typename scalar_type, typename index_type>
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -222,16 +222,6 @@ namespace GridKit
        * preserving the seeded `efd`, latches attached Known inputs, and
        * publishes the reference to an attached `vref` signal.
        *
-       * @warning IEEE Std 421.5-2016 states: “In some programs, if
-       *          \f$K_{E}\f$ is entered as zero, \f$K_{E}\f$ is automatically
-       *          calculated by the program to represent a self-excited shunt
-       *          field and a trimmed rheostat as its initial condition.” GridKit
-       *          preserves the configured \f$K_{E}\f$ and resolves
-       *          \f$K_{E}^{\mathrm{eff}}\f$ using the PSS/E-compatible
-       *          \f$V_R = V_R^{\max}/10 = 0.1 V_R^{\max}\f$ rule. The divisor
-       *          10 is unitless and represents 10% of the maximum regulator
-       *          output.
-       *
        * @return Zero on success; nonzero when the configuration or operating point is rejected.
        */
       template <typename scalar_type, typename index_type>
@@ -322,13 +312,14 @@ namespace GridKit
 
         const ScalarT se0 = SB_ * Math::qramp(efdp0 - SA_);
 
-        RealT ke0 = Ke_;
-        if (ke0 == ZERO<RealT>)
+        RealT ke_eff0 = Ke_eff_;
+        if (Ke_ == ZERO<RealT>)
         {
-          ke0 = (Vrmax_ / 10.0 - static_cast<RealT>(se0)) / static_cast<RealT>(efdp0);
+          ke_eff0 = (Vrmax_ / 10.0 - static_cast<RealT>(se0))
+                    / static_cast<RealT>(efdp0);
         }
 
-        const ScalarT vfe0 = ke0 * efdp0 + se0;
+        const ScalarT vfe0 = ke_eff0 * efdp0 + se0;
         const ScalarT vr0  = vfe0;
         const ScalarT vhv0 = vr0 / Ka_;
 
@@ -361,7 +352,7 @@ namespace GridKit
         const ScalarT xll0  = ev0;
         const ScalarT vref0 = ev0 + vc0 - vs0 - uel_on_ * vuel0;
 
-        Ke_eff_ = ke0;
+        Ke_eff_ = ke_eff0;
 
         y[EFDP] = efdp0;
         y[VC]   = vc0;
@@ -768,6 +759,16 @@ namespace GridKit
        * masks let the residual select signal routing without
        * parameter-dependent control flow, which keeps its structure fixed for
        * sparse automatic differentiation.
+       *
+       * @warning IEEE Std 421.5-2016 states: “In some programs, if
+       *          \f$K_{E}\f$ is entered as zero, \f$K_{E}\f$ is automatically
+       *          calculated by the program to represent a self-excited shunt
+       *          field and a trimmed rheostat as its initial condition.” GridKit
+       *          preserves the configured \f$K_{E}\f$ and resolves
+       *          \f$K_{E}^{\mathrm{eff}}\f$ using the PSS/E-compatible
+       *          \f$V_R = V_R^{\max}/10 = 0.1 V_R^{\max}\f$ rule. The divisor
+       *          10 is unitless and represents 10% of the maximum regulator
+       *          output.
        */
       template <typename scalar_type, typename index_type>
       void Esdc1a<scalar_type, index_type>::setDerivedParameters()
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -222,7 +222,7 @@ namespace GridKit
        * preserving the seeded `efd`, latches attached Known inputs, and
        * publishes the reference to an attached `vref` signal.
        *
-       * @warning IEEE Std 421.4-2014 states: “In some programs, if
+       * @warning IEEE Std 421.5-2016 states: “In some programs, if
        *          \f$K_{E}\f$ is entered as zero, \f$K_{E}\f$ is automatically
        *          calculated by the program to represent a self-excited shunt
        *          field and a trimmed rheostat as its initial condition.” GridKit
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -351,7 +351,7 @@ namespace GridKit
         const ScalarT xll0  = ev0;
         const ScalarT vref0 = ev0 + vc0 - vs0 - uel_on_ * vuel0;
 
-        Ke_ = ke0;
+        Ke_eff_ = ke0;
 
         y[EFDP] = efdp0;
         y[VC]   = vc0;
@@ -564,7 +564,7 @@ namespace GridKit
 
         const ScalarT ec                = std::sqrt(wb[0] * wb[0] + wb[1] * wb[1]);
         const ScalarT ev_target         = vref + vs + uel_on_ * vuel - vc - vf;
-        const ScalarT vfe_target        = Ke_ * efdp + se;
+        const ScalarT vfe_target        = Ke_eff_ * efdp + se;
         const ScalarT efdp_rate         = (vr - vfe) / Te_;
         const ScalarT limited_efdp_rate = awmin(efdp, efdp_rate, ZERO<RealT>);
 
@@ -762,6 +762,8 @@ namespace GridKit
       template <typename scalar_type, typename index_type>
       void Esdc1a<scalar_type, index_type>::setDerivedParameters()
       {
+        Ke_eff_ = Ke_;
+
         // The lags are raised to the floor in place, so a negative value is
         // rejected here while the value as read is still available. verify()
         // reports the count.
--- a/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
+++ b/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
@@ -222,6 +222,16 @@ namespace GridKit
        * preserving the seeded `efd`, latches attached Known inputs, and
        * publishes the reference to an attached `vref` signal.
        *
+       * @warning IEEE Std 421.5-2016 states: “In some programs, if
+       *          \f$K_{E}\f$ is entered as zero, \f$K_{E}\f$ is automatically
+       *          calculated by the program to represent a self-excited shunt
+       *          field and a trimmed rheostat as its initial condition.” GridKit
+       *          preserves the configured \f$K_{E}\f$ and resolves
+       *          \f$K_{E}^{\mathrm{eff}}\f$ using the PSS/E-compatible
+       *          \f$V_R = V_R^{\max}/10 = 0.1 V_R^{\max}\f$ rule. The divisor
+       *          10 is unitless and represents 10% of the maximum regulator
+       *          output.
+       *
        * @return Zero on success; nonzero when the configuration or operating point is rejected.
        */
       template <typename scalar_type, typename index_type>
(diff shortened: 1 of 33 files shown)


nkoukpaizan:
Is this meant to supersede #525 ?

lukelowry:
> Is this meant to supersede #525 ?

Yes, I forgot to mention that

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
```diff
@@ -314,8 +310,15 @@ namespace GridKit
...
 
-        const ScalarT se0  = SB_ * Math::qramp(efdp0 - SA_);
-        const ScalarT vfe0 = (Ke_ + se0) * efdp0;
+        const ScalarT se0 = SB_ * Math::qramp(efdp0 - SA_);
+
+        RealT ke0 = Ke_;
+        if (ke0 == ZERO<RealT>)
+        {
+          ke0 = (Vrmax_ / 10.0 - static_cast<RealT>(se0)) / static_cast<RealT>(efdp0);
+        }
```
nkoukpaizan: I think this still qualifies as changing the input parameter (`Ke_`), and this is not in the README. Can we avoid it in the PR and successfully validate?

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/README.md
```diff
@@ -149,7 +166,7 @@ Symbol                              | Units  | Description
 $e_V$                               | [p.u.] | Voltage-error summing output      |
 $V_{\mathrm{LL}}$                   | [p.u.] | Input lead-lag output             |
 $V_{\mathrm{HV}}$                   | [p.u.] | High-value gate output            |
-$S_E$                               | [p.u.] | Exciter saturation coefficient    | Evaluated at $E_{\mathrm{fd}}'$
+$E_{\mathrm{fd}}'S_E(E_{\mathrm{fd}}')$ | [p.u.] | Scaled-quadratic saturation contribution |
```
nkoukpaizan: You might want to note as $s_e$. The corresponding [change in the implementation](https://github.com/ORNL/GridKit/pull/538/changes#diff-e094939b4f52fb3a2f6d6e929d4d7931e1c3243c912d3bdb7b551311f1df0b35R567) versus the [documented algebraic equation](https://github.com/ORNL/GridKit/blob/80549671a55b0cc24e8e4ce416d8534bba19d8ef/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/README.md?plain=1#L268) took me a while to figure out.

GridKit/Model/PhasorDynamics/Governor/Tgov1/README.md
```diff
@@ -12,99 +12,126 @@ Figure 1: Governor TGOV1 model. Figure courtesy of [PowerWorld](https://www.powe
...
-          )
+  0 &= -\dot P_v
+       + \dfrac{1}{T_1}\text{antiwindup}
+         \left(P_v,g_v;P_v^{\min},P_v^{\max}\right) \\
+  0 &= -\dot P_t-\dfrac{P_t-P_v-T_2\dot P_v}{T_3}.
 \end{aligned}
 ```
 
-CommonMath defines the [Anti-Windup](../../../../CommonMath.md#antiwindup)
+CommonMath defines the [anti-windup](../../../../CommonMath.md#antiwindup)
```
nkoukpaizan: I believe this was changed to `Antiwindup`, but that's minor.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -226,9 +233,14 @@ namespace GridKit
 
         ScalarT efdp = efd0 / (ONE<RealT> + omega * Ispdlim_);
         ScalarT ksat = SB_ * Math::qramp(efdp - SA_);
-        ScalarT ve   = ksat * efdp;
-        ScalarT vr   = Ke_ * efdp + ve;
-        ScalarT vtr  = vr / Ka_;
+        if (Ke_ == ZERO<RealT>)
+        {
+          Ke_ = (Vrmax_ / 10.0 - static_cast<RealT>(ksat)) / static_cast<RealT>(efdp);
+        }
```
nkoukpaizan, line 260: Changing parameter here as well.

nkoukpaizan (commented):
A few comments. Otherwise looks good!

GridKit/Model/PhasorDynamics/Governor/Tgov1/README.md
lukelowry: Addressed in [9082c8a](https://github.com/ORNL/GridKit/pull/538/commits/9082c8a83b84e4f67987ad8269d26b7b555539e5) thank you!

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/README.md
lukelowry: I apologize for that, let me know if the change in [9082c8a](https://github.com/ORNL/GridKit/pull/538/commits/9082c8a83b84e4f67987ad8269d26b7b555539e5) is clearer

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
lukelowry: This is one of the few that I was unsure about in our meeting today. I think that if we instead define a derived parameter as I did in [9082c8a](https://github.com/ORNL/GridKit/pull/538/commits/9082c8a83b84e4f67987ad8269d26b7b555539e5), this will hopefully resolve any confusion. It's not so much changing the parameter as deriving a new parameter. Let me know what you think of my changes

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
lukelowry: See other response

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
nkoukpaizan: Understood. In that case, adding this to the [Model Derived Parameters]( https://github.com/ORNL/GridKit/blob/80549671a55b0cc24e8e4ce416d8534bba19d8ef/GridKit/Model/PhasorDynamics/Exciter/ESDC1A/README.md#model-derived-parameters) section of the documentation should help.

lukelowry: Please see changes in [dbbafb2](https://github.com/ORNL/GridKit/pull/538/commits/dbbafb27de3c0de252300aa1883bf870e07ea510)

abirchfield (approved):
A good compromise that addresses the issues raised in #525 while allowing us to have excellent validation in the short term. Exciting that we are almost ready to complete the 10k case.

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
pelesh: In #525 parameter `Ke_` was overwritten, now the new value is stored in `Ke_eff_` and is used instead of `Ke_`. Unless the new parameter `Ke_eff_` has well defined physical meaning and the governing physics is documented in the README file, this change will fall outside the scope of this PR.

lukelowry:
added specification information and warnings everywhere $K_E^{eff}$ and $K_E$ is mentioned. should be ready!

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
```diff
@@ -314,8 +320,15 @@ namespace GridKit
...
```
nkoukpaizan: Can we use `Ke_eff_` directly in the initialization? It should be correctly set by `setDerivedParameters` with the relevant checks.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -226,9 +243,15 @@ namespace GridKit
...
         ScalarT efdp = efd0 / (ONE<RealT> + omega * Ispdlim_);
         ScalarT ksat = SB_ * Math::qramp(efdp - SA_);
-        ScalarT ve   = ksat * efdp;
-        ScalarT vr   = Ke_ * efdp + ve;
-        ScalarT vtr  = vr / Ka_;
+        Ke_eff_      = Ke_;
+        if (Ke_ == ZERO<RealT>)
+        {
+          Ke_eff_ = (Vrmax_ / 10.0 - static_cast<RealT>(ksat)) / static_cast<RealT>(efdp);
+        }
```
nkoukpaizan: Move to `setDerivedParameters` as well.

nkoukpaizan (approved):
I recommend setting `Ke_eff_` uniformly in `setDerivedParameters` and moving the warnings there (instead of in `initialize`).

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
lukelowry: good point, will do now

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
lukelowry: It needs access to the initialized states so as to be consistent with the other models' usage of `setDerivedParameters`. it needs to have some part of this in initialization. I added the setDerivedParameters to this model because it didn't have it since its older, so good catch regardless. see [d39ae14](https://github.com/ORNL/GridKit/pull/538/commits/d39ae1494c2c623c7054009030c09675d36ef084)

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
lukelowry: Partialyl resolved in [d39ae14](https://github.com/ORNL/GridKit/pull/538/commits/d39ae1494c2c623c7054009030c09675d36ef084)

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
```diff
@@ -314,8 +310,16 @@ namespace GridKit
...
-        const ScalarT se0  = SB_ * Math::qramp(efdp0 - SA_);
-        const ScalarT vfe0 = (Ke_ + se0) * efdp0;
+        const ScalarT se0 = SB_ * Math::qramp(efdp0 - SA_);
+
+        RealT ke_eff0 = Ke_eff_;
+        if (Ke_ == ZERO<RealT>)
+        {
+          ke_eff0 = (Vrmax_ / 10.0 - static_cast<RealT>(se0))
+                    / static_cast<RealT>(efdp0);
+        }
```
nkoukpaizan: Oh, I see `ke_eff0` needs signal values. I missed some of that.

My underlying objection was the `ke0/ke_eff0` followed by `Ke_eff_ = ke_eff0` later. Now, we also end up checking `Ke_ == ZERO<RealT>` to set `ke_eff0`, which is confusing to me. I think the member variable should only be set once, and the conditional should match the value that's being set.

lukelowry: Okay no worries I think we both misunderstood each other. Give me a moment I can make this clearer

lukelowry: Alright, I like this implemetnation ([6881eff](https://github.com/ORNL/GridKit/pull/538/commits/6881efff05f99b15536287cdde55d90f1afe341d)) tell me what you think

lukelowry:
@pelesh @nkoukpaizan Let me know if you think this is ready! Good on my end

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -287,7 +293,8 @@ namespace GridKit
         const RealT limit_tolerance = static_cast<RealT>(4.0) * std::numeric_limits<RealT>::epsilon();
         if (pv0_value < Pvmin_ - limit_tolerance || pv0_value > Pvmax_ + limit_tolerance)
         {
-          Log::error() << "Tgov1: initial valve position is outside limits\n";
+          Log::warning() << "Tgov1: initial valve position is outside [Pvmin, Pvmax]. "
+                            "Check initial dispatch and valve limits\n";
           return 1;
```
nkoukpaizan: @lukelowry Are you sure about this change? If we are returning 1 and not implementing any correction, this looks like an `error` to me.

nkoukpaizan:
> @pelesh @nkoukpaizan Let me know if you think this is ready! Good on my end

Fine by me... I can always find something to nitpick, but I'll stop there. I'll let @pelesh have the last word.

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -228,32 +265,50 @@ namespace GridKit
...
+        const ScalarT pv0    = pm0 + Dt_ * omega0;
+        const ScalarT pturb0 = pv0;
+        const ScalarT pref0  = omega0 + R_ * pv0;
+
+        const RealT pv0_value       = static_cast<RealT>(pv0);
+        const RealT limit_tolerance = static_cast<RealT>(4.0) * std::numeric_limits<RealT>::epsilon();
+        if (pv0_value < Pvmin_ - limit_tolerance || pv0_value > Pvmax_ + limit_tolerance)
+        {
+          Log::warning() << "Tgov1: initial valve position is outside [Pvmin, Pvmax]. "
+                            "Check initial dispatch and valve limits\n";
```
pelesh: This has to error out and return. There is no point in continuing initialization upon such error.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
```diff
@@ -226,9 +243,20 @@ namespace GridKit
...
         ScalarT ksat = SB_ * Math::qramp(efdp - SA_);
-        ScalarT ve   = ksat * efdp;
-        ScalarT vr   = Ke_ * efdp + ve;
-        ScalarT vtr  = vr / Ka_;
+        if (Ke_ == ZERO<RealT>)
+        {
+          Ke_eff_ = (Vrmax_ / 10.0 - static_cast<RealT>(ksat))
+                    / static_cast<RealT>(efdp);
+          Log::warning() << "Ieeet1: Ke is zero so effective Ke is derived during initialization\n";
+        }
```
pelesh, line 256: Two issues here:
- What if `efdp` is zero? I would check if `Ke_eff_` is a valid floating point number and, if not, print error and return 1.
- Outputting warning here will result in too many warnings being printed out for large systems, so users will eventually ignore them all. I think the "warning" stanza in the doxygen docs is sufficient here. I would remove printing warning message here.

pelesh (approved):
All major issues we discussed were addressed and all tests pass.

There is a couple of nitpicking issues that would be good to address before we merge.

GridKit/Model/PhasorDynamics/Exciter/IEEET1/Ieeet1Impl.hpp
pelesh: To address the second bullet, it suffices to do something like this:
```suggestion
          Ke_eff_ = (Vrmax_ / 10.0 - static_cast<RealT>(ksat))
                    / static_cast<RealT>(efdp);
          Log::misc() << "Ieeet1: Ke is zero so effective Ke is derived during initialization\n";
        }
```

GridKit/Model/PhasorDynamics/Exciter/ESDC1A/Esdc1aImpl.hpp
superwhiskers, line 156: this wasn't added in this pull request, but as i have said in #460, this should 1) not increment an error count---this is a bad pattern and 2) should be a common utility function, not something we define in each case it is used. this comment applies to all instances of this closure, not just this one instance

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
```diff
@@ -183,16 +193,17 @@ namespace GridKit
         }
 
         // Resize signal variable data
-        ws_.resize(1);
-        ws_indices_.resize(1);
-        ws_[0]         = 0.0;
-        ws_indices_[0] = INVALID_INDEX<IdxT>;
+        const auto signal_size = static_cast<size_t>(Tgov1ExternalVariables::MAXIMUM);
```
superwhiskers, line 196: just a future note but it may be a good idea to have some kind of function or method to get the size instead of depending upon the implementation detail of `::MAXIMUM` on variable enumerations, etc

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
lukelowry: Addressed in [9e32dc6](https://github.com/ORNL/GridKit/pull/538/commits/9e32dc667e24dcce3d606cc4380e72e67fec4aad)

GridKit/Model/PhasorDynamics/Governor/Tgov1/Tgov1Impl.hpp
lukelowry: okay that is good advice thank you. Will defer for 0.3 only because I do this in other models

Issue #498, opened 2026-07-20
Add reference files for examples that are used as tests


nkoukpaizan:
Several examples are currently used as tests (CMake `add_test`), but don't have reference solutions to verify correctness. They are essentially smoke tests ensuring that the examples runs. Listing the relevant cases for PhasorDynamics here. It would be good to add reference solutions, at least for the smaller cases, with the understanding that validation for larger cases is contingent on merging some models.
- [x] TenGenClassical
- [x] TenGenGenrou
- [x] Hawaii
- [x] NewEngland
- [x] Illinois
- [x] WECC
- [x] Texas
- [x] ACTIVSg10k

cc @lukelowry @abirchfield @pelesh @PhilipFackler

lukelowry:
Good issue to track. I am waiting until the last 6 models are merged so we can do this in one batch and only once

pelesh:
@lukelowry and @nkoukpaizan: Has this been addressed in #577?

nkoukpaizan:
Yes, I believe this has been addressed in #577.

PR #539, opened 2026-08-13
Output post-event monitor values

lukelowry:
## Description
 
Output post-event monitor values. Eases validation burden, and this should be printed to monitor output regardless

 ## Further comments
 
Initially done in #461 and suggested isolated PR per @nkoukpaizan


nkoukpaizan (approved):
Thanks for isolating this. It reminds me #498, as I expected more reference files to change.

PR #542, opened 2026-08-18
Add option for IDA consistent initial conditions mode

Steven-Roberts:
## Description
 
As discussed Monday, this adds the option to pick between [IDA's two modes for computing consistent initial conditions](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDACalcIC).

 ## Proposed changes
 
This add `Ida::setConsistentICType(IdaConsistentICType consistent_ic_type)` and a corresponding `consistent_ic_type` JSON property to set the mode. It keeps the `bool findConsistent` argument to `Ida::initializeSimulation`, so it is still possible to skip the solve.
 
 ## Checklist
 
- [x] All tests pass.
- [x] The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes. If this is a minor PR that is part of a larger fix already included in the file, state so.

### DIFF
--- a/tests/UnitTests/Solver/Dynamic/IdaTests.hpp
+++ b/tests/UnitTests/Solver/Dynamic/IdaTests.hpp
@@ -384,8 +384,8 @@ namespace GridKit
     public:
       ConsistentICTypeEvaluator() = default;
 
-      explicit ConsistentICTypeEvaluator(bool consistent_derivative)
-        : consistent_derivative_(consistent_derivative)
+      explicit ConsistentICTypeEvaluator(bool steady_state)
+        : steady_state_(steady_state)
       {
       }
 
@@ -403,7 +403,7 @@ namespace GridKit
 
         y[0]       = 0.0;
         y[1]       = 10.0;
-        yp[0]      = consistent_derivative_ ? 1.0 : 0.0;
+        yp[0]      = steady_state_ ? 0.0 : 2.0; // Purposefully inconsistent guess for non-steady-state case.
         yp[1]      = 0.0;
         tag_       = {true, false};
         abs_tol[0] = 0.0;
@@ -428,14 +428,14 @@ namespace GridKit
         const auto* y  = y_.getData();
         const auto* yp = yp_.getData();
 
-        f[0] = yp[0] - 1.0;
+        f[0] = yp[0] + y[0] + y[1] - 1.0;
         f[1] = y[1] - y[0];
         f_.setDataUpdated();
         return 0;
       }
 
     private:
-      bool consistent_derivative_{false};
+      bool steady_state_{false};
     };
   } // namespace Model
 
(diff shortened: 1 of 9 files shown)


tests/UnitTests/Solver/Dynamic/IdaTests.hpp
```diff
@@ -503,6 +571,36 @@ namespace GridKit
...
+        {
+          Model::ConsistentICTypeEvaluator<ScalarT, IdxT> model;
+
+          Ida<ScalarT, IdxT> ida(&model);
+          ida.setConsistentICType(AnalysisManager::Sundials::IdaConsistentICType::YA_YDP);
+          ida.configureSimulation();
+          ida.initializeSimulation(0.0);
+
+          success *= (std::abs(model.yp().getData()[0] - 1.0) < 1.0e-10);
+          success *= (std::abs(model.y().getData()[1]) < 1.0e-10);
```
nkoukpaizan: I'm not sure I understand the `1.0e-10` tolerance choice.

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
```diff
@@ -503,6 +571,36 @@ namespace GridKit
...
+
+      TestOutcome consistentICType()
+      {
+        TestStatus success = true;
+
+        {
+          Model::ConsistentICTypeEvaluator<ScalarT, IdxT> model;
+
+          Ida<ScalarT, IdxT> ida(&model);
+          ida.setConsistentICType(AnalysisManager::Sundials::IdaConsistentICType::YA_YDP);
```
nkoukpaizan, line 590: This works well to make `yp` consistent with the equations. If I understood @pelesh's plan correctly, the question is whether we can use `IdaConsistentICType::Y` to solve for inconsistent `y` while fixing `yp=0` (for a system with differential equations that admits a steady state solution). 

I tried to use `IdaConsistentICType::Y` for the Texas test case, unsuccessfully. @Steven-Roberts My understanding was that this would not work, but is it expected to?
```
[IDAICFailFlag] The linesearch algorithm failed: step too small or too many backtracks.
29: terminate called after throwing an instance of 'AnalysisManager::Sundials::SundialsException'
29:   what():  Method in Ida class failed!
```
I could be misunderstanding the plans...

nkoukpaizan (commented):
Looks good to me in terms of providing the option to set the initialization type. I am not sure that if addresses the underlying question about steady-state initialization.

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
Steven-Roberts: Is the concern that it's too arbitrary or not tied to the tolerance IDA is using internally? Ultimately, this just needs to check the algebraic variables is close to 0.

tests/UnitTests/Solver/Dynamic/IdaTests.hpp
Steven-Roberts: Yes, I'm seeing the same issue for the Texas case. Can you explain why you don't expect it to work?

It seems a necessary condition for this working is $f_y(y_0, 0)$ is invertible where $f(y, y')$ is the DAE residual function. Perhaps this is not the case.

I'll need to look closer at the logs to understand why it's failing to converge.

Steven-Roberts: From debugging the solver, it's seemingly getting stuck in a local minimum rather than a true root. I think a better initial guess is needed to get it to converge.

nkoukpaizan: Marking this as resolved, since we have a unit test for getting a steady-state solution with `IdaConsistentICType::Y` and `yp=0`. I think the issues with the examples (e.g. Texas case) are out of scope for this PR.

nkoukpaizan (approved):

Issue #528, opened 2026-08-07
Testing issue: Excessive output from `PhasorDynamicsControllerRepcaTest`


pelesh:
### Failed test outputs or steps to reproduce the failure(s)

    
`PhasorDynamicsControllerRepcaTest` generates excessive output. See below. Perhaps We should reconsider how tests for failures are designed. So many warnings and error messages printed are confusing for a user.

Other tests should be reviewed, as well.

<details>
    
```console
test 48
      Start 48: PhasorDynamicsControllerRepcaTest

48: Test command: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics/test_phasor_controller_repca
48: Working Directory: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics
48: Test timeout computed to be: 1500
48: [MESSAGE] Testing REPCA defaults, parameter floors, and invalid configurations. Logged errors and warnings are expected.
48: [WARNING] Repca: Tfltr, Tfv, Tp, and Tlag below 0.001 s are raised to that floor to keep the controller lags well posed
48: [ERROR] Repca: ir signal is required
48: [ERROR] Repca: ii signal is required
48: [ERROR] Repca: p signal is required
48: [ERROR] Repca: q signal is required
48: [WARNING] Repca: Tfltr, Tfv, Tp, and Tlag below 0.001 s are raised to that floor to keep the controller lags well posed
48: [WARNING] Repca: Tfltr, Tfv, Tp, and Tlag below 0.001 s are raised to that floor to keep the controller lags well posed
48: [WARNING] Repca: Tfltr, Tfv, Tp, and Tlag below 0.001 s are raised to that floor to keep the controller lags well posed
48: [ERROR] Repca: ir signal is required
48: [ERROR] Repca: ii signal is required
48: [ERROR] Repca: p signal is required
48: [ERROR] Repca: q signal is required
48: [ERROR] Repca: mva must define a finite positive component power base
48: [ERROR] Repca: Tfv must be non-negative
48: [WARNING] Repca: Tfltr, Tfv, Tp, and Tlag below 0.001 s are raised to that floor to keep the controller lags well posed
48: [ERROR] Repca: dbdlow <= 0 <= dbdupper is required
48: [ERROR] Repca: dbdlow <= 0 <= dbdupper is required
48: [ERROR] Repca: emin <= 0 <= emax is required
48: [ERROR] Repca: emin <= 0 <= emax is required
[...]

Issue #527, opened 2026-08-07
Testing issue: Misleading output in `PhasorDynamicsSystemSingleComponentTest`


pelesh:
### Failed test outputs or steps to reproduce the failure(s)

`PhasorDynamicsSystemSingleComponentTest` returns warning but passes anyway. If the warning is intentional, we need to add a message stating it. Otherwise, we ned to fix the warning.

        
```console
test 53
      Start 53: PhasorDynamicsSystemSingleComponentTest

53: Test command: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics/test_phasor_system_single_component
53: Working Directory: /Users/55y/src/gridkit/build-clang/tests/UnitTests/PhasorDynamics
53: Test timeout computed to be: 1500
53: --- PASS: Test branch
53: --- PASS: Test bus
53: --- PASS: Test busFault
53: --- PASS: Test ieeet1
53: Testing expected ESDC1A missing-bus configuration error.
53: [ERROR] Esdc1a: bus pointer is null
53: --- PASS: Test esdc1a
53: --- PASS: Test load
53: --- PASS: Test loadZIP
53: --- PASS: Test regca
53: [WARNING] Reecb: any of Trv, Tp, Tiq, or Tpord below 0.001 s is raised to that floor to keep the controller lags well posed
53: --- PASS: Test reecb
53: Testing expected component initialization failure.
53: [ERROR] Regca: terminal voltage magnitude must be at least VA1 at initialization
53: [ERROR] System model initialization failed with status 1
53: --- PASS: Test initializationFailure
53: --- PASS: Test genrou
53: --- PASS: Test genClassical
53: --- PASS: Test tgov1
53: Testing expected HYGOV missing-output configuration error.
53: [ERROR] Hygov: pmech output signal must be assigned
53: --- PASS: Test hygov
53: 
53: 
53: Test Summary
53: ----------------------------
53:     Successful tests:     14
53:     Failed test:          0
53:     Skipped tests:        0
53:     Expected failures:    0
53:     Unexpected successes: 0
53: 
53/60 Test #53: PhasorDynamicsSystemSingleComponentTest ...   Passed    0.03 sec
```

### GridKit™ version

`develop` at aebd086742636d4959ed85e729df1f403f11fff6

### System and environment details

MacOS 15.7.8, clang 18 from MacPorts

pelesh:
#421 adds more warnings to this test:
```
54: --- PASS: Test regca
[...]

PR #548, opened 2026-08-24
Lower verbosity for Phasor Dynamics testing

lukelowry:
## Description
 
Addresses #527 and #528.
 
 
 ## Proposed changes
 
Set verbosity to `NONE` before the expected failure tests
 
 ## Further comments
 
@nkoukpaizan This does not fully address the open issues but it helps with the usability concern, I think. Let me know what you think

### DIFF
--- a/tests/UnitTests/PhasorDynamics/ControllerRepcaTests.hpp
+++ b/tests/UnitTests/PhasorDynamics/ControllerRepcaTests.hpp
@@ -47,8 +47,9 @@ namespace GridKit
       {
         TestStatus success = true;
 
-        noteExpectedLogs("Testing REPCA defaults, parameter floors, and invalid "
-                         "configurations. Logged errors and warnings are expected.");
+        const auto previous_verbosity = Log::verbosity();
+        // Use EVERYTHING to inspect diagnostics from the invalid cases below.
+        Log::setVerbosity(Log::Verbosity::NONE);
 
         PhasorDynamics::Bus<ScalarT, IdxT> bus(1.0, 0.0);
 
@@ -237,6 +238,7 @@ namespace GridKit
                                   floored_residuals,
                                   "floored time constants");
 
+        Log::setVerbosity(previous_verbosity);
         return success.report(__func__);
       }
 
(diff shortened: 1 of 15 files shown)


nkoukpaizan:
A few comments:
- Setting the verbosity to `NONE` in `main` for the tests may be too broad. Perhaps add some comments in the code to the effect this being a temporary solution and how one might change this setting to get more verbose outputs. 
- Note that we are still printing messages, e.g., ` Testing expected ESDC1A missing-bus configuration error.`, even though the errors have been suppressed.
- This addresses the two linked issues, but the excessive output concern applies to several other tests, e.g., `PhasorDynamicsExciterEsdc1aTest`.

lukelowry:
@nkoukpaizan I see, how about my change in [ca00d14](https://github.com/ORNL/GridKit/pull/548/commits/ca00d14b4f9b5cd8480df2f84a84be77bfa0a958)? Changes back to previous verbosity

tests/UnitTests/PhasorDynamics/ControllerRepcaTests.hpp
```diff
@@ -47,8 +47,9 @@ namespace GridKit
       {
         TestStatus success = true;
 
-        noteExpectedLogs("Testing REPCA defaults, parameter floors, and invalid "
-                         "configurations. Logged errors and warnings are expected.");
+        const auto previous_verbosity = Log::verbosity();
+        // Use EVERYTHING to inspect diagnostics from the invalid cases below.
```
nkoukpaizan:
```suggestion
        // Changing the verbosity to NONE to suppress expected errors and warning for the following tests.
        // Use EVERYTHING to inspect diagnostics from the invalid cases below.
```

I like having a comment here. I recommend adding something to this effect to other instances.

lukelowry:
@nkoukpaizan Addressed in [a6b6a75](https://github.com/ORNL/GridKit/pull/548/commits/a6b6a75768fc6aed6c8c5bce05e3f0abad6bfece), I believe.

nkoukpaizan (approved):

PR #550, opened 2026-08-25
Use std::once_flag and std::call_once for time constant warnings.

nkoukpaizan:
## Description
 
This uses `std:once_flag` and `std::call_once` to only warn user about the raised time constants once.
This removes the excessive warnings for this known issue. 
 
 ## Proposed changes
 
- Added `static void logTimeConstantWarning()` to the relevant header files after `TIME_CONSTANT_MINIMUM`
- Moved `Log::Warning()` call to the static method
- Introduced `std:once_flag` and `std::call_once` at the call site, when the warning is warranted.
 
 ## Checklist
 
- [x ] All tests pass.
- [N/A] The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 Will need to better generalize if a similar pattern is needed later.

### DIFF
--- a/GridKit/Model/PhasorDynamics/Controller/REECB/Reecb.hpp
+++ b/GridKit/Model/PhasorDynamics/Controller/REECB/Reecb.hpp
@@ -196,6 +196,8 @@ namespace GridKit
         ScalarT& Vi();
 
         static constexpr RealT TIME_CONSTANT_MINIMUM = static_cast<RealT>(1.0e-3);
+        static void logTimeConstantWarning();
+
         static constexpr RealT VMEAS_MINIMUM         = static_cast<RealT>(0.01);
 
         BusT* bus_{nullptr};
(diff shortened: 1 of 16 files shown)


lukelowry (approved):
Thank you, I would have struggled with this. 

If I run this on two different cases, will it error on the second case if it raises the same warnings?

nkoukpaizan:
> If I run this on two different cases, will it error on the second case if it raises the same warnings?

I don't see a scenario where we'd get an error, and I am not sure I understand the use-case you're describing, so let try these:
- This is thread-safe for contingency analysis, so we get one warning for all contingencies.
- If you have two separate executions of `DynamicSimulation`, you'll get a warning per executable.

lukelowry:
> > If I run this on two different cases, will it error on the second case if it raises the same warnings?
> 
> 
> 
> I don't see a scenario where we'd get an error, and I am not sure I understand the use-case you're describing, so let try these:
> 
> - This is thread-safe for contingency analysis, so we get one warning for all contingencies.
> 
> - If you have two separate executions of `DynamicSimulation`, you'll get a warning per executable.

Okay that resolves my concern. Thank you sir

superwhiskers (approved):
this looks fine to me

PR #552, opened 2026-08-27
Cloning Components in PowerElectronics Module

abdourahmanbarry:
## Description
 
This pull request implements clone functionality for components in the ```PowerElectronics``` module. This gives us the ability to create independent copies of components such that modifications to one do not affect the other.
 
 ## Proposed changes
 
Each component implements the ```isCloneable``` and ```clone``` methods. The ```isCloneable``` method returns ```true``` if the component is cloneable and ```false``` otherwise. The ```clone``` method returns a pointer to a new object with the same state as the component from which it was cloned. After cloning a component, the user should set the data pointers for the clone. 

 ## Checklist
 
- [x] All tests pass.
 
 ## Further Comments
This functionality is useful for the partitioning implementation.

### DIFF
--- a/GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
+++ b/GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
@@ -55,6 +55,25 @@ namespace GridKit
   }
 
   /**
+   * @brief Compute the absolute tolerance for each variable in the model
+   *
+   * @param rel_tol The relative tolerance which can be used to pick the
+   *        absolute tolerance.
+   * @tparam ScalarT Scalar data type
+   * @tparam IdxT Index data type
+   * @return int 0 if successful, non-zero otherwise.
+   *
+   * This represents a "noise" level close to zero for which pure relative
+   * error cannot be used.
+   */
+  template <class ScalarT, typename IdxT>
+  int Capacitor<ScalarT, IdxT>::setAbsoluteTolerance(RealT rel_tol)
+  {
+    abs_tol_.setToConst(static_cast<ScalarT>(rel_tol));
+    return 0;
+  }
+
+  /**
    * @brief Evaluate the resisdual of the Capcitor
    *
    */
@@ -117,6 +136,18 @@ namespace GridKit
   int Capacitor<ScalarT, IdxT>::evaluateAdjointIntegrand()
   {
     return 0;
+  }
+
+  template <class ScalarT, typename IdxT>
+  bool Capacitor<ScalarT, IdxT>::isCloneable() const
+  {
+    return true;
+  }
+
+  template <class ScalarT, typename IdxT>
+  CircuitComponent<ScalarT, IdxT>* Capacitor<ScalarT, IdxT>::clone() const
+  {
+    return new Capacitor<ScalarT, IdxT>(*this);
   }
 
   // Available template instantiations
--- a/GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
+++ b/GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
@@ -139,12 +139,6 @@ namespace GridKit
   }
 
   template <class ScalarT, typename IdxT>
-  bool Capacitor<ScalarT, IdxT>::isCloneable() const
-  {
-    return true;
-  }
-
-  template <class ScalarT, typename IdxT>
   CircuitComponent<ScalarT, IdxT>* Capacitor<ScalarT, IdxT>::clone() const
   {
     return new Capacitor<ScalarT, IdxT>(*this);
--- a/GridKit/Model/PowerElectronics/CircuitComponent.hpp
+++ b/GridKit/Model/PowerElectronics/CircuitComponent.hpp
@@ -27,6 +27,138 @@ namespace GridKit
 
     CircuitComponent() = default;
 
+    CircuitComponent(const CircuitComponent& other)
+      : n_extern_(other.n_extern_),
+        n_intern_(other.n_intern_),
+        extern_indices_(other.extern_indices_),
+        size_(other.size_),
+        nnz_(other.nnz_),
+        size_quad_(other.size_quad_),
+        size_opt_(other.size_opt_),
+        current_jac_size_(other.current_jac_size_),
+
+        // These pointers refer to storage supplied by a parent system.
+        // The copied component must be connected to its own storage later.
+        y_int_(nullptr),
+        yp_int_(nullptr),
+        f_int_(nullptr),
+
+        tag_(other.tag_),
+        time_(other.time_),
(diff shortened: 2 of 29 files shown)


superwhiskers (approved):
this change looks pretty inoffensive. one comment, however:

why are we adding `isCloneable` for only one component (the `Bus`) to not support cloning? additionally, might it be better to implement this by using a subclass of `CircuitComponent` that implies cloneability, or something else? it's not like cloning is going to change at runtime, so it might be cleaner to add some structural way to ensure a component being used supports cloning without doing a runtime check by calling `isCloneable`.

tests/UnitTests/PowerElectronics/ComponentCloneTests.hpp
```diff
@@ -0,0 +1,263 @@
...
+#include <GridKit/Model/PowerElectronics/MicrogridBusDQ/MicrogridBusDQ.hpp>
+#include <GridKit/Model/PowerElectronics/MicrogridLine/MicrogridLine.hpp>
+#include <GridKit/Model/PowerElectronics/MicrogridLoad/MicrogridLoad.hpp>
+#include <GridKit/Testing/TestHelpers.hpp>
+#include <GridKit/Testing/Testing.hpp>
+
+namespace GridKit
+{
+  template <typename ComponentT, typename ScalarT, typename IdxT>
+  bool verifyComponentClone(ComponentT& component)
```
nkoukpaizan: This should be in the `Testing` namespace. Perhaps a private method of `CircuitComponentCloneTests`?

GridKit/Model/PowerElectronics/CircuitComponent.hpp
```diff
@@ -27,6 +27,153 @@ namespace GridKit
 
     CircuitComponent() = default;
 
+    CircuitComponent(const CircuitComponent& other)
```
nkoukpaizan, line 57: Please add some documentation for this constructor.

GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
```diff
@@ -119,6 +138,18 @@ namespace GridKit
...
+  template <class ScalarT, typename IdxT>
+  bool Capacitor<ScalarT, IdxT>::isCloneable() const
+  {
+    return true;
+  }
+
+  template <class ScalarT, typename IdxT>
+  CircuitComponent<ScalarT, IdxT>* Capacitor<ScalarT, IdxT>::clone() const
+  {
+    return new Capacitor<ScalarT, IdxT>(*this);
```
nkoukpaizan, line 144: We are currently not checking that the components are clonable before cloning...

nkoukpaizan (commented):
A few comments. At a higher level:
- I don't see anything preventing calling the copy constructor on non-clonable components.
- It's not clear to me what is the intended functionality of the clones. More documentation would be helpful.

abdourahmanbarry:
> this change looks pretty inoffensive. one comment, however:
> 
> why are we adding `isCloneable` for only one component (the `Bus`) to not support cloning? additionally, might it be better to implement this by using a subclass of `CircuitComponent` that implies cloneability, or something else? it's not like cloning is going to change at runtime, so it might be cleaner to add some structural way to ensure a component being used supports cloning without doing a runtime check by calling `isCloneable`.

I added clone to ```CircuitComponent``` alone for now since it is the one for which I have immediate use for. 

I think having a cloneable subclass such as ```CloneableCircuitComponent``` makes sense. I can try doing that. This will also address some of @nkoukpaizan's comments as well.

GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
abdourahmanbarry: I expect a user to call ```isCloneable``` before calling clone.

alexander-novo:
I talked to @abdourahmanbarry in person. I think it makes a lot more sense to have all components be cloneable and remove the `isCloneable` function. Then for components where the implementation of `clone()` is out of scope for this PR we can simply throw an error stating the function isn't implemented.

The problem with a `CloneableCircuitComponent` interface is that to call `clone()` you'd need to know you have a `CloneableCircuitComponent`. For cloning an entire system, you'd need to know all of the components are a `CloneableCircuitComponent` which would make `component_type = CloneableCircuitComponent` which is the same thing as just having all circuit components as `CloneableCircuitComponent` (modding out dynamic typecasting).

GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
abdourahmanbarry: I have gotten rid of ```isCloneable```. Now non-cloneable classes just throw an error as suggested by @alexander-novo .

abdourahmanbarry:
> A few comments. At a higher level:
> 
> * I don't see anything preventing calling the copy constructor on non-clonable components.
> * It's not clear to me what is the intended functionality of the clones. More documentation would be helpful.

I am using clone as part of my implementation for partition interfaces in the next pull request on this stack. I have also updated the implementation of the copy constructor such that it should be okay to call it independently of the clone. Let me know if these changes look good.

GridKit/Model/PowerElectronics/CircuitComponent.hpp
abdourahmanbarry: I have added more documentation to this.

tests/UnitTests/PowerElectronics/ComponentCloneTests.hpp
abdourahmanbarry: Updated.

GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
nkoukpaizan: I still see a few instances of `isCloneable`.

nkoukpaizan:
> > A few comments. At a higher level:
> > 
> > * I don't see anything preventing calling the copy constructor on non-clonable components.
> > * It's not clear to me what is the intended functionality of the clones. More documentation would be helpful.
> 
> I am using clone as part of my implementation for partition interfaces in the next pull request on this stack. I have also updated the implementation of the copy constructor such that it should be okay to call it independently of the clone. Let me know if these changes look good.

What I was trying to get at is that if `clone()` is the desired path, I think the copy constructor should be `protected`.

GridKit/Model/PowerElectronics/Capacitor/Capacitor.cpp
abdourahmanbarry: Thanks for catching that. I have removed those instances of ```isCloneable``` from the components.

nkoukpaizan (approved):

PR #556, opened 2026-08-28
Eliminate useless state in Genrou and Gensal

lukelowry:
## Description
 
Eliminates the algebraic real and imaginary variables that served no purpose. This does not change anything, and the generator still owns its injection current variables.
 
 ## Proposed changes
 
- Eliminate Norton current variables
 
 ## Checklist

 
- [x] All tests pass.
- N/A The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments

The documentation was already in the cleaner form, not sure when they diverged

### DIFF
--- a/tests/IntegrationTests/PhasorDynamics/PDIntegrationTests.hpp
+++ b/tests/IntegrationTests/PhasorDynamics/PDIntegrationTests.hpp
@@ -243,6 +243,8 @@ namespace GridKit
 
         // Set up simulation
         AnalysisManager::Sundials::Ida<RealT, size_t> ida(&sys);
+        ida.setTolerance(static_cast<RealT>(1.0e-7),
+                         static_cast<RealT>(1.0e-9));
         ida.configureSimulation();
 
         // Run for 1s
@@ -330,7 +332,7 @@ namespace GridKit
 
         auto error_set = runSimulation(set_data, base_name + ".ref.csv");
 
-        RealT error_V_allowed = 2.1e-4;
+        RealT error_V_allowed = 2.01e-4;
         RealT error_w_allowed = 1e-4;
 
         success *= error_set->var_errors[0].max_value < error_V_allowed;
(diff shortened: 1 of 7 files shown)


nkoukpaizan:
A few tests are failing. Likely the ones using the output callback lambda with hard-coded variable indices. We really need to remove those. CC @PhilipFackler

alexander-novo:
I don't think so, this is Ida failing. So something about the model changed in a way that the solver isn't happy about.

tests/IntegrationTests/PhasorDynamics/PDIntegrationTests.hpp
```diff
@@ -330,7 +330,7 @@ namespace GridKit
 
         auto error_set = runSimulation(set_data, base_name + ".ref.csv");
 
-        RealT error_V_allowed = 2.01e-4;
+        RealT error_V_allowed = 2.1e-4;
```
nkoukpaizan: This is no longer consistent with `TwoBusBasic.solver.json`, and `TwoBusBasicJson.cpp` needs to call `ida.setMaxSteps` to fix the failing test. We also need to remove `TwoBusBasicJson.cpp` and use a different consumer test. @PhilipFackler

lukelowry: These tests are more outdated than I thought. In addition to not setting the max steps, it was using an older signature for the function:

```cpp
ida.runSimulation(10.0, nout);
```

which should be

```cpp
ida.runSimulation(10.0, dt_monitor);
```

I also set the tolerances to be `1e-7` and `1e-9` (which is what they were when this test was originally written) since the default tolerance changed at some point in the past year.

@nkoukpaizan let me know what you think of [138e85a](https://github.com/ORNL/GridKit/pull/556/commits/138e85a8eb3a637f3d7a98a1fc3d515f4f3bf5ce)

lukelowry: Rebased, meant [cf5239e](https://github.com/ORNL/GridKit/pull/556/commits/cf5239e649d3eab7fcd06af30bda79d0d7ced309)

nkoukpaizan (approved):
Looks good, thanks @lukelowry

PR #559, opened 2026-08-28
Output the correct variable for TenGenGenrou example

alexander-novo:
## Description
 
The `Genrou` model was updated, and the number of equations was changed, after the `TenGen` example was created. Since the `TenGen` example wasn't updated with these changes, it was outputting the wrong values.
 
 ## Proposed changes
 
Change from a hard-coded magic number for the number of equations in a generator to querying the size at runtime.
 
 ## Checklist
 
- [x] All tests pass.
 
 ## Further comments

### DIFF
--- a/examples/PhasorDynamics/Small/TenGen/Genrou/TenGenGenrou.cpp
+++ b/examples/PhasorDynamics/Small/TenGen/Genrou/TenGenGenrou.cpp
@@ -135,9 +135,8 @@ int main()
     for (size_t i = 0; i < 9; ++i)
     {
       // 18 is offset for variables of 9 buses.
-      // Each generator has 21 equations.
       // We are outputting second equation of each generator.
-      out << yval[18 + 21 * i + 1] << ",";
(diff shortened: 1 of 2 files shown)


examples/PhasorDynamics/Small/TenGen/Genrou/TenGenGenrou.cpp
```diff
@@ -135,9 +135,8 @@ int main()
     for (size_t i = 0; i < 9; ++i)
     {
       // 18 is offset for variables of 9 buses.
-      // Each generator has 21 equations.
       // We are outputting second equation of each generator.
-      out << yval[18 + 21 * i + 1] << ",";
+      out << yval[18 + gen2.size() * i + 1] << ",";
```
lukelowry, line 139: These are a pain to debug, I apologize. The examples for the early models are definitely not ideal

lukelowry (approved):
Easy fix, good catch

PR #492, opened 2026-07-16
Initial Implementation of Subsystem Model for Partition Simulation

abdourahmanbarry:
## Description
 
In this PR, we introduce the initial implementation of the ```SubsystemModel``` for partitioned Power Electronics simulation. The implementation provides the infrastructure required to partition Power Electronic networks, and to evaluate subsystem residuals and Jacobians independently. This lays the foundation for implementing co-simulation methods in future pull requests.
 
 ## Proposed changes
 
- We added a ```SubsystemModel``` class to represent an individual partition and a ```BusPartitionInterface``` component to mark partition boundaries.
- Implemented residual and Jacobian evaluation for subsystem models.
- Added support for user-defined coupling functions required for solving subsystems with multi-stage co-simulation methods.
 
 ## Checklist
 
- [x] All tests pass.
- [x] Code compiles cleanly with flags `-Wall -Wpedantic -Wconversion -Wextra`.
- [x] The new code follows GridKit™ style guidelines.
- [x] There are unit tests for the new code.
- [x] The new code is documented.
- [x] The feature branch is rebased with respect to the target branch.
- [x] I have updated [CHANGELOG.md](/CHANGELOG.md) to reflect the changes in this PR. If this is a minor PR that is part of a larger fix already included in the file, state so.

PR #560, opened 2026-08-28
Add Subsystem and Partition Interfaces for PowerElectronics

abdourahmanbarry:
## Description
 
In this PR, we implement ```SubsystemModel``` for partitioned Power Electronics simulation. The implementation provides the infrastructure required to partition Power Electronics networks, and to evaluate subsystem residuals and Jacobians independently.

This is the third of four stacked pull requests that split PR #492 into simpler, more manageable chunks.
 
 ## Proposed changes
 
We added a ```SubsystemModel``` class to represent an individual partition and a ```BusPartitionInterface``` component to mark partition boundaries. Together, these enable us to partition larger networks for co-simulation and parallel function evaluation.
 
 ## Checklist
 
- [x] All tests pass.


GridKit/Model/PowerElectronics/SubsystemModel.hpp
```diff
@@ -0,0 +1,856 @@
...
+   *
+   * External coupling values must be supplied before residual or Jacobian
+   * evaluation, either directly through the external-data vectors or through a
+   * forcing function.
+   *
+   * @todo Find a better name for this class and its base class.
+   *
+   * @tparam ScalarT Scalar type used by the model.
+   * @tparam IdxT Index type used for variable and connection indices.
+   */
```
superwhiskers, line 50: so, my main question here is why do we need this notion of distinct "systems" and "subsystems" present in the codebase?

when first learning about system partitioning stuff in gridkit, it seemed more natural to me to instead allow the system itself to be a component that can be used in another system. the "non-owning" nature of `SubsystemModel` as implemented here seems like it could be more naturally captured by a `SystemModelRef` or some similar type that encapsulates the notion of "referencing" a system model or something like that (perhaps to match the bus partition interface thing for components---`SystemModelInterface`).

what was the motivation behind this design choice? it seems more natural to build up from smaller components than to break apart.

superwhiskers (commented):
just an architectural question

GridKit/Model/PowerElectronics/SubsystemModel.hpp
abdourahmanbarry: Good question. There are several reasons for doing it this way.

First, when we partition a system, each subsystem is smaller than the original system, so we can no longer directly use the global indices of the components when evaluating the subsystem. We therefore need to map the global system indices to the corresponding local subsystem indices. This mapping is not needed when evaluating the monolithic system, but it is something we must do before we can evaluate partitions independently.

Second, there is an important difference in how the two models are used. ```PowerElectronicsModel``` is designed to construct and own a network. We add components and nodes to it to build up the system. ```SubsystemModel``` expects to receive part of an already existing network. It is not trying to construct another network or own those components. If we combine everything into a single class, the same class would need logic for both constructing a network(not need when you are absorbing a part of an already existing network) and representing a partition of an existing network, in addition to handling local/global index mappings and internal/external variables management. I think keeping those responsibilities separate makes the design easier to understand. In addition, the ```SubsystemModel``` is still a System model because it inherits and reuses a lot of code from the System Model.

I agree with you that, ideally, we would want a System Model that can be treated as a component, and can be nested hierarchically to arbitrary depth. This was actually one of our design considerations. However, we agreed that supporting that design properly would require a substantial amount time, but we do not have much time left under the scope of the AGM project.

superwhiskers: i think that we should open an issue regarding "componentized" system models to replace the partitioning implementation approach implemented in this pull request if doing so right now is too difficult

abdourahmanbarry: Screenshot 2026-09-04 140924

We are already heading down that direction with this implementation. It might need couple of cycles to get there. This is a good starting point I think. I have attached a snapshot of a design document I shared with Slaven couple of months ago. It needs to be update, but this section talks about doing exactly what you are thinking.

GridKit/Model/PowerElectronics/SystemModelPowerElectronics.hpp
```diff
@@ -430,7 +424,12 @@ namespace GridKit
 
         for (IdxT i = 0; i < nnz; ++i)
         {
-          if (component->getNodeConnection(r[i]) != neg1_ && component->getNodeConnection(c[i]) != neg1_)
+          const IdxT row = component->getNodeConnection(r[i]);
+          const IdxT col = component->getNodeConnection(c[i]);
+
+          const bool is_internal_entry = row != neg1_ && col != neg1_ && row < n_intern_ && col < n_intern_;
```
nkoukpaizan, line 430: Minor: The `neg1_` alias no longer makes sense. I recommend using `INVALID_INDEX<IdxT>` directly.

tests/UnitTests/PowerElectronics/SubsystemModelWithHiresTest.hpp
```diff
@@ -0,0 +1,774 @@
...
+
+        for (size_t i = 0; i < system_->size(); ++i)
+        {
+          double error = std::abs(partition_residual[i] - reference_residual[i]) / std::abs(reference_residual[i] + 1);
+          max_error    = std::max(max_error, error);
+        }
+
+        std::cout << "max error " << max_error << std::endl;
+
+        success *= max_error <= std::numeric_limits<RealT>::epsilon();
```
nkoukpaizan, line 653: I see your comment in the other pull request about not using isEqual, because you want to compute `max_error`, but why compute `max_error` if we can check every value instead? I would avoid re-implementing the error calculation logic here.

tests/UnitTests/PowerElectronics/SubsystemModelWithHiresTest.hpp
```diff
@@ -0,0 +1,774 @@
...
+      return 0;
+    }
+
+    CircuitComponent<ScalarT, IdxT>* clone() const
+    {
+      return new HiresComponent3<ScalarT, IdxT>(*this);
+    }
+
+  private:
+    NodeT* node_ref_;
```
nkoukpaizan:
```suggestion
    NodeT* node_ref_{nullptr};
```
Plus matching handling in the destructor or use a `unique_ptr`.

tests/UnitTests/PowerElectronics/SubsystemModelWithHiresTest.hpp
```diff
@@ -0,0 +1,774 @@
...
+
+namespace GridKit
+{
+
+  /*!
+   * @brief Hires Component 1 class.
+   *
+   */
+  template <class ScalarT, typename IdxT>
+  class HiresComponent1 : public CircuitComponent<ScalarT, IdxT>
```
nkoukpaizan, line 25: Will need to update the namespaces after #576

tests/UnitTests/PowerElectronics/SubsystemModelWithHiresTest.hpp
```diff
@@ -0,0 +1,774 @@
...
+
+          auto* partition_jac = partition->getCsrJacobian();
+
+          const auto* row_ptr = partition_jac->getRowData();
+          const auto* cols    = partition_jac->getColData();
+          const auto* vals    = partition_jac->getValues();
+
+          const size_t n = partition->getInternalSize();
+
+          std::vector<std::vector<RealT>> dense_jac(n, std::vector<RealT>(n, 0.0));
```
nkoukpaizan, line 707: Consider using GridKit dense vector (multiple columns) or keeping things sparse with `std::map`. You could use `GridKit::Testing::MapFromCsr`.

tests/UnitTests/PowerElectronics/SubsystemModelWithHiresTest.hpp
```diff
@@ -0,0 +1,774 @@
...
+        : system_(new System()),
+          partition1_(new Subsystem()),
+          partition2_(new Subsystem()),
+          comp1_(new HiresComponent1<ScalarT, IdxT>(&bus_, 1)),
+          bus1_(new HiresBus<ScalarT, IdxT>(&bus_, 2)),
+          comp3_(new HiresComponent3<ScalarT, IdxT>(&bus_, 3))
+      {
+
+        y_  = {1, 2, 3, 4, 5, 6, 7, 8};
+        yp_ = {1, 2, 3, 4, 5, 6, 7, 8};
```
nkoukpaizan, line 513: I would skip the `std::vector`. `std::array` if you really have to.

GridKit/Model/PowerElectronics/PartitionInterface/BusPartitionInterface.hpp
```diff
@@ -0,0 +1,168 @@
...
+    // int evaluateAdjointJacobian();
+    int evaluateAdjointIntegrand() final;
+
+  private:
+    int updateComponentPointers();
+
+    /**
+     * @brief Circuit component participating in the partition split.
+     */
+    component_type* component_;
```
nkoukpaizan, line 116:
```suggestion
    component_type* component_{nullptr};
```
or use `unique_ptr` to clarify ownership.

GridKit/Model/PowerElectronics/PartitionInterface/BusPartitionInterface.cpp
```diff
@@ -0,0 +1,365 @@
...
+    for (size_t i = 0; i < static_cast<size_t>(size_); ++i)
+    {
+      const IdxT connection_index = component_->getNodeConnection(i);
+
+      this->setConnectionNodes(i, connection_index);
+
+      if (!external_indices.contains(static_cast<IdxT>(i)))
+      {
+        continue;
+      }
```
nkoukpaizan, line 122: I usually avoid the `continue` pattern and prefer checking the opposite condtion.

GridKit/Model/PowerElectronics/PartitionInterface/BusPartitionInterface.cpp
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  /**
+   * @brief Eval Internal Residual
+   */
+  template <class ScalarT, typename IdxT>
+  int BusPartitionInterface<ScalarT, IdxT>::evaluateInternalResidual()
+  {
+    return 0;
+  }
```
nkoukpaizan, line 205: Please add the relevant comment from the class documentation ("all variables are external from the interface's point of view") here to explain why this is no-op.

nkoukpaizan (commented):
A few specific comments. 

This seems to work, but I find the logic of the `BusPartitionInterface` constructor and `SubsystemModel` index maps hard to follow. I would like to get inputs from @alexander-novo on those implementations.

I'll take another look after the initial comments have been addressed.

abdourahmanbarry:
@nkoukpaizan @alexander-novo @superwhiskers  I have updated the code to incorporate the suggestions and improved the documentation to make the implementation easier to follow. Let me know what you think.

PR #561, opened 2026-09-01
Bus Fault Jacobian Bug

lukelowry:
## Description
 
The  implementation of the `BusFault` control signal, followed by unfortunate sequencing in `DynamicSimulation` , causes incorrect evaluation of the Enzyme Jacobian.

The problematic term is `DhDy`. When the fault is inactive, the bus residual contribution h is omitted, so `DhDy` must be zero. Enzyme still evaluates the structural `DhDy` block, and the existing status mask was applied before that block was generated. Moving the mask after `DhDy` correctly masks both `DfDwb` and `DhDy`, while leaving `DfDy` active to constrain the fault-current states to zero.
 
 ## Proposed changes
 
- Update tests to reproducing the issue
- Fix bug by updating the Jacobian logic
 
 ## Checklist
 
- [x] All tests pass.
- NA The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes. If this is a minor PR that is part of a larger fix already included in the file, state so.
 
 ## Further comments
 
I observed InitialCondition errors in IDA, and after some eyebrow-raising simulation results, a painful investigation led me to this. It is not clear to me how or why this does not significantly impact the working simulations.

Domain usability:

I think events/external controls need to be polished (i.e. the wart of requiring  `BusFault` at each `Bus` in the case file to use `ContingencyAnalysis`). Conceptually, it would be preferred if `BusFault` were added in `DynamicSimulation` so that `R` and `X` are configurable per simulation and not per case. (See #389 #382 #363 )

### DIFF
--- a/GridKit/Model/PhasorDynamics/BusFault/BusFaultEnzyme.cpp
+++ b/GridKit/Model/PhasorDynamics/BusFault/BusFaultEnzyme.cpp
@@ -66,14 +66,6 @@ namespace GridKit
                                                                                                        J_cols_buffer_,
                                                                                                        J_vals_buffer_,
                                                                                                        nnz_);
-      if (!status_) // Value contributions from DfDwb only when status_
-      {
-        for (IdxT i = nnz_tmp; i < nnz_; ++i)
-        {
-          J_vals_buffer_[i] = 0.0;
-        }
-      }
-
       GridKit::Enzyme::Sparse::DhDy<GridKit::PhasorDynamics::BusFault<ScalarT, IdxT>,
                                     GridKit::Enzyme::Sparse::MemberFunctions::BusResidual>::eval(this,
                                                                                                  static_cast<size_t>(bus_->size()),
@@ -87,6 +79,14 @@ namespace GridKit
                                                                                                  J_cols_buffer_,
                                                                                                  J_vals_buffer_,
                                                                                                  nnz_);
+
+      if (!status_) // Value contributions from DfDwb and DhDy only when status_
+      {
+        for (IdxT i = nnz_tmp; i < nnz_; ++i)
+        {
+          J_vals_buffer_[i] = 0.0;
+        }
+      }
 
       this->constructCoo();
 
--- a/tests/UnitTests/PhasorDynamics/CMakeLists.txt
+++ b/tests/UnitTests/PhasorDynamics/CMakeLists.txt
@@ -15,7 +15,6 @@ target_link_libraries(
   GridKit::phasor_dynamics_bus_fault_dependency_tracking
   GridKit::phasor_dynamics_bus
   GridKit::phasor_dynamics_bus_dependency_tracking
-  GridKit::phasor_dynamics_systemmodel
   GridKit::testing)
 
 add_executable(test_phasor_bustosignaladapter runBusToSignalAdapterTests.cpp)
--- a/tests/UnitTests/PhasorDynamics/CMakeLists.txt
+++ b/tests/UnitTests/PhasorDynamics/CMakeLists.txt
@@ -15,6 +15,7 @@ target_link_libraries(
   GridKit::phasor_dynamics_bus_fault_dependency_tracking
   GridKit::phasor_dynamics_bus
   GridKit::phasor_dynamics_bus_dependency_tracking
+  GridKit::phasor_dynamics_systemmodel
   GridKit::testing)
 
 add_executable(test_phasor_bustosignaladapter runBusToSignalAdapterTests.cpp)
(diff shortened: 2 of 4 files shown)


GridKit/Model/PhasorDynamics/BusFault/BusFaultEnzyme.cpp
```diff
@@ -25,68 +25,32 @@ namespace GridKit
...
-                                                                                                 static_cast<size_t>(y_.getSize()),
-                                                                                                 (bus_->getResidualIndices()).data(),
-                                                                                                 (this->getVariableIndices()).data(),
-                                                                                                 y_.getData(),
-                                                                                                 yp_.getData(),
-                                                                                                 wb_.data(),
-                                                                                                 J_rows_buffer_,
-                                                                                                 J_cols_buffer_,
-                                                                                                 J_vals_buffer_,
-                                                                                                 nnz_);
```
nkoukpaizan, line 89: I believe the minimal fix is moving the branching after the evaluation of the `DhDy` term. The comment would read `Value contributions from DfDwb and DhDy only when status_`.

nkoukpaizan (commented):
> The implementation of the BusFault control signal, followed by unfortunate sequencing in DynamicSimulation , causes incorrect evaluation of the Enzyme Jacobian,

Could you be more specific in the PR description regarding the incorrect terms? I think it's the `DhDy` term.

It's a good catch! Looks like a bug I introduced in #447.

I would keep this PR strictly about the bug fix and revert changes about the sizes of the faults and accessing them in `ContingencyAnalysis` and `DynamicSimulation`. I'm thinking the extra internal variables were a bad idea, and I'd rather have a separate PR to change all components.

GridKit/Model/PhasorDynamics/BusFault/BusFaultEnzyme.cpp
lukelowry: @nkoukpaizan good point I got a bit carried away. will adjust asap

lukelowry: @nkoukpaizan Checkout out [7eb7ea8](https://github.com/ORNL/GridKit/pull/561/commits/7eb7ea8c4c1ecf475b7876591c83f39965b0c166). I branched off my other changes for later

lukelowry: I updated  the PR body as requested, asw ell

tests/UnitTests/PhasorDynamics/runBusFaultTests.cpp
```diff
@@ -12,6 +12,7 @@ int main()
   result += test.zeroInitialResidual(true);
 #ifdef GRIDKIT_ENABLE_ENZYME
   result += test.jacobian(true);
+  result += test.jacobianAfterClearing();
```
nkoukpaizan: `test.jacobian(false)` should cover what `jacobianAfterClearing()` is testing, and we shouldn't need to include the system model. However, the explicit zeros would make the test fail. I recommend a (clearly documented) hack to ignore the extra zeros that differentiate the Enzyme and DependencyTracking Jacobian.

lukelowry: How's [a44df2b](https://github.com/ORNL/GridKit/pull/561/commits/a44df2b2c54ccfca91371c98c28fa937f8ad69c5)? Had ot change DepTrack Jacobian helper too

tests/UnitTests/PhasorDynamics/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ target_link_libraries(
   GridKit::phasor_dynamics_bus_fault_dependency_tracking
   GridKit::phasor_dynamics_bus
   GridKit::phasor_dynamics_bus_dependency_tracking
+  GridKit::phasor_dynamics_systemmodel
```
nkoukpaizan:
```suggestion
```

tests/UnitTests/PhasorDynamics/runBusFaultTests.cpp
```diff
@@ -12,6 +12,7 @@ int main()
   result += test.zeroInitialResidual(true);
```
nkoukpaizan, line 12:
```suggestion
  result += test.zeroInitialResidual(true);
  result += test.zeroInitialResidual(false);
```

lukelowry:
Applied and rebased

nkoukpaizan (approved):

Issue #504, opened 2026-07-22
Configurable MU for CommonMath


lukelowry:
### Summary

The $\mu$ parameter, which represents the sharpness of the piecewise approximants, needs to be configurable per-component/caller instead of a static value.

### Rationale

There is an edge case I had not considered before:

A smoothed band pass window in state space, $f(x; w, \mu)=\sigma(x;\mu)-\sigma(x-w;\mu)$ , has a width $w$.  When the width shrinks to scales below $w<\dfrac{5}{\mu}$, the smooth approximation begin to fail as a piecewise indicator. 

While validating individual governor models I observed the effects of this when the limits $[V^{\min}, V^{\max}]$ become very tight $w=V^{\max}-V^{\min}$. 

### Description

In the motivating example above, I need to be able to set $\mu$ and pass it as an argument to functions such as `antiwindup` or `inside` in `CommonMath`. A global $\mu$ is not preferred because one model could have subsystems with varying constraints on $\mu$!

### Additional information

I temporary solution could be to raise $\mu$ to a larger value, but since that impacts the whole system it hurts performance especially on large cases.

nkoukpaizan:
An optional argument to the math functions that defaults to the `MU<RealT>` constant should work. I recommend a separate PR that addresses this issue if you want to implement it.

pelesh:
It would be good to add a solver handle for adjusting $\mu$. A continuation method would set values of $\mu$ automatically

PR #569, opened 2026-09-04
Initial mu configurability

lukelowry:
## Description

This is an initial, intentionally small change to make the CommonMath smoothing scale runtime-configurable. PhasorDynamics studies can now set `mu` ($\mu$) in a `.solver.json` file, allowing parameter studies without rebuilding GridKit for every value.

## Proposed changes

- Add an optional `mu` with the existing `240.0` default
- Require `mu` to be positive and finite
- Apply it before model construction and contingency workers
- Document the setting and add runtime behavior coverage

## Checklist

- [x] All tests pass.
- N/A The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes.

## Further comments

This is not the best long-term API for configuring $\mu$ at runtime, but without this, the results of our paper are very hard to reproduce without a development environment. For that reason, I would like to have this in before we do the first release.

### DIFF
--- a/GridKit/CommonMath.hpp
+++ b/GridKit/CommonMath.hpp
@@ -10,16 +10,22 @@ namespace GridKit
 {
   namespace Math
   {
+    template <typename RealT>
+    inline constexpr RealT DEFAULT_MU = 240.0;
+
     /**
      * @brief Smoothing scale shared by CommonMath primitives
      *
      * Used by @ref sigmoid, @ref ramp, and functions composed from them to set
      * the width of smooth transitions.
      *
+     * @warning Process-wide setting; configure before constructing models or
+     * launching workers.
+     *
      * @tparam RealT - real data type
      */
     template <typename RealT>
-    inline constexpr RealT MU = 240.0;
+    inline RealT MU = DEFAULT_MU<RealT>;
 
     /**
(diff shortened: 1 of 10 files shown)


GridKit/CommonMath.hpp
```diff
@@ -10,16 +10,22 @@ namespace GridKit
...
      * the width of smooth transitions.
      *
+     * @warning Process-wide setting; configure before constructing models or
+     * launching workers.
+     *
      * @tparam RealT - real data type
      */
     template <typename RealT>
-    inline constexpr RealT MU = 240.0;
+    inline RealT MU = DEFAULT_MU<RealT>;
```
nkoukpaizan, line 28: I don't think we can merge this, even as a temporary solution. The provenance of the value used will be very hard to establish.

lukelowry: I cannot think of an alternative that would be a small change. I would love to have this for the purpose of the paper,  but I see the issue. 

Unless you have any ideas on how to do this safely, I will close this

nkoukpaizan: The ideas I have require widespread changes. We can close for now and keep tracking in #504.

PR #580, opened 2026-09-14
Reject a blank row in the gencost matrix instead of reading past the string

arpitjain099:
`readMatPowerGenCostRow` checks the trailing semicolon with `row[row.size() - 1]` after `rtrim`. On a blank line the string is empty, `size() - 1` wraps, and the subscript reads past the buffer. The gencost loop in `readMatPower` hands every line up to `];` to the row reader without filtering empty ones, so a case file with a stray blank line between two gencost rows reaches it.

Built with `-fsanitize=address` against the header as it stands:

    control (no blank line): parsed, rows=2
    ERROR: AddressSanitizer: stack-buffer-underflow
      #0 readMatPowerGenCostRow MatpowerParser.hpp:156
      #1 readMatPower MatpowerParser.hpp:282

The file that triggers it is a normal case file, nothing malicious, just an extra newline in the matrix.

The condition already throws `matlab_syntax_error` for a row without a trailing semicolon, so adding `row.empty()` to it puts a blank row on that same path rather than inventing new behaviour. I added the case to `test_parse_gencost_row`, which aborts under ASan before the change and passes after it.

One thing I noticed while checking the siblings, not changed here: `readMatPowerGenRow` and `readMatPowerBranchRow` both end with `checkEndOfMatrixRow`, so they reject a blank row, but `readMatPowerBusRow` does not call it. A blank line in `mpc.bus` is currently accepted and appends a bus row with every field zero. Happy to send that separately if you want it tightened up, since it changes what the bus parser accepts rather than fixing a read.


pelesh:
@arpitjain099, thanks for your contribution. Please create an issue with reproducer and suggested solution. Then we can decide how to move forward. I will close this PR until then.

PR #581, opened 2026-09-14
Bug: RTD Doxygen Build

lukelowry:
## Description

Fix some Doxygen comments and links that failed on RTD. 

This branch fixes those: https://gridkit.readthedocs.io/en/lukel-dev/


lukelowry:
@nkoukpaizan @superwhiskers Whoever can get to this first, super quick fix so the RTD can build correctly.

GridKit/Solver/Dynamic/Native/InfNorm.hpp
```diff
@@ -38,7 +38,7 @@ namespace AnalysisManager
          */
         std::unique_ptr<State> scale_;
         /**
-         * @brief The absolute value of yprev. Used to calculate \ref scale_
+         * @brief The absolute value of yprev. Used to calculate `scale_`
```
nkoukpaizan: I would expect `\ref scale_` to work... CC @alexander-novo

nkoukpaizan: I would expect \ref scale_ to work... CC @alexander-novo

lukelowry: RTD did not like it, I think we saw similar issue before @alexander-novo

alexander-novo: I'll take a look at this, but in general, I don't like making the Doxygen documentation worse just to support RTD.

lukelowry: I am trying to get more details on why this causes issues so we don't have to compromise

I don't use Doxygen, so that's why I was swift to correct it this way

lukelowry: Fixed Doxygen needed a namespace for these two for some reason (`\ref InfNorm::scale_` vs `\ref scale_`)

@nkoukpaizan @alexander-novo thank you!

alexander-novo: When I run Doxygen on the Doxyfile included in `docs` on `develop`, I get an error about `std::size_t`, but fixing that produces no more errors for me:
```
anovotny@sibiu:~/GridKit/docs$ doxygen Doxyfile
/home/grads/anovotny/GridKit/GridKit/Utilities/Enum.hpp:42: error: unable to resolve reference to 'std::size_t' for \ref command
anovotny@sibiu:~/GridKit/docs$ doxygen Doxyfile
anovotny@sibiu:~/GridKit/docs$ 
```

It could be the older version of Doxygen? I could try downgrading to the precise version. But also it should be possible to direct RTD to use a newer version of Doxygen

nkoukpaizan (approved):

PR #583, opened 2026-09-16
SUNDIALS Upgrades

Steven-Roberts:
## Description

- Switches SUNDIALS dependency from develop branch to version 7.8.0 which is available with spack now
- Adds a max order option to Ida.cpp and phasor dynamics apps
- Fixes a bug with the absolute tolerances in fixed step mode (thanks @lukelowry for catching this)
- Adds exception handling to phasor dynamics apps. Without this, buffered text wasn't written to file and SUNDIALS logs would get cut short
 
 ## Checklist
 
- [ ] All tests pass.
- [x] The [CHANGELOG.md](/CHANGELOG.md) has been updated to reflect the changes. If this is a minor PR that is part of a larger fix already included in the file, state so.


GridKit/Solver/Dynamic/Ida.cpp
```diff
@@ -1271,6 +1299,7 @@ namespace AnalysisManager
                                            ScalarT rel_tol,
                                            ScalarT abs_tol_override,
                                            IdxT    max_steps,
+                                           int     max_order,
```
nkoukpaizan, line 1302: Is there a particular reason for `int` rather than `IdxT`?

nkoukpaizan (approved):

GridKit/Solver/Dynamic/Ida.cpp
pelesh: Order can be 1, ... , 5. I think it is good to distinguish that the scheme order is much smaller than the vector index range.

